Issue #6, opened 2020-09-11
Parse debugger statements


strager:
JavaScript has a `debugger` keyword which is used for ... debugging. quick-lint-js gets confused when it sees a `debugger` statement.

1. Tell quick-lint-js that `debugger` is a keyword, not a normal identifier. See lex.h, lex-keyword.gperf, and test-lex.cpp.
2. Teach quick-lint-js about `debugger` statements. See parse.h and test-parse.cpp. `debugger` statements should have no impact on the linter (I think).

erlliam:
I will give this a try.

strager:
@erlliam implemented `debugger` in #17.

PR #17, opened 2020-09-13
Parse debugger statements

erlliam:
Parsing a `debugger` statement follows the same behavior as parsing a `break`/`continue` statement.

#6 As far as I am concerned, quick-lint-js already knew that `debugger` was a keyword and not an identifier.

### DIFF
--- a/src/include/quick-lint-js/parse.h
+++ b/src/include/quick-lint-js/parse.h
@@ -158,6 +158,11 @@ class parser {
         this->consume_semicolon();
         break;
 
+      case token_type::kw_debugger:
+        this->lexer_.skip();
+        this->consume_semicolon();
+        break;
+
       case token_type::left_curly:
         v.visit_enter_block_scope();
         this->parse_and_visit_statement_block_no_scope(v);
--- a/test/test-parse.cpp
+++ b/test/test-parse.cpp
@@ -1833,5 +1833,19 @@ TEST(test_parse, continue_statement) {
     EXPECT_THAT(v.visits, IsEmpty());
   }
 }
+
+TEST(test_parse, debugger_statement) {
+  {
+    spy_visitor v;
+    padded_string code(u8"debugger; x;");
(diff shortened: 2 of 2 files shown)


src/include/quick-lint-js/parse.h
```diff
@@ -154,6 +154,7 @@ class parser {
 
       case token_type::kw_break:
       case token_type::kw_continue:
+      case token_type::kw_debugger:
```
strager, line 157: Currently, grouping `debugger` with `break` and `continue` makes sense. However, `break` and `continue` can accept a loop label (not implemented in quick-lint-js yet), and I assume `debugger` doesn't accept a loop label. So while the *code* is currently the same, I think the *meanings* are different, so I'd repeat lines 158-160 for `debugger`.

test/test-parse.cpp
```diff
@@ -1833,5 +1833,12 @@ TEST(test_parse, continue_statement) {
     EXPECT_THAT(v.visits, IsEmpty());
   }
 }
+
+TEST(test_parse, debugger_statement) {
+  {
+    spy_visitor v = parse_and_visit_statement(u8"debugger;");
+    EXPECT_THAT(v.visits, IsEmpty());
```
strager, line 1840: Minor: Checking that `v.visits` is empty is a bit risky. What if the parser consumed no tokens at all? What if the parser didn't consume the semicolon?

One way we've handled this case before: add a canary statement at the end of the JavaScript snippet, parse two statements (instead of one), then assert that the canary statement was seen by the parser. See the `if (true) return; x;` test `test_parse.return_statement` for an example.

Testing this much might be more trouble than necessary. It's easy to go overboard with testing. =X

strager (approved):
This change looks good to me.

Let me know whether you want me to land these changes as-is or if you want to change your PR first.

erlliam:
I have made the changes to my original commit.

PR #19, opened 2020-09-14
Teaching quick-lint-js about number literals like 1_234_567

LeeWannacott:
I left TODO comments for once the error reporting is changed, but the logic for __ and trailing _ is throwing an error is done(I think?).

I called parse underscore in parse decimal to handle numbers like 1.234_567 this also solves 3e1_2 because exponents section calls parse decimal which if it has underscore in the number calls parse underscore.

Update:  for the trailing underscore I added a boolean value.

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -663,11 +663,11 @@ void lexer::parse_number() {
     input = garbage_end;
   };
 
-  input = this->parse_decimal_digits(input);
+  input = this->parse_decimal_digits_and_underscores(input, number_begin);
   bool has_decimal_point = *input == '.';
   if (has_decimal_point) {
     input += 1;
-    input = this->parse_decimal_digits(input);
+    input = this->parse_decimal_digits_and_underscores(input, number_begin);
     has_decimal_point = true;
   }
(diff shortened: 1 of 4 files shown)


src/lex.cpp
```diff
@@ -709,9 +711,33 @@ void lexer::parse_number() {
   this->input_ = input;
 }
 
+bool has_trailing_underscore = false;
```
strager, line 714: Blocking: Do not use a global variable for this. Use a local variable.

src/lex.cpp
```diff
@@ -709,9 +711,33 @@ void lexer::parse_number() {
...
+      }
+    }
+  }
+  if (has_trailing_underscore == true) {
+    // TODO error_reporter: SyntaxError: Numeric separators are not allowed at the end of numeric literals
+  }
+  return input;
+}
+
 const char8* lexer::parse_decimal_digits(const char8* input) noexcept {
```
strager, line 734: Discussed in person: parse_decimal_digits and parse_underscores are doing the same thing. Merge them into one function.

strager (changes requested):

PR #20, opened 2020-09-15
Help option

jeduardofr:
Here is the implementation i did. I created a simple lambda helper to print the different options but i used `const char *` in order to pass the arguments, don't know if that's right. I saw that you had a padded-string helper but not sure if i should use it or this is ok.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,6 +17,7 @@
 #include <cerrno>
 #include <cstdio>
 #include <cstring>
+#include <iomanip>
 #include <iostream>
 #include <quick-lint-js/char8.h>
 #include <quick-lint-js/error.h>
@@ -91,6 +92,8 @@ class any_error_reporter {
 
 void process_file(padded_string_view input, error_reporter *,
                   bool print_parser_visits);
+
+void print_help_message();
 }
 }
 
@@ -98,6 +101,10 @@ int main(int argc, char **argv) {
   quick_lint_js::vector_instrumentation::register_dump_on_exit_if_requested();
 
   quick_lint_js::options o = quick_lint_js::parse_options(argc, argv);
+  if (o.help) {
+    quick_lint_js::print_help_message();
+    return 0;
+  }
   if (!o.error_unrecognized_options.empty()) {
     for (const auto &option : o.error_unrecognized_options) {
       std::cerr << "error: unrecognized option: " << option << '\n';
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -290,6 +292,19 @@ void process_file(padded_string_view input, error_reporter *error_reporter,
...
+  const int max_width = 36;
+
+  auto print_helper = [&](const char *abbr, const char *message) {
+    std::cout << std::setw(max_width) << std::left << abbr << message << std::endl;
+  };
+
+  std::cout << "Usage: quick-lint-js [OPTIONS]... [FILE]..."
+            << std::endl << std::endl
+            << "OPTIONS" << std::endl;
+  print_helper("--debug-p, --debug-parser-visits", "Output travel when parsing file");
```
strager, line 304: I wouldn't mention the `--debug-p` shortcut.

strager, line 304: Maybe don't bother documenting --debug options at all? They're for contributors, not users.

src/main.cpp
```diff
@@ -290,6 +292,19 @@ void process_file(padded_string_view input, error_reporter *error_reporter,
...
 
 void print_help_message() {
+  const int max_width = 36;
+
+  auto print_helper = [&](const char *abbr, const char *message) {
+    std::cout << std::setw(max_width) << std::left << abbr << message << std::endl;
+  };
+
+  std::cout << "Usage: quick-lint-js [OPTIONS]... [FILE]..."
+            << std::endl << std::endl
```
strager, line 302: Prefer `"\n`" over `std::endl`. It's shorter and faster (not that speed matters much here...).

src/main.cpp
```diff
@@ -290,6 +292,19 @@ void process_file(padded_string_view input, error_reporter *error_reporter,
 }
 
 void print_help_message() {
+  const int max_width = 36;
```
strager, line 295: Nit: I don't like `const` variables.

     int max_width = 36;

src/options.cpp
```diff
@@ -180,6 +181,9 @@ options parse_options(int argc, char** argv) {
       } else {
         next_vim_file_bufnr = bufnr;
       }
+    } else if (parser.match_flag_option("--help"sv, "--h"sv)) {
+      o.help = true;
+      goto done_parsing_options;
```
strager, line 186: Seems kinda weird that we stop parsing options here. I'd continue parsing options after `--help`.

strager (approved):

strager:
This PR still has conflicts with master. Do you need help resolving the conflicts?

PR #27, opened 2020-09-16
Report unexpected characters in hex literal

### LINKED ISSUE #5: Report error for invalid hexadecimal digits: 0x12abq
quick-lint-js reports an error if a decimal or binary number literal contains bogus digits:

```javascript
// hello.js:2:5: error: unexpected characters in number literal
1234abcd
```
Give hexadecimal literals the same error-reporting treatment. See parse_binary_number in lex.cpp, and tests in test-lex.cpp.

welenofsky:
This may resolve #5 if I understood the request.

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -615,9 +615,13 @@ const char8* lexer::end_of_previous_token() const noexcept {
 
 void lexer::parse_hexadecimal_number() {
   QLJS_ASSERT(this->is_hex_digit(this->input_[0]) || this->input_[0] == '.');
-  while (this->is_hex_digit(this->input_[0])) {
-    this->input_ += 1;
+  const char8* input = this->input_;
+
+  while (this->is_hex_digit(*input)) {
+    input += 1;
   }
+
+  this->input_ = check_garbage_in_number_literal(input);
 }
 
 void lexer::parse_binary_number() {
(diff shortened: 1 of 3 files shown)


src/lex.cpp
```diff
@@ -628,18 +632,22 @@ void lexer::parse_binary_number() {
...
+        break;
       default:
         goto done_parsing_garbage;
     }
   }
-done_parsing_garbage:
+  done_parsing_garbage:
   const char8* garbage_end = input;
   if (garbage_end != garbage_begin) {
     this->error_reporter_->report(error_unexpected_characters_in_number{
```
strager, line 653: Line 658: report_garbage_in_number_literal accepts the input pointer, the assigns to `this->input_` at the end. This is different from every other function in this file; functions either exclusively manipulate `this->input_`, or they accept the input pointer and return the modified input pointer. I think report_garbage_in_number_literal should do one or the other. For example:

```cpp
const char8* lexer::report_garbage_in_number_literal(const char8* input) {
```

src/lex.cpp
```diff
@@ -628,18 +632,22 @@ void lexer::parse_binary_number() {
     input += 1;
   }
 
+  report_garbage_in_number_literal(input);
+}
+
+void lexer::report_garbage_in_number_literal(const char8* input) {
```
strager, line 638: Nit: I think `check_garbage_in_number_literal` is a better name, since this function doesn't always report an error.

strager, line 638: Thanks for making a function! It's much easier to read `lexer::parse_binary_number` now.

strager (approved):
Looks good! Let me know if you want to make changes, or if I should land this PR as-is.

src/lex.cpp
welenofsky: Updated in latest commit. I agree that is a better name.

src/lex.cpp
welenofsky: Updated in latest commit.

Issue #12, opened 2020-09-12
Clean up temp dirs created by tests


strager:
The tests in test-file.cpp create a temporary directory on the file system, but they don't clean up when they're done. When a test finishes, delete the temporary directory and any files which might exist inside.

PR #29, opened 2020-09-17
Implementation of tearDown method for tests to clean-up temp files

dylanelliott27:
In reference to issue #12

### DIFF
--- a/test/test-file.cpp
+++ b/test/test-file.cpp
@@ -22,6 +22,7 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <iostream>
+#include <optional>
 #include <quick-lint-js/assert.h>
 #include <quick-lint-js/char8.h>
 #include <quick-lint-js/file.h>
@@ -50,8 +51,18 @@ void write_file(filesystem::path, const std::string &content);
 class test_file : public ::testing::Test {
  public:
   filesystem::path make_temporary_directory() {
-    // TODO(strager): Delete the directory when the test finishes.
-    return quick_lint_js::make_temporary_directory();
+    temp_directory_path = quick_lint_js::make_temporary_directory();
+    return temp_directory_path.value();
+  }
+
+ private:
+  std::optional<filesystem::path> temp_directory_path;
+
+ protected:
+  void TearDown() override {
+    if (temp_directory_path.has_value()) {
+      filesystem::remove_all(temp_directory_path.value());
+    }
   }
 };
 


test/test-file.cpp
```diff
@@ -46,38 +46,41 @@ namespace quick_lint_js {
...
+
+ protected:
+  void TearDown() override { filesystem::remove_all(temp_file_path); }
 };
 
 TEST_F(test_file, read_regular_file) {
-  filesystem::path temp_file_path =
-      this->make_temporary_directory() / "temp.js";
-  write_file(temp_file_path, "hello\nworld!\n");
+  temp_file_path = this->make_temporary_directory();
```
strager, line 62: Assigning to `temp_file_path` everywhere is cumbersome and error-prone. Instead of assigning to `temp_file_path` here, can we assign to `temp_file_path` inside the `make_temporary_directory` function (on line 54)?

test/test-file.cpp
```diff
@@ -46,38 +46,41 @@ namespace quick_lint_js {
...
 
 class test_file : public ::testing::Test {
  public:
   filesystem::path make_temporary_directory() {
-    // TODO(strager): Delete the directory when the test finishes.
     return quick_lint_js::make_temporary_directory();
   }
+
+ protected:
+  void TearDown() override { filesystem::remove_all(temp_file_path); }
```
strager, line 58: Almost-blocking: What happens if `filesystem::remove_all` is given an empty path? (A call with an empty path happens if a test doesn't assign to `temp_file_path`.) I think we should only call `filesystem::remove_all` if we create a directory. Perhaps we could use [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) or [`std::vector`](https://en.cppreference.com/w/cpp/container/vector) for this?

test/test-file.cpp
strager, line 62: Nit: The name `temp_file_path` is now misleading. `temp_file_path` refers to a directory, but the variable name implies it refers to a non-directory file.

strager (commented):
Your PR looks like it gets the job done, but at the expense of readability in the tests IMO.

My main concern is the `filesystem::remove_all` call doing who-knows-what if `make_temporary_directory` isn't called. I think we should be forgiving of mistakes.

test/test-file.cpp
```diff
@@ -46,40 +47,51 @@ namespace quick_lint_js {
 namespace {
 filesystem::path make_temporary_directory();
 void write_file(filesystem::path, const std::string &content);
+std::optional<filesystem::path> temp_directory_path;
```
strager, line 50: Blocking: This shouldn't be a global variable. Put this variable in the test_file class.

test/test-file.cpp
```diff
@@ -46,40 +47,51 @@ namespace quick_lint_js {
...
+ protected:
+  void TearDown() override {
+    if (temp_directory_path.has_value()) {
+      filesystem::remove_all(temp_directory_path.value());
+    }
   }
 };
 
 TEST_F(test_file, read_regular_file) {
-  filesystem::path temp_file_path =
```
strager, line 59: Blocking: Undo these variable renames.

strager (changes requested):
Looks good aside from my comments marked 'Blocking'.

test/test-file.cpp
```diff
@@ -50,14 +51,25 @@ void write_file(filesystem::path, const std::string &content);
...
+    if (temp_directory_path.has_value()) {
+      filesystem::remove_all(temp_directory_path.value());
+    }
   }
 };
 
 TEST_F(test_file, read_regular_file) {
   filesystem::path temp_file_path =
       this->make_temporary_directory() / "temp.js";
+
```
strager, line 72: Nit: Undo this addition.

strager (approved):
Do you want to rewrite your commit message? It looks weird.

PR #30, opened 2020-09-18
 Report error when `throw` statement is incomplete or has newline.

masamu:
`expected expression` is reported when `throw` is followed by
 semicolon.
 `unexpected newline` is reported when `throw` is followed by a newline.

### DIFF
--- a/src/include/quick-lint-js/error.h
+++ b/src/include/quick-lint-js/error.h
@@ -72,6 +72,14 @@
       .error(u8"BigInt literal has a leading 0 digit", where))                 \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_expected_expression_before_newline, { source_code_span where; },   \
+      .error(u8"expected expression before newline", where))                   \
+                                                                               \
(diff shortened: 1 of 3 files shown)


src/include/quick-lint-js/error.h
```diff
@@ -142,6 +142,14 @@
       error_unexpected_identifier, { source_code_span where; },                \
       .error(u8"unexpected identifier", where))                                \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_expected_expression, { source_code_span where; },                  \
```
strager, line 146: Nit: Order this alphabetically in the error list.

strager (approved):
Looks good to me. Let me know if I should land this as-is, or if you want to make some changes first.

src/include/quick-lint-js/error.h
masamu: I just pushed an updated version with the fixes.

strager:
Landed as commit 5c81385e8c53e099d3d6633f997fb91f15edbb09. (I rebased and ran clang-format on the sources.)

PR #38, opened 2020-09-23
Combine double error for const variable assignment before declaration

A-thanasios:


src/lint.cpp
```diff
@@ -423,16 +422,23 @@ void linter::propagate_variable_declarations_to_parent_scope() {
...
   switch (var->kind) {
     case variable_kind::_const:
     case variable_kind::_import:
       if (var->is_global_variable) {
         this->error_reporter_->report(
             error_assignment_to_const_global_variable{assignment});
       } else {
-        this->error_reporter_->report(error_assignment_to_const_variable{
-            var->declaration(), assignment, var->kind});
+        if (!is_assigned_before_declaration) {
```
strager, line 433: Nit: I dislike double negatives; line 438 executes if `! !is_assigned_before_declaration`. Invert this `if` to avoid double negatives.

src/include/quick-lint-js/error.h
```diff
@@ -53,6 +53,16 @@
...
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_assignment_to_const_variable_before_its_declaration,               \
+      {                                                                        \
+        identifier declaration;                                                \
+        identifier assignment;                                                 \
+        variable_kind var_kind;                                                \
+      },                                                                       \
+   .error(u8"assignment to const variable before its declaration", assignment) \
+          .note(u8"const variable declared here", declaration))                \
```
strager, line 64: :+1:

strager (approved):
Very nice! I like it.

I normally dislike `bool` parameters, but I think this is okay.

Let me know if you want to make changes, or if I should land this as-is.

A-thanasios:
I felt iffy about the bool parameter, but could not find a better way to determine the order in which an assignment and a declaration has occurred at the function level.

An alternative solution would have been to check for variable_kind _const or _import and not global variable before calling this function where the bool would be true (below). This would avoid having to pass the bool, but duplicating and overriding behavior in the process. Did not seem as good.

        switch (used_var.kind) {
          case used_variable_kind::assignment:
            this->report_error_if_assignment_is_illegal(declared,
                                                        used_var.name);
                error_assignment_before_variable_declaration{
                    .assignment = used_var.name, .declaration = name});
                                                        used_var.name,
                                                        /*is_assigned_before_declaration=*/true);
            break;
 
I have pushed a change for your Nit. I've no further thoughts on this if you don't.

PR #39, opened 2020-09-24
Report missing semicolon for variable declarations and report debugger as invalid binding in let statement

erlliam:
For report missing semicolon for variable declarations, I left out the `if`, `try`, `with`, `class`, and `function` cases. I wasn't too sure if they should all be included. On second thought they probably should be included.

I was going to add more cases, for report invalid binding in let statement, but things are already getting messy. I'm not sure if you should be able to recover from the `if` case (just telling the lexer to skip). For the number and `debugger` case, I believe you should be able to just skip over and continue because it is one token.

I wasn't sure how to title the commit. I wanted to have two separate commits, "Report missing semicolon for variable declarations" and "Report debugger as invalid binding in let statement", but I couldn't figure out how to split the commit into two.

Also not sure about the test cases, they might be redundant?

### DIFF
--- a/src/include/quick-lint-js/parse.h
+++ b/src/include/quick-lint-js/parse.h
@@ -908,6 +908,7 @@ class parser {
         case token_type::number:
           this->error_reporter_->report(
               error_invalid_binding_in_let_statement{this->peek().span()});
+          this->lexer_.skip();
           break;
         default:
           if (first_binding) {
--- a/src/include/quick-lint-js/parse.h
+++ b/src/include/quick-lint-js/parse.h
@@ -372,9 +372,7 @@ class parser {
       case token_type::kw_let:
       case token_type::kw_var:
         this->parse_and_visit_let_bindings(v, this->peek().type);
-        if (this->peek().type == token_type::semicolon) {
-          this->lexer_.skip();
-        }
+        this->consume_semicolon();
         break;
 
       case token_type::kw_function:
--- a/src/include/quick-lint-js/parse.h
+++ b/src/include/quick-lint-js/parse.h
@@ -905,6 +905,13 @@ class parser {
           this->parse_and_visit_binding_element(v, declaration_kind);
           break;
         case token_type::kw_if:
+        case token_type::kw_break:
+        case token_type::kw_continue:
+        case token_type::kw_debugger:
+        case token_type::kw_false:
+        case token_type::kw_null:
+        case token_type::kw_true:
+        case token_type::kw_void:
         case token_type::number:
           this->error_reporter_->report(
               error_invalid_binding_in_let_statement{this->peek().span()});
--- a/test/test-parse.cpp
+++ b/test/test-parse.cpp
@@ -350,6 +350,17 @@ TEST(test_parse, parse_invalid_let) {
                               error_invalid_binding_in_let_statement, where,
                               offsets_matcher(&code, 4, 6))));
   }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"let debugger");
+    parser p(&code, &v);
+    p.parse_and_visit_statement(v);
+    EXPECT_EQ(v.variable_declarations.size(), 0);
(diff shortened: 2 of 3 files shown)


src/include/quick-lint-js/parse.h
```diff
@@ -906,8 +904,10 @@ class parser {
           break;
         case token_type::kw_if:
         case token_type::number:
+        case token_type::kw_debugger:
```
strager, line 907: Nit: Sort these cases alphabetically:
```suggestion
        case token_type::kw_debugger:
        case token_type::kw_if:
        case token_type::number:
```

test/test-parse.cpp
```diff
@@ -1826,5 +1837,71 @@ TEST(test_parse, debugger_statement) {
...
+    parser p(&code, &v);
+    p.parse_and_visit_statement(v);
+    p.parse_and_visit_statement(v);
+    EXPECT_THAT(v.variable_declarations,
+                ElementsAre(spy_visitor::visited_variable_declaration{
+                    u8"x", variable_kind::_let}));
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"console"}));
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
```
strager, line 1854: Minor: The test talks about *reporting* missing semicolons, but these lines show that QLJS should *not* report a missing semicolon in this case. Move this code to a different test, or rename the test.

test/test-parse.cpp
```diff
@@ -1826,5 +1837,71 @@ TEST(test_parse, debugger_statement) {
...
+                    u8"x", variable_kind::_let}));
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"console"}));
+    source_position::offset_type end_of_let_statement =
+        strlen(u8"let x = 2");
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_missing_semicolon_after_expression, where,
+                              offsets_matcher(&code, end_of_let_statement ,
+                                              end_of_let_statement ))));
+  }
```
strager, line 1872: :+1:

test/test-parse.cpp
```diff
@@ -1826,5 +1837,71 @@ TEST(test_parse, debugger_statement) {
...
+    EXPECT_THAT(v.variable_declarations,
+                ElementsAre(spy_visitor::visited_variable_declaration{
+                    u8"x", variable_kind::_const}));
+    source_position::offset_type end_of_const_statement =
+        strlen(u8"const x");
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_missing_semicolon_after_expression, where,
+                              offsets_matcher(&code, end_of_const_statement ,
+                                              end_of_const_statement ))));
+  }
```
strager, line 1888: :+1: Good error reporting!

test/test-parse.cpp
```diff
@@ -1826,5 +1837,71 @@ TEST(test_parse, debugger_statement) {
...
+                ElementsAre(spy_visitor::visited_variable_declaration{
+                    u8"x", variable_kind::_var}));
+    source_position::offset_type end_of_var_statement =
+        strlen(u8"var x");
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_missing_semicolon_after_expression, where,
+                              offsets_matcher(&code, end_of_var_statement ,
+                                              end_of_var_statement ))));
+  }
+}
```
strager, line 1905: Minor: `var` isn't really different from `const`, so I'd delete this almost-duplicate test.

strager (approved):
Looks great. I'll help you split this into multiple commits.

strager:
Looks great! I think it's a bit weird that "Continue parsing after invalid let binding" exists with no test changes, but whatevs.

Issue #46, opened 2020-09-30
Report error for trailing underscores in number literals


strager:
If a number literal has trailing underscores (123_), the syntax is invalid. quick-lint-js should report an error for such constructs. Edit error.h, lex.cpp, and test-lex.cpp to make quick-lint-js help users find this bug.

GauravPatil8778:
i want to work on this

eiffel-fl:
Sorry @GauravPatil8778 I was not aware that we first have to say in the issue that we want to work on this...
Next time, I will first say that I want to work on it to avoid this problem...

PR #48, opened 2020-10-01
Detect trailing underscore(s)

eiffel-fl:
Hi.

First, I do hope you are all fine and the same for your relatives.

I wrote a bit of code which normally solves issue #46 and also extends this by detecting trailing underscores like in "123456__".

I am not a pro C++ developer so maybe I made some mistake.
So, feel free to comment and I will improve the code.

Best regards.

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -841,11 +841,12 @@ template <class Func>
 char8* lexer::parse_digits_and_underscores(Func&& is_valid_digit,
                                            char8* input) noexcept {
   bool has_trailing_underscore = false;
+  const char8* garbage_begin = nullptr;
   while (is_valid_digit(*input)) {
     has_trailing_underscore = false;
     input += 1;
     if (*input == '_') {
-      const char8* garbage_begin = input;
+      garbage_begin = input;
       has_trailing_underscore = true;
       input += 1;
       if (*input == '_') {
@@ -855,15 +856,22 @@ char8* lexer::parse_digits_and_underscores(Func&& is_valid_digit,
           input += 1;
         }
 
-        this->error_reporter_->report(
-            error_number_literal_contains_consecutive_underscores{
-                source_code_span(garbage_begin, input)});
+        if (is_valid_digit(*input)) {
+          this->error_reporter_->report(
+              error_number_literal_contains_consecutive_underscores{
+                  source_code_span(garbage_begin, input)});
+        } else {
+          this->error_reporter_->report(
+            error_number_literal_contains_trailing_underscores{
+              source_code_span(garbage_begin, input)});
+        }
       }
     }
   }
-  if (has_trailing_underscore == true) {
-    // TODO error_reporter: SyntaxError: Numeric separators are not allowed at
-    // the end of numeric literals
+  if (garbage_begin != nullptr && has_trailing_underscore == true) {
+    this->error_reporter_->report(
+      error_number_literal_contains_trailing_underscores{
+        source_code_span(garbage_begin, input)});
   }
   return input;
 }
--- a/test/test-lex.cpp
+++ b/test/test-lex.cpp
@@ -314,6 +314,28 @@ TEST(test_lex, lex_number_with_multiple_groups_of_consecutive_underscores) {
   }
 }
 
+TEST(test_lex, lex_number_with_trailing_underscore) {
+  check_tokens_with_errors(
+    u8"123456_", {token_type::number},
+    [](padded_string_view input, const auto& errors) {
+      EXPECT_THAT(errors,
+                  ElementsAre(ERROR_TYPE_FIELD(
+                    error_number_literal_contains_trailing_underscores,
+                    underscores, offsets_matcher(input, 6, 7))));
+    });
+}
+
+TEST(test_lex, lex_number_with_trailing_underscores) {
+  check_tokens_with_errors(
+    u8"123456___", {token_type::number},
+    [](padded_string_view input, const auto& errors) {
+      EXPECT_THAT(errors,
+                  ElementsAre(ERROR_TYPE_FIELD(
+                    error_number_literal_contains_trailing_underscores,
+                    underscores, offsets_matcher(input, 6, 9))));
+    });
+}
+
 TEST(test_lex, lex_strings) {
   check_single_token(u8R"('hello')", token_type::string);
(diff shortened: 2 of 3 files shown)


src/include/quick-lint-js/error.h
```diff
@@ -145,6 +145,12 @@
       .error(u8"number literal contains trailing underscore",                  \
               underscores))                                                    \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_number_literal_contains_trailing_underscores,                      \
+      { source_code_span underscores; },                                       \
+      .error(u8"number literal contains trailing underscores",                 \
+              underscores))                                                    \
```
strager: Blocking: Let's not make two error types here. Two error types for basically the same thing makes the code more complicated IMO. Let's make only one error type: error_number_literal_contains_trailing_underscores

test/test-lex.cpp
```diff
@@ -299,6 +299,17 @@ TEST(test_lex, lex_number_with_trailing_underscore) {
...
+TEST(test_lex, lex_number_with_trailing_underscores) {
+  check_tokens_with_errors(
+    u8"123456___", {token_type::number},
+    [](padded_string_view input, const auto& errors) {
+      EXPECT_THAT(errors,
+                  ElementsAre(ERROR_TYPE_FIELD(
+                    error_number_literal_contains_trailing_underscores,
+                    underscores, offsets_matcher(input, 6, 9))));
+    });
+}
```
strager, line 311: :+1:

strager (changes requested):
Looks great! But see my "Blocking:" comment.

src/include/quick-lint-js/error.h
eiffel-fl: Thank your for the review!

I modified the code so there is only one error type.
The new error type works for both "123456_" and "123456__".

src/lex.cpp
```diff
@@ -849,15 +850,22 @@ char8* lexer::parse_decimal_digits_and_underscores(char8* input) noexcept {
...
+        }
       }
     }
   }
   if (has_trailing_underscore == true) {
-    // TODO error_reporter: SyntaxError: Numeric separators are not allowed at
-    // the end of numeric literals
+    this->error_reporter_->report(
+        error_number_literal_contains_trailing_underscores{
+            source_code_span(garbage_begin, input)});
```
strager, line 868: Blocking: Clang reports an [error when compiling this code](https://github.com/strager/quick-lint-js/runs/1228354630):

```
[ 49%] Building CXX object src/CMakeFiles/quick-lint-js-lib.dir/lex.cpp.o
/Users/runner/work/quick-lint-js/quick-lint-js/src/lex.cpp:875:30: error: variable 'garbage_begin' may be uninitialized when used here [-Werror,-Wconditional-uninitialized]
            source_code_span(garbage_begin, input)});
                             ^~~~~~~~~~~~~
/Users/runner/work/quick-lint-js/quick-lint-js/src/lex.cpp:845:29: note: initialize the variable 'garbage_begin' to silence this warning
  const char8* garbage_begin;
                            ^
                             = nullptr
1 error generated.
```

[GCC reports a similar error.](https://github.com/strager/quick-lint-js/runs/1228354758) These errors cause CI builds to fail. (CI builds are misconfigured on pull requests (my fault), so I didn't see this issue until I tried merging.)

I think this diagnostic is a false positive (i.e. there's no bug in your code), but we should fix the diagnostic anyway.

Idea: Instead of having two variables (`garbage_begin` and `has_trailing_underscore`), we could use a single variable. I'm thinking `garbage_begin` can be `nullptr` initially, and we check `if (garbage_begin) {` on line 365.

strager (changes requested):
Looks great, but compilers are confused. See my 'Blocking' comment.

src/lex.cpp
eiffel-fl: I do not know have an error with my compiler (`gcc (Debian 8.3.0-6)`) but I agree with you on the false positive.
Indeed, `garbage_begin` can be uninitialized after the loop.
But if we enter the if after the loop, `garbage_begin` had been set since it is set before `has_trailing_underscore` which is the `if` condition.

I think there is a problem with your solution.
Indeed, if we only condition the outer loop `if` with `garbage_begin` a number like 123__456 will print both consecutive and trailing underscores error.
To make the compiler happy and having all cases well covered I add `garbage_begin != nullptr` to the condition of outer loop if.

src/lex.cpp
```diff
@@ -842,11 +842,12 @@ void lexer::parse_number() {
 
 char8* lexer::parse_decimal_digits_and_underscores(char8* input) noexcept {
   bool has_trailing_underscore = false;
+  const char* garbage_begin = nullptr;
```
strager, line 845: Blocking: This variable's type should be `const char8*`, not `const char*`. It might work on your machine because your compiler doesn't support `char8_t` (therefore `char8` is the same as `char`), but on CI and on my machine, `char8` and `char` are different types. This leads to a build failure: https://travis-ci.org/github/strager/quick-lint-js/builds/734270869

eiffel-fl: My bad... I forgot to check CI result and only thought that if it is correct on my machine it is universally correct... I will change this habit now.
This should now be corrected!

src/lex.cpp
```diff
@@ -856,15 +857,22 @@ char8* lexer::parse_decimal_digits_and_underscores(char8* input) noexcept {
...
+            error_number_literal_contains_trailing_underscores{
+              source_code_span(garbage_begin, input)});
+        }
       }
     }
   }
-  if (has_trailing_underscore == true) {
-    // TODO error_reporter: SyntaxError: Numeric separators are not allowed at
-    // the end of numeric literals
+  if (garbage_begin != nullptr and has_trailing_underscore == true) {
```
strager, line 872: It looks like the MSVC compiler doesn't like `and`: https://github.com/strager/quick-lint-js/runs/1236834344

Let's use `&&` instead (or figure out how to make MSVC understand `and`).

eiffel-fl: Normally it's corrected!
Congratulations for finding that MSVC does not like `and` in its (really verbose) logs.

strager:
Landed as commit c242fc86952ff2111a6fe27da44dac50e35a8907. Thanks!

eiffel-fl:
Perfect!
Thank you!

PR #49, opened 2020-10-01
Returns EXIT_FAILURE if an error was reported

eiffel-fl:
Hi.

First, I do hope you are all fine and the same for your relatives.

I modified linter code to first uses EXIT_SUCCESS and EXIT_FAILURE macros instead of 0 and 1.
I do not know if this a good practice in C++ but in C this a good one (see [this post](https://stackoverflow.com/a/8868139)).

Then I added a static variable to error_exporter class.
This variable can be seed as a fuse, it is initialized as false and set to true by text_reporter::report.
Finally, it is tested in main and if is true main returns EXIT_FAILURE.

For now, I only modified text_reporter::report and not other reporters.
Please let me know if I should do it or if there is problem with the code.
Indeed, I am not very skilled in C++ so using static variable is maybe a bad approach.

Best regards.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -104,17 +104,17 @@ int main(int argc, char **argv) {
   quick_lint_js::options o = quick_lint_js::parse_options(argc, argv);
   if (o.help) {
     quick_lint_js::print_help_message();
-    return 0;
+    return EXIT_SUCCESS;
   }
   if (!o.error_unrecognized_options.empty()) {
     for (const auto &option : o.error_unrecognized_options) {
       std::cerr << "error: unrecognized option: " << option << '\n';
     }
-    return 1;
+    return EXIT_FAILURE;
   }
   if (o.files_to_lint.empty()) {
     std::cerr << "error: expected file name\n";
-    return 1;
+    return EXIT_FAILURE;
   }
 
   quick_lint_js::any_error_reporter reporter =
@@ -129,7 +129,7 @@ int main(int argc, char **argv) {
   }
   reporter.finish();
 
-  return 0;
+  return EXIT_SUCCESS;
 }
 
 namespace quick_lint_js {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -20,7 +20,7 @@
 #include <iomanip>
 #include <iostream>
 #include <quick-lint-js/char8.h>
-#include <quick-lint-js/error.h>
+#include <quick-lint-js/error-tape.h>
 #include <quick-lint-js/file.h>
 #include <quick-lint-js/language.h>
 #include <quick-lint-js/lex.h>
@@ -44,9 +44,11 @@ class any_error_reporter {
   static any_error_reporter make(output_format format) {
     switch (format) {
     case output_format::gnu_like:
-      return any_error_reporter(text_error_reporter(std::cerr));
+      return any_error_reporter(
+          error_tape<text_error_reporter>(text_error_reporter(std::cerr)));
     case output_format::vim_qflist_json:
-      return any_error_reporter(vim_qflist_json_error_reporter(std::cout));
+      return any_error_reporter(error_tape<vim_qflist_json_error_reporter>(
+          vim_qflist_json_error_reporter(std::cout)));
     }
     QLJS_UNREACHABLE();
   }
@@ -55,40 +57,45 @@ class any_error_reporter {
     std::visit(
         [&](auto &r) {
           using reporter_type = std::decay_t<decltype(r)>;
-          if constexpr (std::is_base_of_v<vim_qflist_json_error_reporter,
-                                          reporter_type>) {
-            r.set_source(input, file.path, file.vim_bufnr);
+          if constexpr (std::is_base_of_v<
+                            error_tape<vim_qflist_json_error_reporter>,
+                            reporter_type>) {
+            r.get_reporter()->set_source(input, file.path, file.vim_bufnr);
           } else {
-            r.set_source(input, file.path);
+            r.get_reporter()->set_source(input, file.path);
           }
         },
-        this->reporter_);
+        this->tape_);
   }
 
   error_reporter *get() noexcept {
-    return std::visit([](error_reporter &r) { return &r; }, this->reporter_);
+    return std::visit([](error_reporter &r) { return &r; }, this->tape_);
+  }
+
+  bool get_error() noexcept {
+    return std::visit([](auto &r) { return r.get_error(); }, this->tape_);
   }
 
   void finish() {
     std::visit(
         [&](auto &r) {
           using reporter_type = std::decay_t<decltype(r)>;
-          if constexpr (std::is_base_of_v<vim_qflist_json_error_reporter,
-                                          reporter_type>) {
-            r.finish();
+          if constexpr (std::is_base_of_v<
+                            error_tape<vim_qflist_json_error_reporter>,
+                            reporter_type>) {
+            r.get_reporter()->finish();
           }
         },
-        this->reporter_);
+        this->tape_);
   }
 
  private:
-  using reporter_variant =
-      std::variant<text_error_reporter, vim_qflist_json_error_reporter>;
+  using tape_variant = std::variant<error_tape<text_error_reporter>,
+                                    error_tape<vim_qflist_json_error_reporter>>;
 
-  explicit any_error_reporter(reporter_variant &&reporter)
-      : reporter_(std::move(reporter)) {}
+  explicit any_error_reporter(tape_variant &&tape) : tape_(tape) {}
 
-  reporter_variant reporter_;
+  tape_variant tape_;
 };
 
 void process_file(padded_string_view input, error_reporter *,
@@ -128,6 +135,10 @@ int main(int argc, char **argv) {
                                 o.print_parser_visits);
   }
   reporter.finish();
+
+  if (reporter.get_error() == true) {
+    return EXIT_FAILURE;
+  }
 
   return EXIT_SUCCESS;
 }
--- a/src/quick-lint-js/error-tape.h
+++ b/src/quick-lint-js/error-tape.h
@@ -1,0 +1,65 @@
+// quick-lint-js finds bugs in JavaScript programs.
+// Copyright (C) 2020  Matthew Glazar
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef QUICK_LINT_JS_ERROR_TAPE_H
+#define QUICK_LINT_JS_ERROR_TAPE_H
+
+#include <quick-lint-js/error.h>
+#include <quick-lint-js/text-error-reporter.h>
+#include <quick-lint-js/vim-qflist-json-error-reporter.h>
+
+namespace quick_lint_js {
+template <typename T>
+class error_tape final : public error_reporter {
+ public:
+  error_tape(T reporter) : reporter_(reporter), error_(false) {}
+
+  T *get_reporter() { return &(this->reporter_); }
+
+  bool get_error(void) { return this->error_; }
+
+  void set_error(void) { this->error_ = true; }
+
+#define QLJS_ERROR_TYPE(name, struct_body, format) \
+  void report(name e) override final {             \
+    set_error();                                   \
+    reporter_.report(e);                           \
+  }
+  QLJS_X_ERROR_TYPES
+#undef QLJS_ERROR_TYPE
+
+  void report_fatal_error_unimplemented_character(
+      const char *qljs_file_name, int qljs_line, const char *qljs_function_name,
+      const char8 *character) override final {
+    reporter_.report_fatal_error_unimplemented_character(
+        qljs_file_name, qljs_line, qljs_function_name, character);
+  }
+
+  void report_fatal_error_unimplemented_token(
+      const char *qljs_file_name, int qljs_line, const char *qljs_function_name,
+      token_type type, const char8 *token_begin) override final {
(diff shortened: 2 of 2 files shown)


strager:
> For now, I only modified text_reporter::report and not other reporters.
> Please let me know if I should do it or if there is problem with the code.

The `error_reporter::error` should be set if using *any* `--output-format`. The fact that we need to modify most `error_reporter` derived classes to make this work is a code smell. Instead, I think we should create a single `error_reporter` class designed to set the  `error_reporter::error` variable. (My idea reminds me of the Decorator design pattern.)

eiffel-fl:
First, thank you for the review!
I do agree with your comment, if we modify each subclass we will add the same code for each subclasses and this is not a good practice.
This pull request was sort of a draft.

I checked the decorator design pattern, it is elegant and seems to solve our problem.
I will try to implement that!

eiffel-fl:
I tried to implement your solution with the decorator but I struggle with C++ and templating things...

In my code, I created a class `error_decorator` which inherits from `error_reporter` and has an `error_reporter` as member.
Then I tried 2 different approaches:
1. Putting a `format` function into `error_reporter` but I failed because a template function cannot be pure virtual so I cannot write:
`virtual error_formatter format() = 0`
2. Using a `std::variant`, like what you did in `any_reporter`, in `error_decorator` and use `std::visit` to call `format`.
But here I also have the same problem because it seems that `std::visit` does not like returning template type...

Do you have a tip so I can continue to write the code? For now I will stay this PR as a draft.

strager:
@eiffel-fl The `error_reporter` class says nothing about `format`. Why would we call `format` from `error_decorator`? I'm confused. Can you share what you've written so far so I can understand how the code is evolving?

eiffel-fl:
Here is the [error_decorator.cpp](https://framabin.org/p/?dc303bd6668088ac#ZsQnEWBwXrxLKirs7dCaoi5qjcXI7MqMH9SrbACGUjo=) and its [.h](https://framabin.org/p/?0ab9621e31674cb4#/6K7zdPPEX9asruJ/jvvBqslTvFLacFIG3M3kW9t1/s=) for my second attempt.

I think I misunderstood something about the decorator pattern.
But for me we call `format` on the `error_reporter` from the `error_decorator`.
Let me know if I am totally wrong, I will correct the code.

strager:
> Here is the error_decorator.cpp and its .h for my second attempt.

Instead of calling `format_error` or `r.format()`, call `r.report()`:

```cpp
void error_decorator::report(name e) {
  error_reporter::error = true;
  std::visit([&](auto &r) {
    return r.report(e);
  }, reporter_);
}
```

eiffel-fl:
I was looking for complicated solutions while I should have think simple...
Thank you for the tip! Normally the code is now OK!

If I wrote something poorly object please let me know and I will rewrite it correctly.

src/quick-lint-js/error.h
```diff
@@ -278,6 +278,8 @@ class error_reporter {
   static void write_fatal_error_unimplemented_token(
       const char *qljs_file_name, int qljs_line, const char *qljs_function_name,
       token_type, const char8 *token_begin, const locator *, std::ostream &);
+
+  static bool error;
```
strager, line 282: Blocking: This variable belongs in `error_decorator`, not here in `error_reporter`.

Almost-blocking: Also, this variable should not be global. It should be tied to a particular `error_decorator` instance.

src/quick-lint-js/error_decorator.h
```diff
@@ -0,0 +1,54 @@
...
+#include <quick-lint-js/error.h>
+#include <quick-lint-js/text-error-reporter.h>
+#include <quick-lint-js/vim-qflist-json-error-reporter.h>
+#include <variant>
+
+namespace quick_lint_js {
+class error_decorator : public error_reporter {
+ public:
+  using reporter_variant = std::variant<text_error_reporter,
+    vim_qflist_json_error_reporter>;
```
strager, line 29: I think `error_decorator` should not be coupled to any particular `error_reporter` derived class. Ideas:

* Make the decorated `error_reporter` type a template parameter.
* Make the decorated `error_reporter` object live outside `error_decorator`, and give `error_decorator` a pointer to the decorated `error_reporter` object.

src/error_decorator.cpp
```diff
@@ -0,0 +1,43 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#include <quick-lint-js/error_decorator.h>
+#include <variant>
+
+namespace quick_lint_js {
+#define QLJS_ERROR_TYPE(name, struct_body, format_call)   \
+void error_decorator::report(name e) {                    \
+  error_reporter::error = true;                           \
+  std::visit([&](auto r) { r.report(e); }, reporter_);    \
```
strager, line 24: Blocking: Do not copy the underlying reporter. Use the existing reporter object:
```suggestion
  std::visit([&](auto &r) { r.report(e); }, reporter_);    \
```

src/quick-lint-js/error_decorator.h
```diff
@@ -0,0 +1,54 @@
+// quick-lint-js finds bugs in JavaScript programs.
```
strager, line 1: Nit: Call this file `error-decorator.h` instead of `error_decorator.h`

src/quick-lint-js/error_decorator.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef QUICK_LINT_JS_ERROR_DECORATOR_H
+#define QUICK_LINT_JS_ERROR_DECORATOR_H
+
+#include <quick-lint-js/error.h>
+#include <quick-lint-js/text-error-reporter.h>
+#include <quick-lint-js/vim-qflist-json-error-reporter.h>
+#include <variant>
+
+namespace quick_lint_js {
+class error_decorator : public error_reporter {
```
strager, line 26: Have you thought about other names for this class? Some ideas from Twitch chat:

* ErrorDecoratorCheckToSeelfErrorFactory
* errorTape
* errno_but_different
* ErrorWriter
* ErrorReader
* error_statistics_collector
* ErrorDelegator
* ErrorSeeker
* ErrorManager
* ErrorTracker
* error_state
* error
* error_status
* ErrorCounter
* Error_God

strager (changes requested):
Looking good so far!

See my comments marked 'Blocking'.

src/quick-lint-js/error.h
eiffel-fl: I put this variable as a member in `error_decorator`.

src/quick-lint-js/error_decorator.h
eiffel-fl: I gave a `error_reporter *` to `error_decorator`.
I had to modify heavily the `any_error_reporter` class in `main.cpp` to do that so let me know if my modifications are bad or not so C++ (I am not really used to template).

src/error_decorator.cpp
eiffel-fl: I totally modified this snippet so `std::visit` is not used anymore.

src/quick-lint-js/error_decorator.h
eiffel-fl: I changed the name of the added files and their new names follow the naming convention;

src/quick-lint-js/error_decorator.h
eiffel-fl: I choose to call it `error_tape` because I like the `errorTape` name but the code is mainly snake case.

src/quick-lint-js/error-tape.h
```diff
@@ -0,0 +1,65 @@
...
+namespace quick_lint_js {
+class error_tape : public error_reporter {
+ public:
+  error_tape(error_reporter *reporter)
+    : reporter_(reporter),
+    error_ (false) {}
+
+  ~error_tape(){
+    delete reporter_;
+  }
```
strager, line 33: Almost-blocking: If the user of `error_tape` wants to transfer ownership of the `error_reporter` to the `error_tape`, the constructor should accept a `std::unique_ptr` instead of a raw pointer.

Blocking: Within `error_tape`, we should store the pointer in a `std::unique_ptr` instead of storing a raw pointer. A raw pointer is bug-prone here. There is a risk of double delete because `error_tape` has a copy constructor which copies the pointer.

src/main.cpp
```diff
@@ -44,51 +44,48 @@ class any_error_reporter {
...
-            r.set_source(input, file.path, file.vim_bufnr);
-          } else {
-            r.set_source(input, file.path);
-          }
-        },
-        this->reporter_);
+    text_error_reporter *text_reporter = dynamic_cast<
+      text_error_reporter *>(this->tape_.get_reporter());
+    vim_qflist_json_error_reporter *vim_reporter = dynamic_cast<
+      vim_qflist_json_error_reporter *>(this->tape_.get_reporter());
```
strager, line 58: I want to avoid `dynamic_cast` if possible. In this case, I think a separate `bool` or `enum` variable would be better than `dynamic_cast`. Or we could use `std::variant` holding pointers.

Alternatively, we could make `error_tape` a template, and make the underlying `error_reporter` type a template parameter. Then we could keep the old style of using `std::variant`. For example:
```cpp
using reporter_variant =
    std::variant<error_tape<text_error_reporter>, error_tape<vim_qflist_json_error_reporter>>;
```
If you go down the template route, I recommend having the `error_tape` construct the underlying `error_reporter`, and perfect-forwarding the constructor arguments.

src/quick-lint-js/error-tape.h
```diff
@@ -0,0 +1,65 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef QUICK_LINT_JS_ERROR_DECORATOR_H
+#define QUICK_LINT_JS_ERROR_DECORATOR_H
```
strager, line 18: Nit: Update include guards to match the file's name:
```suggestion
#ifndef QUICK_LINT_JS_ERROR_TAPE_H
#define QUICK_LINT_JS_ERROR_TAPE_H
```

strager (changes requested):

src/quick-lint-js/error-tape.h
eiffel-fl: I corrected this on my side.

src/main.cpp
eiffel-fl: On my side, I tried to use template, but I had trouble converting `reporter_variant` to `error_reporter *`.
I can understand template but `std::variant` is a bit obscure for me... I will try to implement what you suggested, this should help me converting the variant to the true type.

src/quick-lint-js/error-tape.h
eiffel-fl: I think I will go with a template way so I will remove all about pointer into the code.

src/main.cpp
eiffel-fl: I finally managed to go with a complete template way.
Please let me know if there are still problems (e.g. I had linking errors when template functions were defined in `error-tape.cpp` so I defined them in `error-tape.h`).

src/main.cpp
```diff
@@ -55,40 +55,52 @@ class any_error_reporter {
...
+        this->tape_);
   }
 
   error_reporter *get() noexcept {
-    return std::visit([](error_reporter &r) { return &r; }, this->reporter_);
+    try {
+      return &std::get<error_tape<text_error_reporter>>(this->tape_);
+    } catch(std::bad_variant_access&) {
+      return &std::get<error_tape<vim_qflist_json_error_reporter>>(this->tape_);
+    }
```
strager, line 73: Why don't we use `std::visit` like with the original code?
```suggestion
    return std::visit([](error_reporter &r) { return &r; }, this->tape_);
```

src/main.cpp
```diff
@@ -55,40 +55,52 @@ class any_error_reporter {
...
+  }
+
+  bool get_error() noexcept {
+    try {
+      return std::get<error_tape<text_error_reporter>>(this->tape_).get_error();
+    } catch(std::bad_variant_access&) {
+      return std::get<error_tape<vim_qflist_json_error_reporter>>(this->tape_).get_error();
+    }
```
strager, line 81: Same with `get`; use `std::visit` if possible.

strager (approved):
Looks great!

src/main.cpp
eiffel-fl: This is normally corrected.

src/main.cpp
eiffel-fl: I would like to use `std::visit`.
But the code would be:
`return std::visit([](error_tape<> &r) { return r.get_error(); }, this->tape_);`
And this does not compile because error_tape needs a type as template and I do not see which type to use...

strager: Ah, right. And `return std::visit([](error_reporter &r) { return r.get_error(); }, this->tape_);` won't work because `error_reporter::get_error` isn't a thing.

Making the lambda a template should work though:
```cpp
return std::visit([](auto &r) { return r.get_error(); }, this->tape_);
```

eiffel-fl: This is now corrected!

I should have think to use `auto`... This is when you write code like that you understand the ++ of C++ compared to C :laughing:.

strager:
One last issue (hopefully...): clang-format. See the failing check on this PR: https://github.com/quick-lint/quick-lint-js/pull/49/checks?check_run_id=1252624418

Let me know if the instructions are unclear or you need help getting clang-format working.

eiffel-fl:
Normally this should be good!
I did not understand this check, I though it was checking if the code was modified with `git diff`.
Do you think it is possible to add a `clang-format` target to `cmake` so it does automatically?

strager:
>  Do you think it is possible to add a `clang-format` target to `cmake` so it does automatically?

That is possible. However, I think it's confusing if *building* your code *modifies* your code.

>  I did not understand this check, I though it was checking if the code was modified with `git diff`.

The job checks if running `tools/format` changes any files.

eiffel-fl:
> That is possible. However, I think it's confusing if _building_ your code _modifies_ your code.

I do not find it odd if there is a make` target called `format` to run `./tools/format` and for example another one called `check` to run `cppcheck`.
 
> The job checks if running `tools/format` changes any files.

I read precisely the CI and finally understand the check.

Normally the code passes now all check so you can merge it whenever you want!

eiffel-fl:
Perfect! Thank you!

Issue #50, opened 2020-10-02
Parse <!-- --> comments


strager:
JavaScript allows `<!--` and `-->` for comments. quick-lint-js doesn't understand such comments. Edit lex.cpp and test-lex.cpp. See skip_block_comment and skip_line_comment for related code.

- [x] `<!--`
- [ ] `-->`

strager:
@hsteinmueller is working on this task. GitHub won't let me assign them for some reason, though.

strager:
`-->` implemented in commit 751c68f1e8de56091d66d057f90ebeb10dfc0080.

PR #51, opened 2020-10-02
Parse <!-- --> comments

hsteinmueller:
#50 
I implemented it the way you suggested, but according to [this stackover answer](https://stackoverflow.com/questions/5780047/html-comments-in-a-javascript-block/#answer-18638833) and [the specification](https://tc39.es/ecma262/#sec-html-like-comments) this seems to be a bit more complex.
Also I hope using the conditional operator like this is fine.


strager:
> I implemented it the way you suggested, but according to this stackover answer and the specification this seems to be a bit more complex.

Implement according to the standard, not my informal description. =]

src/lex.cpp
```diff
@@ -1157,8 +1160,12 @@ void lexer::skip_whitespace() {
 QLJS_WARNING_POP
 
 void lexer::skip_block_comment() {
```
strager, line 1162: Blocking: Hmm. Let's instead create a separate `skip_html_comment` function. Three reasons:

* Performance. The extra checks for `-` inside the loop probably matter when parsing block comments. (I can check the performance impact if you want.)
* Readability. Block comment parsing is complicated because performance matters. I suspect HTML comments are rare, so performance is less important. We can implement less-optimized (thus probably more readable) code for HTML comments. For example, see what `skip_block_comment` looked like before commit fb2b951bf605dd928cf97d00832c5011f1858336: https://github.com/strager/quick-lint-js/blob/be4ef473ea1127617393fc038cd2da863f967849/src/lex.cpp#L762-L787
* Correctness. HTML comments have different rules than block comments (as you noticed). I suspect implementing these rules correctly in the same function as `skip_block_comment` will become really difficult. This is only a hunch, though; maybe you're implementing all the rules already.

src/lex.cpp
```diff
@@ -1229,14 +1242,14 @@ void lexer::skip_block_comment() {
...
 
 found_comment_end:
-  this->input_ = c + 2;
+  this->input_ = html_comment ? c + 3 : c + 2;
   this->skip_whitespace();
   return;
 
   QLJS_UNREACHABLE();
 found_end_of_file:
   this->error_reporter_->report(error_unclosed_block_comment{
```
strager, line 1251: Minor: This error type should be different for HTML comments, e.g. `error_unclosed_html_comment`.

strager (changes requested):
See my 'Blocking:' comment.

A quick look at the StackOverflow answer and your tests tell me that we're implementing HTML comments incorrectly. I didn't stare closely at the spec though.

hsteinmueller:
Ok, so I did some more research.
Apparently `<!-- -->` doesn't behave like a block comment.
`<!--` is treated like `//`, this is implemented now in `skip_html_line_comment()`

`-->` is treated like `//` if at the beginning of a line (with optional whitespaces) or after a closing '*/'
e.g. this is valid:
```
/*
 */ --> comment
--> comment
```
this is not:
```
<!-- comment
var x = 1; -->
```
I'm not really sure how to implement this behavior, especially since something like `while (max --> min)` is also valid.

strager:
> `-->` is treated like `//` if at the beginning of a line ...
>
> ...
>
> I'm not really sure how to implement this behavior

`lexer::skip_whitespace` sets `this->last_token_.has_leading_newline`. If `this->last_token_.has_leading_newline` is set, and you see `-->`, then you know the `-->` is at the beginning of a line (or follows a block comment containing a newline).

... Except, for the very first line in a file, I think `this->last_token_.has_leading_newline` is false. I'm not sure how to handle this case. Perhaps we need another bool variable in the lexer, like `bool is_first_line_`. Or perhaps we can scan backwards looking for a newline character... Gross.

This task looks harder than I initially thought. I should have peeked at the spec before marking this good-first-issue. x_x

strager:
To manage the complexity, we could parse `<!--` comments now, and leave `-->` comments for a later PR.

src/lex.cpp
```diff
@@ -256,7 +256,10 @@ void lexer::parse_current_token() {
     break;
 
   case '<':
-    if (this->input_[1] == '=') {
+    if (this->input_[1] == '!') {
+      this->skip_html_line_comment();
```
strager, line 260: Almost-blocking: This would break code like `x<!y` right? (I think `x<!y` should be parsed as `x` `<` `!` `y`, not as `x` followed by an HTML line comment.) Maybe `x<!-y` too. (Perhaps these cases deserves some tests!)

src/lex.cpp
```diff
@@ -1241,6 +1244,23 @@ void lexer::skip_block_comment() {
...
+    int newline_size = this->newline_character_size(c);
+    if (newline_size > 0) {
+      this->input_ = c + newline_size;
+      this->skip_whitespace();
+      break;
+    }
+    if (*c == u8'\0' && this->is_eof(c)) {
+      this->input_ = c;
+      break;
+    }
```
strager, line 1260: Minor: Hmm, is this loop body exactly the same as `lexer::skip_line_comment`'s? Perhaps we should factor the common code. Maybe later.

strager (approved):
Looks great! Thought we might have broken working code. Double-check please.

Please give me the green light to merge this (or make changes if you wish).

src/lex.cpp
hsteinmueller: Yes, this would break.
I tried fixing it by adding more checks for `-` and `-`, is that ok?
Also added 2 tests.

src/lex.cpp
hsteinmueller: Added the check for `<!--` to `skip_line_comments()`
and adjusted `skip_length`.

hsteinmueller:
> Looks great! Thought we might have broken working code. Double-check please.
> 
> Please give me the green light to merge this (or make changes if you wish).

I tried to implement your suggestions.
You can merge if you want to or make more suggestions. :blush:

src/lex.cpp
```diff
@@ -1243,8 +1248,11 @@ void lexer::skip_block_comment() {
 
 void lexer::skip_line_comment() {
   QLJS_ASSERT((this->input_[0] == '/' && this->input_[1] == '/') ||
-              (this->input_[0] == '#' && this->input_[1] == '!'));
-  for (char8* c = this->input_ + 2;; ++c) {
+              (this->input_[0] == '#' && this->input_[1] == '!') ||
+              (this->input_[0] == '<' && this->input_[1] == '!' &&
+               this->input_[2] == '-' && this->input_[3] == '-'));
+  uint8_t skip_length = this->input_[0] == '<' ? 4 : 2;
+  for (char8* c = this->input_ + skip_length;; ++c) {
```
strager, line 1255: For future reference: I like to avoid unsigned types like `uint8_t` when doing math. The conversion rules often bite you! (Doesn't matter here, though.)
```suggestion
  int skip_length = this->input_[0] == '<' ? 4 : 2;
  for (char8* c = this->input_ + skip_length;; ++c) {
```

strager (approved):

strager:
Landed as commit c149716daaf0227c0b14f491f85ebcc641505f06. (I squashed all commits in this PR, rebased, and ran clang-format.)

PR #66, opened 2020-10-10
Run CI on pull requests

### LINKED ISSUE #59: Run CI on pull requests
The GitHub Actions workflows for building and testing only run on my branches, not on pull requests. Make them run on pull requests, too.

Previous attempt: https://github.com/strager/quick-lint-js/commit/e55b699d22a74140a31f9e9896657dc71ec0a391

erlliam:
Resolves #59 
GitHub actions should now work with pull requests. I tested it out on my own repository.

[pull_request opened](https://github.com/erlliam/quick-lint-js/pull/2)
[pull_request synchronized](https://github.com/erlliam/quick-lint-js/pull/1)

It confused me at first because the push event was also triggered, but it makes sense since I own the repository (it was a push and pull request).

### DIFF
--- a/.github/workflows/build-and-test-web-demo.yml
+++ b/.github/workflows/build-and-test-web-demo.yml
@@ -15,7 +15,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 name: build and test web demo
-on: [push]
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   build:
@@ -25,8 +28,14 @@ jobs:
       - uses: mymindstorm/setup-emsdk@v6
         with:
           version: 2.0.4
-      - name: check out
-        uses: actions/checkout@v1
+      - name: checkout
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: checkout
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: C++ configure
         run: emcmake cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -15,7 +15,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 name: build and test
-on: [push]
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   build:
@@ -56,8 +59,15 @@ jobs:
         ) }}
 
     steps:
-      - name: check out
-        uses: actions/checkout@v1
+      - name: checkout
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: checkout
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
       - name: configure
         run: |
           env | grep '^CMAKE\|^QUICK_LINT_JS' | sort
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -15,14 +15,23 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 name: check format
-on: [push]
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   check_format:
     name: check format
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
       - run: clang-format --version
       - run: ./tools/format
       - run: git diff --exit-code
--- a/.github/workflows/simple-build.yml
+++ b/.github/workflows/simple-build.yml
@@ -15,7 +15,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 name: simple build
-on: [push]
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
 
 jobs:
   build:
@@ -34,7 +37,14 @@ jobs:
       CXXFLAGS: ${{ matrix.toolchain.CXXFLAGS }}
 
     steps:
-      - name: check out
-        uses: actions/checkout@v1
+      - name: checkout
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: checkout
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
       - name: build
         run: '"${CXX}" ${CXXFLAGS} -o quick-lint-js -I src -I vendor/boost src/*.cpp vendor/boost/libs/container/src/*.cpp vendor/boost/libs/container/src/alloc_lib.c'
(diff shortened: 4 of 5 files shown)


erlliam:
Well you can see the GitHub actions working in this pull request.
Although it seems like using the branch ci to put my changes on top of was the wrong thing to do.
I think I will redo the changes using the actual master branch's workflow folder.

strager:
TL;DR: Rebase your commits atop master.

It looks like GitHub is trying to be super smart. I think this is what it's doing:

1. Look for workflow files in commit 4079266 (PR's HEAD). See which ones need to be triggered.
2. Git-merge commit 4079266 (PR's HEAD) with commit cfe2edc (PR's destination; strager:master).
   1. For .github/workflows/simple-build.yml, from commit 4079266's perspective, this changes the command to run.
   2. For .github/workflows/simple-build.yml, from commit cfe2edc's perspective, this adds a different checkout step.
3. Reload the workflow files from the merge commit.
4. Run the workflow.
   1. The merged workflow file says to check out commit 4079266 (PR's HEAD). This commit has the file `src/include/quick-lint-js/assert.h`.
   2. The merged workflow file says to run with `-I src` (instead of `-I src/include`). This command assumes that `src/quick-lint-js/assert.h` exists, but that file doesn't exist at commit 4079266 (PR's HEAD).

GitHub's philosophy is to run GitHub Actions on the PR's HEAD *merged with* the PR's destination. I want GitHub actions to run on the PR's HEAD *unmolested*. That's why we need extra code for the checkout step. But it looks like GitHub's philosophy applies to workflow files too, and we can't influence that.

If you rebase your changes atop master, I think the errors will go away. The fundamental problem still exists, but it will affect 5% of PRs or fewer.

.github/workflows/build-and-test.yml
```diff
@@ -56,8 +59,14 @@ jobs:
...
-      - name: check out
-        uses: actions/checkout@v1
+      - name: push check out
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: pull request check out
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
strager, line 69: Could we factor this into a reusable action? I think we can create a [composite run steps action](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-composite-run-steps-action) and [put that action inside `.github/actions/`](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/about-actions#choosing-a-location-for-your-action)

.github/workflows/check-format.yml
```diff
@@ -16,19 +16,26 @@
...
           ref: ${{ github.event.pull_request.head.sha }}
-      - run: clang-format --version
-      - run: ./tools/format
-      - run: git diff --exit-code
+      - name: clang format version
+        run: clang-format --version
+      - name: format script
+        run: ./tools/format
+      - name: git diff
+        run: git diff --exit-code
```
strager, line 41: Why'd you refactor this?

.github/workflows/build-and-test.yml
erlliam: I was thinking about doing that once I started working on the changes

.github/workflows/check-format.yml
erlliam: I was just trying to keep consistency with the other changes I made.

erlliam:
Well after much trail and error, I don't think it's possible to reuse steps for GitHub actions. This person seems to have [a similar experience.](https://stackoverflow.com/questions/59757355/reuse-portion-of-github-action-across-jobs) The alternative to the code duplication is probably to modify the actions/checkout action.

strager:
*sigh*

Let's just duplicate the code then.

erlliam:
Alright I've rebased, should be good to go.

strager:
It looks like our GitHub Actions workflows are already running on pull requests. However, the workflow runs don't show up on GitHub's UI for PRs.

Example: In this PR (#66), clicking the green checkmark on the latest commit reveals a bunch of "(pull_request)" workflow runs. If you look at the [PR's source branch](https://github.com/erlliam/quick-lint-js/tree/ci), clicking the green checkmark reveals a bunch of "(push)" runs. The runs have different run IDs.

Your PR appears to cause GitHub Actions to duplicate its work. The "(pull)" and "(pull_request)" workflow runs seem to be independent; they have their own IDs:

* https://github.com/erlliam/quick-lint-js/runs/1240884070
* https://github.com/quick-lint/quick-lint-js/runs/1240884145

I think we need to do one of the following:

1. Make the "(pull)" runs appear in the PR UI, or
2. Disable the "(pull)" runs for PRs.

I think we can approximate option 2 by only enabling "(pull)" for the `quick-lint/quick-lint-js` repository. But this means that contributors won't get CI feedback unless they create a PR.

What do you think, @erlliam ?

erlliam:
Option one is ideal, but it sounds like it's out of our hands.
It is interesting how for my test PRs, the push event is displayed in the UI. I assume this happens because I own the repository. Strange.

Option two sounds less than ideal. This would be the last resort that no one wants.

I would consider this a problem with GitHub, considering the push event is displayed "if you own the repository".

strager:
> considering the push event is displayed "if you own the repository".

I think you're right. I think this is the logic of GitHub Actions:

* When you push to your fork, *your fork*'s workflows run. That means *you* are billed for your own use of GitHub Actions workers.
* When you submit a PR to my fork, *my fork*'s workflows run. That means *I* am billed for *your* use of GitHub Actions workers.
* When looking at a PR on someone's fork, everyone sees workflows owned by that fork, and not workflows owned by other forks.
  * If you submit a PR to my fork, *my fork*'s workflows are visible (e.g. "(pull_request)"), but not your fork's workflows (e.g. "(push)").
  * If you submit a PR to your own fork, all of your fork's workflows are visible, including "(pull_request)" and "(push)".

In other words:

where is branch | where is PR   | workflows run on your fork | workflows run on my fork | runs visible on PR
--------------- | ------------- | -------------------------- | ------------------------ | ------------------
**your fork**   | **your fork** | push, pull_request         | N/A                      | push, pull_request
**your fork**   | **my fork**   | push                       | pull_request             | pull_request

From a security and billing perspective, this makes sense.

Let's go with the current approach: create workflow runs for both push and pull_request.

.github/workflows/simple-build.yml
```diff
@@ -34,7 +37,13 @@ jobs:
...
+      - name: checkout (push)
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: checkout (pull request)
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: build
         run: '"${CXX}" ${CXXFLAGS} -o quick-lint-js -I src -I vendor/boost src/*.cpp vendor/boost/libs/container/src/*.cpp vendor/boost/libs/container/src/alloc_lib.c'
```
strager, line 49: Nit: Let's separate the checkout from the important work with a blank line (here and in the other workflows):

```suggestion
        with:
          ref: ${{ github.event.pull_request.head.sha }}

      - name: build
        run: '"${CXX}" ${CXXFLAGS} -o quick-lint-js -I src -I vendor/boost src/*.cpp vendor/boost/libs/container/src/*.cpp vendor/boost/libs/container/src/alloc_lib.c'
```

.github/workflows/check-format.yml
```diff
@@ -15,14 +15,22 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 name: check format
-on: [push]
+on:
+  push:
```
strager, line 19: This is legal YAML? O_o Today I learned.

.github/workflows/build-and-test-web-demo.yml
```diff
@@ -25,8 +28,14 @@ jobs:
       - uses: mymindstorm/setup-emsdk@v6
         with:
           version: 2.0.4
-      - name: check out
-        uses: actions/checkout@v1
+      - name: checkout (push)
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: checkout (pull request)
```
strager, line 34: Can we give these steps the same name ("check out"), or do they need to be distinct?

strager (approved):

strager:
Beautiful! I'm gonna merge this.

strager:
Landed as commit cadbfbb1db664e1016a76e7cc9b2ffcc55243e6f.

PR #67, opened 2020-10-10
Test continue parsing after invalid let binding

erlliam:
In pull request #39 I forgot to write a test for the commit "Continue parsing after invalid let binding".

### DIFF
--- a/test/test-parse.cpp
+++ b/test/test-parse.cpp
@@ -361,6 +361,21 @@ TEST(test_parse, parse_invalid_let) {
                               error_invalid_binding_in_let_statement, where,
                               offsets_matcher(&code, 4, 12))));
   }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"let true, true, y\nlet x;");
+    parser p(&code, &v);
+    p.parse_and_visit_statement(v);
+    p.parse_and_visit_statement(v);
+    EXPECT_EQ(v.variable_declarations.size(), 2);
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(error_invalid_binding_in_let_statement,
+                                     where, offsets_matcher(&code, 4, 8)),
+                    ERROR_TYPE_FIELD(error_invalid_binding_in_let_statement,
+                                     where, offsets_matcher(&code, 10, 14))));
+  }
 }
 
 TEST(test_parse, parse_and_visit_import) {


test/test-parse.cpp
```diff
@@ -361,6 +361,32 @@ TEST(test_parse, parse_invalid_let) {
...
                               offsets_matcher(&code, 4, 12))));
   }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"let true, true, y\nlet x;");
+    parser p(&code, &v);
+    p.parse_and_visit_statement(v);
+    p.parse_and_visit_statement(v);
+    EXPECT_THAT(v.variable_declarations.size(), 2);
```
strager, line 371: Nit: Use `EXPECT_EQ` instead of `EXPECT_THAT` where possible. It compiles faster (I think).
```suggestion
    EXPECT_EQ(v.variable_declarations.size(), 2);
```

test/test-parse.cpp
```diff
@@ -361,6 +361,32 @@ TEST(test_parse, parse_invalid_let) {
...
+  {
+    spy_visitor v;
+    padded_string code(u8"let x, y, true;");
+    parser p(&code, &v);
+    p.parse_and_visit_statement(v);
+    EXPECT_THAT(v.variable_declarations.size(), 2);
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_invalid_binding_in_let_statement, where,
+                              offsets_matcher(&code, 10, 14))));
+  }
```
strager, line 389: This test looks about the same as the previous test. What's the difference?

strager (approved):
Looks good to me. Let me know when I should merge it.

test/test-parse.cpp
erlliam: This test does look blatantly useless. Not sure what the goal was there.

erlliam:
Should be good to merge.

test/test-parse.cpp
```diff
@@ -361,6 +361,21 @@ TEST(test_parse, parse_invalid_let) {
...
+    p.parse_and_visit_statement(v);
+    p.parse_and_visit_statement(v);
+    EXPECT_EQ(v.variable_declarations.size(), 2);
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_invalid_binding_in_let_statement, where,
+                              offsets_matcher(&code, 4, 8)),
+                          ERROR_TYPE_FIELD(
+                              error_invalid_binding_in_let_statement, where,
+                              offsets_matcher(&code, 10, 14))));
+  }
```
strager, line 378: Run clang-format (e.g. using the tools/format script). We're using clang-format version 9.0.1 for this project.

(I wouldn't normally mention this. Because you're contributing regularly, you should be doing things The Right Way.)

erlliam:
(that was using clang-format 7.0.1, probably not good to merge)

erlliam:
Now I am confident that it should be good to merge. I formatted using clang-format 9.0.1. Doesn't appear to have made a change, but it was done The Right Way.

PR #74, opened 2020-10-15
Issue #50 Parse -->

chapmanvoris:
Added support for --> comments.


strager:
Discussed in person: This change breaks valid code, such as `while (x --> 0) { }`. We should only treat `-->` as starting a comment if it's at the beginning of a line.

PR #75, opened 2020-10-15
Add version information

### LINKED ISSUE #71: Add --version option
Add a `--version` option to the quick-lint-js command-line utility which prints quick-lint-js' version number and exits without linting.

Detect `--version` in options.cpp, add a variable to the `option` struct in options.cpp, test the option parsing in test-option.cpp, and print the version message in main.cpp.

See also the `--help` option introduced in PR #20.

spicyyboi:
This adds two flags, --v and --version, along with implementations of two tests in test-options.cpp and therefore closes #71.

As I could not find any version info, I set the version string to "1.0.0", I don't know if that's okay.
I'd be happy to hear your feedback on this!

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -37,6 +37,12 @@
 #include <string>
 #include <variant>
 
+// Version information
+#define QUICK_LINT_JS_VERSION_MAJOR 1
+#define QUICK_LINT_JS_VERSION_MINOR 0
+#define QUICK_LINT_JS_VERSION_PATCHLEVEL 0
+#define QUICK_LINT_JS_VERSION_STRING "1.0.0"
+
 namespace quick_lint_js {
 namespace {
 class any_error_reporter {
@@ -95,6 +101,7 @@ void process_file(padded_string_view input, error_reporter *,
                   bool print_parser_visits);
 
 void print_help_message();
+void print_version_information();
 }
 }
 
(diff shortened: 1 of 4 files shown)


spicyyboi:
As some checks weren't happy about me leaving unused version information macros floating around in main.cpp, I removed them.

src/main.cpp
```diff
@@ -37,6 +37,9 @@
 #include <string>
 #include <variant>
 
+// Version information
```
strager, line 40: Nit: This comment adds no value IMO. Delete it.

src/main.cpp
```diff
@@ -37,6 +37,9 @@
...
+#define QUICK_LINT_JS_VERSION_STRING "1.0.0"
```
strager, line 41: Minor: Let's use `0.1.0` for now.

src/main.cpp
```diff
@@ -304,9 +312,14 @@ void print_help_message() {
   print_option("--output-format=[FORMAT]",
                "Format to print feedback where FORMAT is one of:");
   print_option("", "gnu-like (default if omitted), vim-qflist-json");
+  print_option("--v, --version", "Print version information");
```
strager, line 315: Shouldn't it be `-v`, not `--v`, for the short form? (See `-h` and `--help`.)

Actually, it looks like `-h` is broken. O_o Maybe we should make `-v` and `-h` work in a separate commit.

strager (approved):
Looks great! Let me know when I should merge it.

spicyyboi:
That's good to hear! I've incorporated the first two of your suggestions and I think the pull request can be merged now.

strager:
Landed as commit 2264fe0231ab12ab8a318245c32f9da1d766f152.

Issue #78, opened 2020-10-17
Support -h and -v options


strager:
quick-lint-js' command-line options parser supports `--help`, but not `-h`. Teach it to parse `-h` as if it was `--help`. Same with `--version` and `-v`.

spicyyboi:
I am interested in working on this!

strager:
Thanks @spicyyboi

PR #79, opened 2020-10-19
Add support for -h and -v options

spicyyboi:
This addresses #78. Previously, quick-lint-js only supported options starting with "--".
By adding the method `match_flag_shorthand` to the `parser` class allows to define shorthand flags (usually beginning with '-') for common options.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -324,10 +324,10 @@ void print_help_message() {
   print_option("--output-format=[FORMAT]",
                "Format to print feedback where FORMAT is one of:");
   print_option("", "gnu-like (default if omitted), vim-qflist-json");
-  print_option("--v, --version", "Print version information");
+  print_option("-v, --v, --version", "Print version information");
   print_option("--vim-file-bufnr=[NUMBER]",
                "Select a vim buffer for outputting feedback");
-  print_option("--h, --help", "Print help message");
+  print_option("-h, --h, --help", "Print help message");
 }
 
 void print_version_information() {
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -59,6 +59,18 @@ class arg_parser {
     return arg_value;
   }
 
+  bool match_flag_shorthand(std::string_view option_shorthand) noexcept {
+    if (!this->option_.has_value()) {
+      return false;
+    }
+    bool matches = this->option_->arg_key == option_shorthand;
+
+    if (matches) {
+      this->advance(1);
+    }
+    return matches;
+  }
+
   bool match_flag_option(std::string_view full_option_name,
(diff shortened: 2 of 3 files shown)


src/main.cpp
```diff
@@ -324,10 +324,10 @@ void print_help_message() {
   print_option("--output-format=[FORMAT]",
                "Format to print feedback where FORMAT is one of:");
   print_option("", "gnu-like (default if omitted), vim-qflist-json");
-  print_option("--v, --version", "Print version information");
+  print_option("-v, --v, --version", "Print version information");
   print_option("--vim-file-bufnr=[NUMBER]",
                "Select a vim buffer for outputting feedback");
-  print_option("--h, --help", "Print help message");
+  print_option("-h, --h, --help", "Print help message");
```
strager, line 330: Let's drop `--h` and `--v` from the help. Too noisy.

src/options.cpp
```diff
@@ -180,9 +192,11 @@ options parse_options(int argc, char** argv) {
       } else {
         next_vim_file_bufnr = bufnr;
       }
-    } else if (parser.match_flag_option("--help"sv, "--h"sv)) {
+    } else if (parser.match_flag_option("--help"sv, "--h"sv) ||
+               parser.match_flag_shorthand("-h"sv)) {
```
strager, line 196: Shorthand options should always be a single character, so should we just give a single character to the function?
```suggestion
    } else if (parser.match_flag_option("--help"sv, "--h"sv) ||
               parser.match_flag_shorthand('h')) {
```

test/test-options.cpp
```diff
@@ -231,6 +241,14 @@ TEST(test_options, invalid_option) {
     EXPECT_EQ(o.error_unrecognized_options[0], "--debug-parse-visits-xxx"sv);
     EXPECT_THAT(o.files_to_lint, IsEmpty());
   }
+
+  {
+    options o = parse_options({"-version", "foo.js"});
+    ASSERT_EQ(o.error_unrecognized_options.size(), 1);
+    EXPECT_EQ(o.error_unrecognized_options[0], "-version"sv);
+    EXPECT_THAT(o.files_to_lint, IsEmpty());
+  }
```
strager, line 250: :+1:

strager (approved):
Looks great! Let me know when I should merge.

spicyyboi:
Great! I've changed the function signature of `match_flag_shorthand` and the help message. I think this can be merged now.

strager:
Landed as commit 17f427964df17e3337e2c8d5d7e36cb78be56c2b.

strager:
@spicyyboi Can you please email me at strager.nds@gmail.com?

Issue #37, opened 2020-09-22
`x += 1` reports two errors, should report one


strager:

strager:
I thought I linked this issue with PR #81. Apparently not! My bad.

mushrom:
This seems simple enough, I could take a swing at it.

I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
@mushrom Sure, you can work on this task.

@FMMazur made some progress on this task in #81. Perhaps you can pick up where they left off.

mushrom:
Hmmmm, this is a lot more complicated than it looks, could reframe the problem a bit :eyes:

Current thought is that rather than having a new error type, you could (optionally) condense use errors in the output formatter, so if you have repeated use errors on the same line in the text formatter, you could output the use types and the number of errors, or something like that. This would be cleaner than adding a new use_and_assignment_... error type imo, and having a toggleable condensed/verbose mode would be a nice touch. Could further simplify things by merging the errors for assignment/use of an undeclared variable.

How does that sound, could go ahead with that if that seems fine.

strager:
> Current thought is that rather than having a new error type, you could (optionally) condense use errors in the output formatter, so if you have repeated use errors on the same line in the text formatter, you could output the use types and the number of errors, or something like that.

I think you're only thinking of the CLI use case. The main use case of quick-lint-js is as an editor extension. In an editor, we want to highlight issues inline, and allow the programmer to see a description for each error. Grouping the errors by line doesn't make sense in general. For example, notice the squigglies under both `y`s on the first line of this snippet:
<img width="441" alt="Screen Shot 2021-04-11 at 00 54 21" src="https://user-images.githubusercontent.com/48666/114296487-7a0ad400-9a60-11eb-98b0-b2cac6c57110.png">

Your suggestion does make sense for the CLI, but I don't see how it's related to this task.

> This would be cleaner than adding a new use_and_assignment_... error type imo

I think context-specific error messages are helpful for beginners. It also makes the linter feel more human and less robotic. I think a new error type makes sense.

Prior art:
* E001 & E002 & E003 & E004
  * E002 assignment to const global variable
  * E003 assignment to const variable
  * E004 assignment to const variable before its declaration
  * E001 variable assigned before its declaration
* E007 & E008
  * E007 classes cannot be named 'let'
  * E008 let statement cannot declare variables named 'let'
[...]

PR #81, opened 2020-10-21
Report one #37

FMMazur:
Added more brief return when a variable its used and assigned.

test with:
```js
x += 1
x += y
x.p += y

x += y + z;
```

returns:
```sh
../test.js:1:1: error: use and assignment of undeclared variable: x
../test.js:2:6: error: use of undeclared variable: y
../test.js:2:1: error: use and assignment of undeclared variable: x
../test.js:3:1: error: use of undeclared variable: x
../test.js:3:8: error: use of undeclared variable: y
../test.js:5:6: error: use of undeclared variable: y
../test.js:5:10: error: use of undeclared variable: z
../test.js:5:1: error: use and assignment of undeclared variable: x
```

PR #83, opened 2020-10-26
Report one error for 'x += 1' case

dannycho7:
Solves #37

### DIFF
--- a/test/test-lint.cpp
+++ b/test/test-lint.cpp
@@ -766,12 +766,11 @@ TEST(
 
   EXPECT_THAT(v.errors, IsEmpty());
 }
-
 TEST(test_lint,
      name_of_named_function_expression_is_not_usable_outside_function) {
   const char8 declaration[] = u8"f";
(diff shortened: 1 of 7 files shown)


test/test-lint.cpp
```diff
@@ -1247,9 +1261,9 @@ TEST(test_lint, use_variable_declared_in_grandparent_function) {
 }
 
 TEST(test_lint, use_for_loop_let_variable_before_or_after_loop) {
+  const char8 use_after[] = u8"element";
```
dannycho7, line 1264: I made these ordering changes because the address of these arrays is decreasing from top to bottom.

strager:
@dannycho7, someone else is already working on this issue (#81). Their PR takes a different approach than your PR.

I'll likely end up landing #81 because it appeared first. Do you want me to review your PR anyway?

cc @FMMazur

dannycho7:
@strager That's unfortunate. #81 wasn't linked to the issue by the time I sent this out :\

Issue #56, opened 2020-10-08
Improve error recovery for unterminated string literals


strager:
Inspired by issue #53.

In the following code, the programmer probably intends to escape the newline with a backslash or use a template literal:

```javascript
let x = "hello
world";
```

Currently, quick-lint-js will interpret this code as having the following tokens:

1. `kw_let`
2. `identifier` (`x`)
3. `equal`
4. `string` (`"hello`)
5. `identifier` (`world`)
6. `string` (`";`)

Ideally, instead, it'd interpret this code as having the following tokens:

1. `kw_let`
2. `identifier` (`x`)
3. `equal`
4. `string` (`"helloworld"`)
5. `semicolon`

We can't just munch until the next `"`, 'cus that would be poor error recovery for code like this:

```javascript
console.log("starting se
startServer(port);
```

Idea: if we see an end-of-line when parsing a string literal, check the following line. If the following line has exactly one matching `"` or `'`, treat that matching quote as the end of the current string. Otherwise, treat the end-of-line as the end of the current string.

The code for string literal parsing is in lex.cpp, and the existing tests are in test-lex.cpp.

A-thanasios:
Looking at this. Since the lexer only parses `\r` and `\n` as line terminators, the `\r\n` terminator is making the fix a bit uglier. One work around to this would be to parse all consecutive `\n` characters after the first `\r` or `\n` in a string literal. Is that fine?

strager:
> Since the lexer only parses `\r` and `\n` as line terminators, the `\r\n` terminator is making the fix a bit uglier.

I don't know what would be uglier about handling CRLF (`\r\n`). Can you show some code?

Code exists in locator::cache_offsets_of_lines (src/location.cpp) to parse CRLF. I'm a bit surprised we don't already have anything in lexer to deal with CRLF.

> One work around to this would be to parse all consecutive \n characters after the first \r or \n in a string literal. Is that fine?

Wouldn't that make the following code lex differently depending on whether CRLF (`\r\n`) or LF (`\n`) is used for terminating lines?

```javascript
let x = "hello


world";
```

A-thanasios:
> I don't know what would be uglier about handling CRLF (`\r\n`). Can you show some code?
> 
> Code exists in locator::cache_offsets_of_lines (src/location.cpp) to parse CRLF. I'm a bit surprised we don't already have anything in lexer to deal with CRLF. I guess we haven't needed it yet.
> 

[...]

PR #85, opened 2020-10-28
Improve error recovery for unterminated string literals.

A-thanasios:
#56

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -451,27 +451,28 @@ retry:
 
       case '\n':
       case '\r': {
-        char8* c_up_to_first_matching_quote = nullptr;
-        char8* c_end = c;
-        if (c_end[0] == '\r' && c_end[1] == '\n') {
-          c_end += 2;
+        char8* matching_quote = nullptr;
+        char8* current_c = c;
+        if (current_c[0] == '\r' && current_c[1] == '\n') {
+          current_c += 2;
         } else {
-          c_end += 1;
+          current_c += 1;
         }
         for (;;) {
-          if (*c_end == opening_quote) {
-            if (c_up_to_first_matching_quote) {
+          if (*current_c == opening_quote) {
+            if (matching_quote) {
               break;
             }
-            c_up_to_first_matching_quote = ++c_end;
-          } else if (*c_end == '\r' || *c_end == '\n' ||
-                     (*c_end == '\0' && this->is_eof(c_end))) {
-            if (c_up_to_first_matching_quote) {
-              c = c_up_to_first_matching_quote;
+            matching_quote = current_c;
+            ++current_c;
+          } else if (*current_c == '\r' || *current_c == '\n' ||
+                     (*current_c == '\0' && this->is_eof(current_c))) {
+            if (matching_quote) {
+              c = matching_quote + 1;
             }
             break;
           } else {
-            ++c_end;
+            ++current_c;
           }
         }
         this->error_reporter_->report(error_unclosed_string_literal{
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -450,10 +450,34 @@ retry:
         }
 
       case '\n':
-      case '\r':
+      case '\r': {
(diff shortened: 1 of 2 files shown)


test/test-lex.cpp
```diff
@@ -525,6 +525,72 @@ TEST(test_lex, lex_strings) {
...
+    padded_string input(u8"'separated" + string8(line_terminator) + u8"hello'");
+    lexer l(&input, &v);
+    EXPECT_EQ(l.peek().type, token_type::string);
+    l.skip();
+    EXPECT_EQ(l.peek().type, token_type::end_of_file);
+
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_unclosed_string_literal, string_literal,
+                    offsets_matcher(&input, 0, 16 + line_terminator.size()))));
```
strager, line 539: Could we use `input.size()` instead of `16 + line_terminator.size()` here?

test/test-lex.cpp
```diff
@@ -525,6 +525,72 @@ TEST(test_lex, lex_strings) {
...
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(
+            ERROR_TYPE_FIELD(error_unclosed_string_literal, string_literal,
+                             offsets_matcher(&input, 0, 10)),
+            ERROR_TYPE_FIELD(
+                error_unclosed_string_literal, string_literal,
+                offsets_matcher(&input, 15 + 2 * line_terminator.size(),
+                                16 + 2 * line_terminator.size()))));
+  }
```
strager, line 564: :+1: I like this test.

src/lex.cpp
```diff
@@ -450,10 +450,34 @@ void lexer::parse_current_token() {
         }
 
       case '\n':
-      case '\r':
+      case '\r': {
+        char8* c_up_to_first_matching_quote = nullptr;
```
strager, line 454: Nit: I'd name this variable `matching_quote`, and decrement it. (See my comment on lines 466-470.)

src/lex.cpp
```diff
@@ -450,10 +450,34 @@ void lexer::parse_current_token() {
...
+        for (;;) {
+          if (*c_end == opening_quote) {
+            if (c_up_to_first_matching_quote) {
+              break;
+            }
+            c_up_to_first_matching_quote = ++c_end;
+          } else if (*c_end == '\r' || *c_end == '\n' ||
+                     (*c_end == '\0' && this->is_eof(c_end))) {
+            if (c_up_to_first_matching_quote) {
+              c = c_up_to_first_matching_quote;
```
strager, line 470: (From comment on line 454)

This is what I mean by 'decrement':
```suggestion
            c_up_to_first_matching_quote = c_end;
            ++c_end;
          } else if (*c_end == '\r' || *c_end == '\n' ||
                     (*c_end == '\0' && this->is_eof(c_end))) {
            if (c_up_to_first_matching_quote) {
              c = c_up_to_first_matching_quote + 1;
```

src/lex.cpp
```diff
@@ -450,10 +450,34 @@ void lexer::parse_current_token() {
         }
 
       case '\n':
-      case '\r':
+      case '\r': {
+        char8* c_up_to_first_matching_quote = nullptr;
+        char8* c_end = c;
```
strager, line 455: Hmm, do you think this code would be better in its own function? In a separate function, better variable names might come more naturally.

strager (approved):

test/test-lex.cpp
A-thanasios: Thanks :hear_no_evil:

A-thanasios:
Ready to merge

strager:
It looks like some builds are failing on CI: https://github.com/quick-lint/quick-lint-js/runs/1535159043

Please fix the builds. (I think you want a `narrow_cast<int>(...)`.)

strager:
Landed as commits b879f5138, 93e63d2c5, and 65e936aaa. Thanks!

PR #86, opened 2020-10-28
Update mymindstorm/setup-emsdk version

jgil-r:
Updated mymindstorm/setup-emsdk from v6 to v7 to fix @actions/core environment file syntax change warnings


strager:
Beautiful! Thanks.

PR #87, opened 2020-11-18
Support unicode command-line arguments on Windows

### LINKED ISSUE #59: Run CI on pull requests

### LINKED ISSUE #65: Support wchar_t in command-line parsing
The command-line parser uses `char`. This is [likely] problematic on Windows if command-line arguments contain non-ASCII characters. Make the command-line parser support `wchar_t` and `WinMain`.

This change might have ripple effects, such as the `CreateFileA` call in `read_file`. We should probably do some UTF-16-to-UTF-8 conversions to ease the transition.

erlliam:
Resolves #65

Ignore the .gitignore and maybe src/CMakeLists.txt changes. Sorry for the commit log being a joke, will rebase when ready.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -28,6 +28,7 @@
 #include <quick-lint-js/have.h>
 #include <quick-lint-js/math-overflow.h>
 #include <quick-lint-js/narrow-cast.h>
+#include <quick-lint-js/utf-16.h>
 #include <string>
 
 #if QLJS_HAVE_FCNTL_H
@@ -199,9 +200,15 @@ read_file_result read_file(const char *path, posix_fd_file_ref file) {
 
 #if defined(QLJS_FILE_WINDOWS)
 read_file_result read_file(const char *path) {
-  // TODO(strager): Use CreateFileW.
-  HANDLE handle = ::CreateFileA(
-      path, /*dwDesiredAccess=*/GENERIC_READ,
+  std::optional<std::wstring> wpath = quick_lint_js::mbstring_to_wstring(path);
+  if (!wpath) {
+    DWORD error = ::GetLastError();
+    return read_file_result::failure(std::string("failed to convert ") + path +
+                                     " to wstring\n" +
+                                     windows_error_message(error));
+  }
+  HANDLE handle = ::CreateFileW(
+      wpath->c_str(), /*dwDesiredAccess=*/GENERIC_READ,
       /*dwShareMode=*/FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE,
       /*lpSecurityAttributes=*/nullptr,
       /*dwCreationDisposition=*/OPEN_EXISTING,
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -37,6 +37,7 @@
 #include <quick-lint-js/text-error-reporter.h>
 #include <quick-lint-js/translation.h>
 #include <quick-lint-js/unreachable.h>
+#include <quick-lint-js/utf-16.h>
 #include <quick-lint-js/vector.h>
 #include <quick-lint-js/version.h>
 #include <quick-lint-js/vim-qflist-json-error-reporter.h>
@@ -105,6 +106,8 @@ class any_error_reporter {
   tape_variant tape_;
 };
 
+[[noreturn]] void handle_options(quick_lint_js::options o);
+
 void process_file(padded_string_view input, error_reporter *,
                   bool print_parser_visits);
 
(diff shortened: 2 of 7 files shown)


strager (changes requested):
As discussed in person:

* Looks good in general
* Memory leaks need fixin'
* Some changes should be put into a separate PR

src/main.cpp
```diff
@@ -382,5 +407,51 @@ void print_help_message() {
...
+      /*cchWideChar=*/-1,
+      /*lpMultiByteStr=*/nullptr,
+      /*cbMultiByte=*/0,
+      /*lpDefaultChar=*/NULL,
+      /*lpUsedDefaultChar=*/NULL);
+  if (size_required == 0) {
+    goto fail;
+  }
+  char *mbstring;
+  mbstring = new char[size_required];
```
erlliam, line 434: I'm pretty sure I have to allocate the string on the heap for mbargv. I don't like how I do a loop to de-allocate all the strings returned from this function. I think I read about smart pointers and how when a vector destructs it will call the destructor for all the things it is storing. Which means that the if I use smart pointers, I won't have to loop to de-allocate the string. Sounds like it's what I ought to use.

strager: > I'm pretty sure I have to allocate the string on the heap for mbargv.

Correct. `std::string` is probably a good option for that. A smart pointer like `std::unique_ptr<char>` would also work.

src/main.cpp
```diff
@@ -333,6 +363,53 @@ class multi_visitor {
 QLJS_STATIC_ASSERT_IS_PARSE_VISITOR(
     multi_visitor<debug_visitor, debug_visitor>);
 
+void handle_options(quick_lint_js::options o) {
```
strager, line 366: Nit: To keep the diff readable, move this code back to where it was originally. (Introduce an unnamed namespace block as needed.)

src/file.cpp
```diff
@@ -198,10 +198,42 @@ read_file_result read_file(const char *path, posix_fd_file_ref file) {
 #endif
 
 #if defined(QLJS_FILE_WINDOWS)
+std::optional<std::wstring> mbstring_to_wstring(const char *mbstring) {
```
strager, line 201: Minor: I'd move this function (and also move the mbargv class) into their own module. This will let you (or anyone else) write tests for the code.

strager:
Can you document what testing your performed (either in the commit message or in the pull request's description)?

strager (approved):
Overall, this looks great!

erlliam:
I still have to automate/write tests. I am planning on creating a "fake argc, argv", convert it using my class and then passing the result to `quick_lint_js::parse_options` and checking which "flags" are set in the options class.

strager:
> I still have to automate/write tests.

Automated tests are not preventing this pull request from merging, IMO.

> I am planning on creating a "fake argc, argv", convert it using my class and then passing the result to quick_lint_js::parse_options and checking which "flags" are set in the options class.

A test or two sounds good. The test would be Windows-only, because mbargv is Windows-only.

Is it really necessary to test mbargv with parse_options, though? Couldn't we test just mbargv, without involving parse_options?

src/quick-lint-js/utf-16.h
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef QUICK_LINT_JS_UTF_16_H
+#define QUICK_LINT_JS_UTF_16_H
+
+#include <optional>
+#include <string>
```
strager, line 21: Also include `<vector>`.

src/quick-lint-js/utf-16.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace quick_lint_js {
+#if defined(_WIN32)
+class mbargv {
+ public:
+  explicit mbargv(int argc, wchar_t **wargv);
+  ~mbargv();
```
strager, line 28: Follow the Rule of Five (or the Rule of Three): if you declare a destructor, you should also declare a copy constructor, a move constructor, a copy assignment operator, and a move assignment operator.

To simplify things, you could delete the copy functions to satisfy the rule:
```cpp
mbargv(const mbargv &) = delete;
mbargv &operator=(const mbargv &) = delete;
```

strager (approved):

strager:
Landed as commit 9558fdd71718f4ef65434c971155a74bcbfa5862.

PR #90, opened 2020-11-26
Report parse error when there's no digits exist on number literals.

### LINKED ISSUE #43: Report error when binary/octal/hex number has no digits
In debug builds, quick-lint-js crashes on the following JavaScript snippets:

```javascript
let x = 0x;
```
```javascript
let y = 0b;
```

This is because quick-lint-js assumes that a digit follows `0x` or `0b`, which is not necessarily true.

Teach quick-lint-js to report a user-friendly error on incomplete number literals. For example:
```
hello.js:1:9: binary number literal has no digits
```

Note: support for `0o`-prefixed octal number literals will be added in #7.

catac1:
This pull request is for trying to resolve #43.  

- [x] binary
- [x] octal
- [x] hex

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -782,7 +782,7 @@ void lexer::parse_binary_number() {
 
   if (input == this->input_) {
     this->error_reporter_->report(
-        error_no_digits_in_binary_number{source_code_span(input, input)});
+        error_no_digits_in_binary_number{source_code_span(this->last_token_.begin, input)});
   } else {
     this->input_ = check_garbage_in_number_literal(input);
   }
@@ -795,7 +795,7 @@ void lexer::parse_octal_number(octal_kind kind) {
 
   if (input == this->input_) {
     this->error_reporter_->report(
-        error_no_digits_in_octal_number{source_code_span(input, input)});
+        error_no_digits_in_octal_number{source_code_span(this->last_token_.begin, input)});
     return;
   }
 
@@ -922,7 +922,7 @@ void lexer::parse_hexadecimal_number() {
 
   if (input == this->input_) {
     this->error_reporter_->report(
-        error_no_digits_in_hex_number{source_code_span(input, input)});
+        error_no_digits_in_hex_number{source_code_span(this->last_token_.begin, input)});
   } else {
     this->input_ = check_garbage_in_number_literal(input);
   }
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -108,8 +108,6 @@
   case '8':                     \
   case '9'
 
-// TODO: add more space characters.
-
 namespace quick_lint_js {
 string8_view identifier::normalized_name() const noexcept {
   const char8* begin = this->span_.begin();
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -759,6 +759,7 @@ char8* lexer::check_garbage_in_number_literal(char8* input) {
     switch (*input) {
     QLJS_CASE_DECIMAL_DIGIT:
     QLJS_CASE_IDENTIFIER_START:
+    case u8'.':
       input += 1;
       break;
     default:
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -798,6 +798,13 @@ void lexer::parse_octal_number(octal_kind kind) {
 
   input = this->parse_octal_digits(input);
 
+  char8 c = *(input + 1);
+  if (input == this->input_ && (this->is_space(c) || c == u8'\0')) {
+    this->error_reporter_->report(
+        error_no_digits_in_octal_number{source_code_span(input, input)});
+    return;
+  }
+
   if (kind == octal_kind::sloppy && is_digit(*input)) {
     this->input_ = input;
     this->parse_number();
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -749,10 +749,14 @@ const char8* lexer::end_of_previous_token() const noexcept {
   return this->last_last_token_end_;
 }
 
+template <class Error>
 char8* lexer::check_garbage_in_number_literal(char8* input) {
   char8* garbage_begin = input;
   for (;;) {
     switch (*input) {
+    // if we see a DECIMAL_DIGIT at this point it either means we're in
+    // strict mode or following another number character (such as a `.`),
+    // both of which are invalid
     QLJS_CASE_DECIMAL_DIGIT:
     QLJS_CASE_IDENTIFIER_START:
     case u8'.':
@@ -765,8 +769,8 @@ char8* lexer::check_garbage_in_number_literal(char8* input) {
 done_parsing_garbage:
   char8* garbage_end = input;
   if (garbage_end != garbage_begin) {
-    this->error_reporter_->report(error_unexpected_characters_in_number{
-        source_code_span(garbage_begin, garbage_end)});
+    this->error_reporter_->report(
+        Error{source_code_span(garbage_begin, garbage_end)});
     input = garbage_end;
   }
 
@@ -781,10 +785,11 @@ void lexer::parse_binary_number() {
   }
 
   if (input == this->input_) {
-    this->error_reporter_->report(
-        error_no_digits_in_binary_number{source_code_span(this->last_token_.begin, input)});
+    this->error_reporter_->report(error_no_digits_in_binary_number{
+        source_code_span(this->last_token_.begin, input)});
   } else {
-    this->input_ = check_garbage_in_number_literal(input);
+    this->input_ = check_garbage_in_number_literal<
+        error_unexpected_characters_in_binary_number>(input);
   }
 }
 
@@ -794,8 +799,8 @@ void lexer::parse_octal_number(octal_kind kind) {
   input = this->parse_octal_digits(input);
 
   if (input == this->input_) {
-    this->error_reporter_->report(
-        error_no_digits_in_octal_number{source_code_span(this->last_token_.begin, input)});
+    this->error_reporter_->report(error_no_digits_in_octal_number{
+        source_code_span(this->last_token_.begin, input)});
     return;
   }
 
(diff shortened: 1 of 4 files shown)


test/test-lex.cpp
```diff
@@ -161,22 +161,21 @@ TEST(test_lex, lex_binary_numbers) {
...
-                              error_unexpected_characters_in_number, characters,
-                              offsets_matcher(&input, 3, 4))));
-  }
+TEST(test_lex, fail_lex_binary_number_no_digits) {
+  check_tokens_with_errors(
+      u8"0b", {token_type::number},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors, ElementsAre(ERROR_TYPE_FIELD(
+                                error_no_digits_in_binary_number,
+                                characters, offsets_matcher(input, 2, 2))));
```
strager: Blocking: Shouldn't we highlight the `0b`? As written, the end of the file is highlighted.
```suggestion
                                characters, offsets_matcher(input, 0, 2))));
```

src/lex.cpp
```diff
@@ -771,14 +777,19 @@ char8* lexer::check_garbage_in_number_literal(char8* input) {
...
   }
 
-  this->input_ = check_garbage_in_number_literal(input);
+  char8 c = *(input + 1);
+  if (input == this->input_ && (this->is_space(c) || c == u8'\0')) {
+    this->error_reporter_->report(error_no_digits_in_binary_number{
+        source_code_span(input, input)});
+  } else {
+    this->input_ = check_garbage_in_number_literal(input);
+  }
```
strager: What about code like this:
```javascript
let x = 0b;
let y = myArray[0b];
```
I think your implementation would fail here. You're only checking for `\0`/EOF and whitespace, not for characters like `;` and `]`. Maybe I'm wrong, though; I didn't analyze this super carefully.

What about an implementation like this:
```suggestion
  this->input_ = check_garbage_in_number_literal(input);
  if (input == this->input_) {
    this->error_reporter_->report(error_no_digits_in_binary_number{
        source_code_span(input, input)});
  }
```

test/test-lex.cpp
```diff
@@ -178,6 +178,16 @@ TEST(test_lex, fail_lex_binary_number_no_digits) {
...
 
+TEST(test_lex, fail_lex_binary_number) {
+  check_tokens_with_errors(
+      u8"0b1.1", {token_type::number},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors, ElementsAre(ERROR_TYPE_FIELD(
+                                error_unexpected_characters_in_number,
+                                characters, offsets_matcher(input, 3, 5))));
+      });
+}
```
strager: Looks good to me!

src/lex.cpp
```diff
@@ -759,6 +759,7 @@ char8* lexer::check_garbage_in_number_literal(char8* input) {
     switch (*input) {
     QLJS_CASE_DECIMAL_DIGIT:
     QLJS_CASE_IDENTIFIER_START:
+    case u8'.':
```
strager, line 762: Would this change also affect hex, octal, and decimal literals? Should we test those as well?

strager (changes requested):
See comments marked 'Blocking'.

I also think we're only handling a few cases, but we should generalize to support many more cases.

test/test-lex.cpp
catac1: this is resolved in commit 22c093d81ac5b544d3b4f92e07ecf2fe1423b3dd

src/lex.cpp
catac1: This function was affecting both binary and hexadecimal cases not the octal case. 
There's exact code path in the octal case so I made it polymorphic over `QLJS_ERROR_TYPE` 
in commit  83f70d13eaa579a6a13cbe24478d24cf96b4d64e using template parameter.

src/lex.cpp
catac1: Indeed, my code was failing on the cases  you mentioned. so I changed the code as you suggested above and
It's now more generalized than before. But I don't know how to write the test case for multiple multiple expression.
I only checked a few cases by manually running the executable.

src/lex.cpp
catac1: And We have simple tests for all the hex, octal and decimal literals. octal test case has been there already.

# new test cases links
- `"0b1.1"` [binary test case](https://github.com/catac1/quick-lint-js/blob/1538d2a80a24032d71d70e17a35e9d2852acbf25/test/test-lex.cpp#L174)
- `"0xf.f"` [hex test case](https://github.com/catac1/quick-lint-js/blob/1538d2a80a24032d71d70e17a35e9d2852acbf25/test/test-lex.cpp#L265)

this is very simplistic test. so we have to add more.

test/test-lex.cpp
catac1: I don't have to pass the number literal prefix pointer to parse functions because there's [`lex::last_token_.begin`](https://github.com/catac1/quick-lint-js/blob/1538d2a80a24032d71d70e17a35e9d2852acbf25/src/quick-lint-js/lex.h#L365) 
member variable and it is set to prefix location before all litenal number parse function calls at here [lex.cpp#L150](https://github.com/catac1/quick-lint-js/blob/1538d2a80a24032d71d70e17a35e9d2852acbf25/src/lex.cpp#L150)

src/lex.cpp
catac1: ok, `check_tokens_with_errors` in `test-lex.cpp` expects to testing ends with [`token_type::end_of_file`](https://github.com/catac1/quick-lint-js/blob/1538d2a80a24032d71d70e17a35e9d2852acbf25/test/test-lex.cpp#L1426) 
so we can't test like `"0b;"` to test `parse_binary_number` because it will ends with `token_type::semicolon`

catac1: I think there's three options here.

1. Just check errors manually without using `check_tokens_with_errors`.
   Easy but terrible code duplication expected.

2. Write another function which expecting `token_type::semicolon` after the parsing.
   Also easy with less code duplication than  option 1 but still not flexible enough because there may exist 
   different expected ending `token_type` in the future.

3. Add additional parameter `token_type` to the `check_tokens_with_errors` with default parameter `token_type::end_of_file`.  Most flexible one in these options. 

I'm not sure about option 2 and 3 is in the scope of this pull request.

strager: check_tokens_with_errors takes a list of tokens, not just a single token. So you can write something like this:

```cpp
check_tokens_with_errors(
    u8"0b;", {token_type::number, token_type::semicolon},
    ...);
```

catac1: All right. I pushed some more test cases. but I found another failing case  even with this change.
Currently, the `lex` produces `{token_type::number, token_type::identifier}`  for  `"0b_"`  because `lex::parse_binary_number` doesn't consume underscore characters when there's no digit exists in number literals.  

Below cases are unclear to me what is the correct range of the `error_no_digits_in_binary_number` error.

1. ```
   A        B
   0b_11    0b_11
   ^^^      ^^^^^
   ```               

 
2. ```
   A        B        C
   0b_13    0b_13    0b_13
   ^^^      ^^^^     ^^^^^
   ```


3. ```
   A        B         C
   0b__11   0b__11    0b__11          
   ^^^      ^^^^      ^^^^^^
   ```

strager: Hmm, that's a good question. For these cases, I think we should create a new error type, `error_number_literal_contains_leading_underscores`. That's out of scope for this PR, I think. Let's file an issue, and accept your current behavior for now.

src/lex.cpp
```diff
@@ -108,6 +108,8 @@
   case '8':                     \
   case '9'
 
+// TODO: add more space characters.
+
```
strager, line 112: This comment looks vestigial. Delete?

src/quick-lint-js/error.h
```diff
@@ -260,12 +260,38 @@
...
+             characters))                                                      \
+                                                                               \
   QLJS_ERROR_TYPE(                                                             \
       error_unexpected_characters_in_octal_number,                             \
       { source_code_span characters; },                                        \
       .error(QLJS_TRANSLATE("unexpected characters in octal literal"),         \
              characters))                                                      \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_unexpected_characters_in_hex_number,                               \
```
strager, line 276: Nit: Alphabetically sort the errors by the error name. `error_unexpected_characters_in_hex_number` should come before `error_unexpected_characters_in_octal_number`, for example.

src/quick-lint-js/error.h
```diff
@@ -260,12 +260,38 @@
       error_unexpected_control_character, { source_code_span character; },     \
       .error(QLJS_TRANSLATE("unexpected control character"), character))       \
                                                                                \
```
strager, line 262: Line 254-257: Could we delete `error_unexpected_characters_in_number`? Is it used anywhere still?

strager (approved):
Looks great! Let me know when I should land this.

src/quick-lint-js/error.h
catac1: it's still used at [lex::parse_numbrer](https://github.com/catac1/quick-lint-js/blob/5c835c280253897dd4095007ba19f639e1cecac1/src/lex.cpp#L846) the test case for it are [test-lex.cpp#L320](https://github.com/catac1/quick-lint-js/blob/5c835c280253897dd4095007ba19f639e1cecac1/test/test-lex.cpp#L320).

src/lex.cpp
catac1: removed it at f1854c0b9168e7f16d85fbf96a8603cb10fff831

catac1:
I think it's ready to merge.

catac1:
I really don't mind squashing the commits.

strager:
Landed as commit d2d645a8d67055db88e5a6a027c3b0e4af76f30e. Thanks!

PR #91, opened 2020-12-02
Add header files to CMakeLists.txt files

erlliam:


strager:
An explanation would be nice to include in the commit message.

Should we also list header files for quick-lint-js-test and for quick-lint-js-benchmark-support?

src/CMakeLists.txt
```diff
@@ -69,6 +69,58 @@ quick_lint_js_add_library(
   vector.cpp
   vim-qflist-json-error-reporter.cpp
   wasm-demo-error-reporter.cpp
+  quick-lint-js/assert.h
```
strager, line 72: Nit: I'd like this list to be sorted lexicographically, such that `quick-lint-js/assert.h` appears between `parse.cpp` and `text-error-reporter.cpp`.

strager (approved):

erlliam:
It should be good to go.

strager:
Landed as commit 0f6cfbdcd. Thanks!

PR #129, opened 2021-01-12
📝 Merge README.md and plugin/vscode/README.md

vegerot:
README.md in the root describes QLJS's values, and the README.md in
plugin/vscode/ gives some examples as well as an animation.  This commit
combines the best from each

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,20 @@
 
 quick-lint-js finds bugs in JavaScript programs.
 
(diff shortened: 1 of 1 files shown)


strager:
I have other plans for the **values** section. I think it belongs not in the VS Code readme. Please undo the changes to /plugin/vscode/README.md.

The changes to /README.md look good, though.

Issue #121, opened 2020-12-27
CI: track binary size over time


strager:

strager:
In #61 I documented an idea for storing data and making charts. That idea could apply to tracking binary sizes too.

strager:
Scripts implemented in commit 32ecbe8f67a716d8ef5b6036d8085ea737b27216.

The scripts are not automated by CI, so the system isn't ideal. Probably premature to invest too much time into automation. (The scripts themselves will likely provide little value anyway...)

Issue #61, opened 2020-10-09
Run precise benchmarks on CI


strager:
We could run some precise benchmarks on CI. Examples:

* Maximum memory usage (ASLR should be disabled)
* Instruction count (ASLR should be disabled?)
* Cycle count (ASLR should be disabled) (needs a CPU emulator)

Valgrind might help a lot.

Also, we need to store the collected data somewhere, and have some fancy dashboards to present the data.

strager:
Memory usage:

Valgrind's memcheck tool provides some details, but not a high-water-mark:

```
==14522== HEAP SUMMARY:
==14522==     in use at exit: 130,789 bytes in 182 blocks
==14522==   total heap usage: 791 allocs, 609 frees, 211,956 bytes allocated
```

Instruction count:

Valgrind can provide this information. I think we can run the Cachegrind tool and track its [global statistics](https://www.valgrind.org/docs/manual/cg-manual.html#cg-manual.the-global). We need to specify cache sizes and such; I think the default is inherited from the host, making the results non-portable. (But libc versions also play a factor in portability...)

Database:

We could use a Git repository as a database. GitHub Actions should be able to create a Git commit containing JSON files with the measurements. We could make a program which extracts data from this Git repository and generates charts. Hacky, but easy to implement and debug.

strager:
Database:

I set up an SFTP/HTTPS server for builds: https://c.quick-lint-js.com/builds/ We could use this server to store benchmark results too. Probably easier to deal with than Git.

strager:
Deno has some precise benchmarks which could inspire us: https://deno.land/benchmarks

PR #137, opened 2021-01-16
precise benchmarks on CI and Metric dumps

cxxboy:
#61 & #121

### DIFF
--- a/ci/precise_benchmarks.py
+++ b/ci/precise_benchmarks.py
@@ -4,9 +4,11 @@ import sys
 import json
 from time import gmtime, strftime
 import platform
+import git
+
 
 if(len(sys.argv) != 2):
-    print("Usage: ./precise_benchmarks.py <PROCESS NAME>")
+    print("Usage: ./precise_benchmarks.py <ABS PATH TO PROCESS>")
     exit(1);
 
 def run(cmd):
@@ -57,11 +59,37 @@ def cachegrind_processor(code, out):
 
 def create_json_from_dict(result_dict, file_name):
     with open(file_name, "w") as json_dump_file:
-        json.dump(result_dict, json_dump_file, sort_keys=True, indent=4)
+        json.dump(result_dict, json_dump_file, sort_keys=False, indent=4)
+
+def memcheck_processor(code, out):
+    memcheck_linesplit = []
+    for line in out.splitlines():
+        memcheck_linesplit.append(line)
+    memcheck_split = memcheck_linesplit[-15:] 
+    memcheck_results = []
+    for index, values in enumerate(memcheck_split):
+        token_index = get_colon_index(memcheck_split[index])
+        if token_index > 0:
+            token_value = memcheck_split[index][token_index:].strip()
+            preprocessed_token_value = preprocess_brackets(token_value)
+            memcheck_results.append(preprocessed_token_value)
+    return memcheck_results
+
 
 def main():
-    code, out, err = run(["valgrind", "--tool=cachegrind", sys.argv[1]])
-    cachegrind_results = cachegrind_processor(code, out.decode('utf-8'))
+    code_cache, out_cache, err = run(["valgrind", "--tool=cachegrind", sys.argv[1]])
+    cachegrind_results = cachegrind_processor(code_cache, out_cache.decode('utf-8'))
+    code_mem, out_mem, err = run(["valgrind", "--tool=memcheck", sys.argv[1]])
+    memcheck_results = memcheck_processor(code_mem, out_mem.decode('utf-8'))
+    repo = git.Repo(search_parent_directories=True)
+    useful_bench_info = {
+            "Operating System:": platform.platform(),
+            "Platform Uname:": platform.uname(),
+            "Platform Processor:": platform.processor(),
+            "System Type:": platform.system(),
+            "Binary Size:": f"{os.path.getsize(sys.argv[1])} bytes",
+            "Git Commit:": repo.head.object.hexsha,
+    }
     cachegrind_result_dict = {
             "I_ref": cachegrind_results[0],
             "I1_misses": cachegrind_results[1],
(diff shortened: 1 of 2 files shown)


ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+        stderr = subprocess.STDOUT,
+    )
+    stdout, stderr = proc.communicate()
+    return proc.returncode, stdout, stderr
+def get_colon_index(string):
+    index = 0
+    found = False
+    for char in string:
+     if char == ':':
+         Found = True
```
strager, line 25: This isn't changing the same variable referenced on lines 22 and 28. This looks like a bug.

However, the bug doesn't matter, because we return on the next line anyway. So what's the point of this assignment? Delete it.

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+     token_index = get_colon_index(generic_split[index])
+     if token_index > 0:
+         token_value = generic_split[index][token_index:].strip()
+         preprocessed_token_value = preprocess_brackets(token_value)
+         generic_results.append(preprocessed_token_value)
+    return generic_results
+
+def create_json_from_dict(result_dict, file_name):
+    with open(file_name, "w") as json_dump_file:
+     json.dump(result_dict, json_dump_file, sort_keys=False, indent=4)
```
strager, line 60: Minor: Why sort_keys false? Don't we want consistent output?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+
+def remove_comma(string):
+    returner = ""
+    for char in string:
+     if char != ',':
+         returner += str(char)
+     else:
+         continue
+    return returner
```
strager, line 69: Why not `string.replace(",", "")`?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+
+def remove_percent(string):
+    returner = ""
+    for char in string:
+     if char != '%':
+         returner += str(char)
+     else:
+         continue
+    return returner
```
strager, line 78: Why not `string.replace("%", "")`?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    for char in string:
+     if char == ' ':
+         has_space = True 
+         index+1
+         break;
+     index = index + 1
+    if has_space:
+     return index
+    else:
+     return -1
```
strager, line 92: Why not `string.index(" ")`?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+
+def percent_to_float(percent):
+    return float(percent/100)
```
strager, line 95: I think `/100` gives a float already, so there's no need to cast explicitly.
```suggestion
    return percent/100
```

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
+import subprocess
+import os
+import sys
+import json
+from time import gmtime, strftime
+import platform
+import git
```
strager, line 7: Blocking: This script won't work unless the user installs extra packages. List the `git` dependency in a `requirements.txt` file.

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    cachegrind_results = generic_processor(code_cache, out_cache.decode('utf-8'))
+    code_mem, out_mem, err = run(["valgrind", "--tool=memcheck", sys.argv[1]])
+    memcheck_results = generic_processor(code_mem, out_mem.decode('utf-8'))
+    code_instr, out_instr, err = run(["valgrind", "--tool=callgrind", sys.argv[1]])
+    instr_results = generic_processor(code_instr, out_instr.decode('utf-8'))
+    repo = git.Repo(search_parent_directories=True)
+    useful_bench_info = {
+            "Operating System": platform.platform(),
+            "Platform Uname": platform.uname(),
+            "Binary Size": int(os.path.getsize(sys.argv[1])),
```
strager, line 113: Why `int()`? Does `getsize` not return a number?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+            "Git Commit": repo.head.object.hexsha,
+    }
+    cachegrind_result_dict = {
+            "I_ref": int(remove_comma(cachegrind_results[0].strip())),
```
strager, line 117: Almost-blocking: If Valgrind adds a new field, or changes the order of fields, this code will silently give us bogus data. Could we parse data in a way such that we don't need to list each field?

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+def percent_to_float(percent):
+    return float(percent/100)
+
+def split_commas_into_token(string):
+    return [str_.strip() for str_ in string.split(',')]
+def split_space_into_token(string):
+    return [str_.strip() for str_ in string.split(' ')]
+ 
+def main():
+    code_cache, out_cache, err = run(["valgrind", "--tool=cachegrind", sys.argv[1]])
```
strager, line 103: When using cachegrind, we should specify cache parameters such as cache sizes explicitly. Otherwise, the results here will vary depending on the CPU of the host.

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+    final_dict = {
+            "System Information": useful_bench_info,
+            "Metrics": metrics_dict
+    }
+    output_file_name = strftime("%Y-%m-%d %H:%M:%S", gmtime())
+    create_json_from_dict(final_dict, output_file_name)
+    
+if __name__ == "__main__":
+    main()
+    os.system("rm -f cachegrind.out* callgrind.out*")
```
strager, line 150: Blocking: Don't delete files you didn't create! If the user happened to have some cachegrind or callgrind files, this line would delete them. Naughty!

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
...
+import os
+import sys
+import json
+from time import gmtime, strftime
+import platform
+import git
+
+if(len(sys.argv) != 2):
+    print("Usage: ./precise_benchmarks.py <ABS PATH TO PROCESS>")
+    exit(1);
```
strager, line 11: Minor: Let's use `argparse` instead.

ci/precise_benchmarks.py
```diff
@@ -0,0 +1,150 @@
+import subprocess
```
strager, line 1: Nit: Format the code using `black`.

strager (changes requested):
There are a few major problems with this script. See comments marked 'Blocking'.

It looks like you're new to Python. I suggested a few places where you can use standard Python functions.

strager:
@cxxboy Do you plan on continuing this PR?

cxxboy:
@strager No. Sorry for ghosting. Since the issue is already fixed, I have no plan to work on it anymore.

Issue #127, opened 2021-01-11
Website: dark mode


strager:
Create a dark mode theme for the website.

LeeWannacott:
Could possibly try using: https://www.npmjs.com/package/darkmode-js

strager:
> Could possibly use this library: https://www.npmjs.com/package/darkmode-js

A configuration widget sounds like overkill for a basic webpage like quicklintjs.com.

I'd prefer a modest sprinkle of [`prefers-color-scheme`](https://developer.mozilla.org/en-us/docs/Web/CSS/@media/prefers-color-scheme).

LeeWannacott:
I didn't know that existed, I should read over the MDN Web Docs to become aware of what exists. 

I can have a go at implementing this challenging issue if you wish.

strager:
> I can have a go at implementing this challenging issue if you wish.

Sure, you may hack on this task.

strager:
Deployed dark mode at commit b9451e223fb30fb0e271a44732e39a317754ffcc (https://github.com/quick-lint/quick-lint-js/actions/runs/598272894).

PR #146, opened 2021-01-29
Website: dark mode #127

LeeWannacott:
I basically went with a solarized dark theme like in GNOME Terminal. The apple logo is coloured(colored for american's) white by using filter to invert the colour; probably not the best implementation. I might write a script that can change fill from SVG files sounds like a good little project...

Let me know if you want changes. Cheers.


strager:
Style opinions about dark mode:
* The background colors for the feature table and the installation comparison table are too bright. Darken them.
* Visited link color is low contrast. Make visited links brighter. Maybe also lighten unvisited links too.
* The demo page doesn't have dark mode. :cry:

website/index.html
```diff
@@ -112,6 +112,17 @@
...
+
+    @media (prefers-color-scheme: dark) {
+      a:link {color:#268BD2;}
+      a:visited {color:#6C71C4;}
+      body {background-color:  #002B36; color: white;}
+      td {color:black;}
+    }
+    @media (prefers-color-scheme: light) {
+      body {background:#fff; color: #333; }
+    }
```
strager, line 124: Could we move most of this into main.css?

strager, line 124: Nit: Format by running `yarn fmt`.

website/install.html
```diff
@@ -100,6 +100,19 @@
       .os-column-table > .windows {
         grid-area: windows;
       }
+
+    @media (prefers-color-scheme: dark) {
+      a:link {color:#268BD2;}
+      a:visited {color:#6C71C4;}
+      body {background-color:  #002B36; color: white;}
+      td {color:black;}
+      .macos-logo-color {filter: invert(85%);}
```
strager, line 109: Nit: Can we color-match the content logo from apple.com? https://www.apple.com/v/home/s/images/logos/watch-series-6/logo_hero__trsrpeu6koqy_medium_2x.png

website/install.html
```diff
@@ -1019,7 +1032,7 @@ <h2>
             alt="GNU/Linux"
             title="GNU/Linux"
           />
-          <img class="os-logo" src="macos.svg" alt="macOS" title="macOS" />
+          <img class="os-logo macos-logo-color" src="macos.svg" alt="macOS" title="macOS" />
```
strager, line 1035: Nit: I'd call this class just `macos-logo`. The fact that it changes the color is an implementation detail.

website/install.html
strager, line 109: Instead of creating a class, could we select based on the `src` attribute? For example, `.os-logo[src="macos.svg"]`

LeeWannacott: I color matched (it is pretty close to white) by figuring out the color of the logo which is #fafafa and then changing the svg file to the opposite hex value which is pretty close to black and then inverting that to get #fafafa.

website/index.html
LeeWannacott: I moved most of it into main  darkening table color's didn't seem to work when they were in main though.

website/install.html
LeeWannacott: I ended up doing this therefore not needing class macos-logo.

strager:
Blocking: demo/index.html: In dark mode, error highlights look bad.

Blocking: demo/index.html: In light mode, code is light gray and impossible to read.

Otherwise, looks good!

website/demo/index.html
```diff
@@ -55,6 +55,24 @@
...
+        text-decoration-style: wavy;
+      }
+      mark:empty::before {
+        color: red;
+        content: url("carat.svg") / "^";
+        content: url("carat.svg");
+        margin-left: -2px; /* (width of SVG) / 2 */
+        margin-top: 4px; /* height of SVG */
+        position: absolute;
+      }
```
strager, line 74: Blocking: Remove duplicated rules.

website/demo/index.html
```diff
@@ -82,6 +99,24 @@
...
+        a:visited {
+          color: #979ef8;
+        }
+        #code-input {
+          color: white;
+        }
+        body {
+          background-color: #002b36;
+          color: white;
+        }
```
strager, line 115: Isn't this already in main.css? Why duplicate the rules?

website/license.html
```diff
@@ -1,5 +1,25 @@
...
+        background-color: #002b36;
+        color: white;
+      }
+    }
+    @media (prefers-color-scheme: light) {
+      body {
+        background: #fff;
+        color: #333;
+      }
+    }
```
strager, line 21: Isn't this already in main.css? Why duplicate the rules?

website/main.css
```diff
@@ -75,3 +75,26 @@ main > h2:first-child {
...
+  }
+  .os-logo[src="macos.svg"] {
+    filter: invert(100%);
+  }
+}
+
+@media (prefers-color-scheme: light) {
+  body {
+    background: #fff;
+    color: #333;
```
strager, line 98: Minor: Why are we changing the text color of light mode? Making it explicitly black is fine, but you're making it off-black here.

website/main.css
```diff
@@ -75,3 +75,26 @@ main > h2:first-child {
...
+  a:visited {
+    color: #979ef8;
+  }
+  body {
+    background-color: #002b36;
+    color: white;
+  }
+  .os-logo[src="macos.svg"] {
+    filter: invert(100%);
+  }
```
strager, line 92: Fancy! :+1:

strager (changes requested):

website/main.css
LeeWannacott: Thanks for teaching how to use source. The filter invert is a bit of a hack : ].

website/main.css
LeeWannacott: Have made text explicitly black and background explicitly white. Have left prefers-color -scheme light in-case you want to change colors later on.

website/license.html
LeeWannacott: Duplicated rules have been deleted. I left <style></style> tags just in-case you end up changing style later on. However, probably unnecessary.

website/demo/index.html
LeeWannacott: Duplicated rules have been deleted.

website/demo/index.html
LeeWannacott: Duplicated rules have been deleted.

strager:
Merged as commit 8827a9868.

Issue #22, opened 2020-09-16
Error on invalid hex escape sequences: "hello\x1qworld"


strager:
String literals can contain hexadecimal escape sequences (starting with `\x`). If the escape sequence is invalid, quick-lint-js should report an error. Teach quick-lint-js about such errors in lex.cpp, error.h, and test-lex.cpp.

AngelJA:
Started a WIP PR: #156 

Just started probing though.  Hoping for some feedback on stream.

PR #156, opened 2021-02-03
Error on invalid hex escape sequences #22

AngelJA:
Resolves: #22

- add new error: type error_invalid_hex_escape_sequence
- span returned by the error always points to the escape prefix `\x`


src/lex.cpp
```diff
@@ -606,6 +606,10 @@ const char8* lexer::parse_string_literal() noexcept {
           ++c;
           break;
         }
+      case 'x':
+        // TODO (angel): do stuff here to identify invalid hex escape sequences?
+        // use parse_hexadecimal_number?
```
AngelJA, line 611: looked at parse_hexadecimal_number more closely and this doesn't seem likely, just wanted to avoid writing redundant code

strager (changes requested):
Discussed in person. Your test looks very wrong. The implementation looks like a good starting point.

AngelJA:
I think I've fixed my test. I lengthened it a bit to include a valid and some invalid hex escapes. I left a todo to add more tests. I'm not great at test writing, so I'll spend some time studying yours to make sure I have good coverage.

Also added a simple implementation.

I've written this to output something like node would when trying to execute, "invalid hex escape sequence". But I've noticed other linters points to a single character with something like "expected hexadecimal digit". Preference?

test/test-lex.cpp
```diff
@@ -649,10 +649,29 @@ TEST_F(test_lex, lex_strings) {
                                 offsets_matcher(input, 0, 14))));
       });
 
-  // TODO(strager): Report invalid hex escape sequences. For example:
```
strager, line 652: Also include a positive test case (around line 550).

src/lex.cpp
```diff
@@ -606,6 +606,16 @@ const char8* lexer::parse_string_literal() noexcept {
           ++c;
           break;
         }
+      case 'x':
+        for (int i = 0; i < 2; ++i) {
+          ++c;
+          if (!is_hex_digit(*c)) {
+            this->error_reporter_->report(error_invalid_hex_escape_sequence{
+                source_code_span(c-2-i, c+1)});
```
strager, line 614: Instead of computing `c-2-i`, store the start of the escape sequence in a variable outside the loop (e.g. after line 609).

test/test-lex.cpp
```diff
@@ -649,10 +649,29 @@ TEST_F(test_lex, lex_strings) {
...
-  //
-  // "hello\x1qworld"
-  // '\x'
+  // TODO (angel): add more tests
```
strager, line 652: Some edge cases I thought of:
```javascript
"\x
```
```javascript
"\x1
```
```javascript
"\x"
```
```javascript
"\x\xyz"
```

src/lex.cpp
AngelJA: done, it made sense to me to place it slightly further up, let me know if that's no good

test/test-lex.cpp
AngelJA: Added two strings with hex escapes and no errors, not sure if you had something else in mind.

test/test-lex.cpp
AngelJA: Added all these test cases.

AngelJA:
Made the changes mentioned  and left some comments.

For reference this test file: 
```javascript
let a = '\x;
let b = '\x1;
let c = '\x';
let d = '\x\xyz';
```
outputs this
```bash
test.js:1:10: error: invalid hex escape sequence: \x; [E060]
test.js:1:9: error: unclosed string literal [E040]
test.js:2:10: error: unexpected '\' in identifier [E043]
test.js:2:10: error: missing semicolon after expression [E027]
test.js:2:10: error: unexpected '\' in identifier [E043]
test.js:3:10: error: invalid hex escape sequence: \x' [E060]
test.js:4:10: error: invalid hex escape sequence: \x\ [E060]
test.js:4:12: error: invalid hex escape sequence: \xy [E060]
test.js:2:10: warning: use of undeclared variable: \x1 [E057]
```

test/test-lex.cpp
```diff
@@ -649,10 +651,77 @@ TEST_F(test_lex, lex_strings) {
...
+      });
+
+  this->check_tokens_with_errors(
+      u8"'\\x1", {token_type::string},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors,
+                    ElementsAre(
+                      ERROR_TYPE_FIELD(
+                        error_invalid_hex_escape_sequence, escape_sequence,
+                        offsets_matcher(input, 1, 5)),
```
strager, line 676: Blocking: 5 is out of bounds. The span should end at 4 instead.

Valid indexes are 0-3 (`'\x1`) and 4 (null terminator / end of file):
```
0: '
1: \
2: x
3: 1
4: (eof)
```

test/test-lex.cpp
```diff
@@ -649,10 +651,77 @@ TEST_F(test_lex, lex_strings) {
...
+                        offsets_matcher(input, 0, 3))
+                    )
+        );
+      });
+
+  this->check_tokens_with_errors(
+      u8"'\\x1", {token_type::string},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors,
+                    ElementsAre(
```
strager, line 673: Almost-blocking: The order of errors shouldn't matter. Use `UnorderedElementsAre` instead.

test/test-lex.cpp
```diff
@@ -649,10 +651,77 @@ TEST_F(test_lex, lex_strings) {
...
+      });
+
+  this->check_tokens_with_errors(
+      u8"'\\x'", {token_type::string},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors,
+                    ElementsAre(
+                      ERROR_TYPE_FIELD(
+                        error_invalid_hex_escape_sequence, escape_sequence,
+                        offsets_matcher(input, 1, 4))
```
strager, line 691: Minor: I'd expect the error squiggly to be on `\x` ([1-3)), not on `\x'` ([1-4)).

test/test-lex.cpp
```diff
@@ -649,10 +651,77 @@ TEST_F(test_lex, lex_strings) {
...
+      u8"'\\x\\xyz'", {token_type::string},
+      [](padded_string_view input, const auto& errors) {
+        EXPECT_THAT(errors,
+                    ElementsAre(
+                      ERROR_TYPE_FIELD(
+                        error_invalid_hex_escape_sequence, escape_sequence,
+                        offsets_matcher(input, 1, 4)),
+                      ERROR_TYPE_FIELD(
+                        error_invalid_hex_escape_sequence, escape_sequence,
+                        offsets_matcher(input, 3, 6))
```
strager, line 706: Hmm, these errors overlap. Is that a good idea? Not sure.

strager (approved):
Overall, looks good to me. Clean up the PR description and I'll merge it.

test/test-lex.cpp
AngelJA: Updated the tests where this applies.

test/test-lex.cpp
AngelJA: Updated to match this description. This also solves the out of bounds and overlap issues.

test/test-lex.cpp
AngelJA: This was resolved when the output was changed to only include `\x`.

test/test-lex.cpp
AngelJA: Some overlapping may be tough to avoid. For example, should invalid hex escape sequence also be raised inside an unclosed string?

Anyway, this error should never overlap anymore since it's been changed to always output `\x`.

AngelJA:
The changes mentioned in the last review have all been made. Just lmk if you want me to also implement more specific errors, like incomplete hex escape, in addition to invalid hex escape.

Issue #124, opened 2021-01-03
Report error: object keys cannot be template literals


strager:
The following code is illegal:

```js
let o = {`key`: value};
```

Report a nice diagnostic for this issue.

singalhimanshu:
Can you please point me to what files should I look at?

strager:
> Can you please point me to what files should I look at?

I think test-parse-expression.cpp and parse.cpp. In the latter, there should be a function called something like `parse_object_literal`.

strager:
@singalhimanshu How is progress on this task? Do you need any assistance?

alexanderkarlis:
@singalhimanshu Think you said last night it would be alright if I gave this one a crack? Thanks!

PR #157, opened 2021-02-04
fix for template literals in object keys

alexanderkarlis:
Hey, @strager . Made a PR for this [issue](https://github.com/quick-lint/quick-lint-js/issues/124). First time contributing to a c++ project. 

I'm a little unsure if the implementation is what's desired, with multiple errors being returned for this template literal in object (see tests). 

Tests passed locally, and formatted with `clang-format`. 

Let me know if I missed anything.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -849,6 +849,16 @@ expression_ptr parser::parse_object_literal() {
       QLJS_PARSER_UNIMPLEMENTED();
     }
     switch (this->peek().type) {
+    case token_type::complete_template:
+    case token_type::incomplete_template:
+      this->skip();
+      if (this->peek().type == token_type::colon) {
+        const char8 *backtick_location = this->lexer_.end_of_previous_token();
+        this->error_reporter_->report(error_unexpected_template_string{
+            source_code_span(backtick_location, backtick_location)});
+        this->skip();
+        break;
+      }
     case token_type::comma:
     case token_type::end_of_file:
     case token_type::right_curly:
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -746,7 +746,7 @@ next:
 template <class... Args>
 expression_ptr parser::parse_arrow_function_body(
     function_attributes attributes, const char8 *parameter_list_begin,
-    Args &&...args) {
+    Args &&... args) {
   function_guard guard = this->enter_function(attributes);
   if (this->peek().type == token_type::left_curly) {
     buffering_visitor *v = this->expressions_.make_buffering_visitor();
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -746,7 +746,7 @@ next:
 template <class... Args>
 expression_ptr parser::parse_arrow_function_body(
     function_attributes attributes, const char8 *parameter_list_begin,
-    Args &&... args) {
+    Args &&...args) {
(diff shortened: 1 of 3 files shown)


test/test-parse-expression.cpp
```diff
@@ -1442,6 +1442,21 @@ TEST_F(test_parse_expression, malformed_object_literal) {
                                 error_invalid_lone_literal_in_object_literal,
                                 where, offsets_matcher(p.locator, 1, 4))));
   }
+
+  {
+    test_parser p(u8"{`key`: 1};"_sv);
+    expression_ptr ast = p.parse_expression();
+    EXPECT_EQ(summarize(ast), "object(literal, ?)");
```
strager, line 1449: This looks unexpected to me. If we see a template literal followed by a `:`, I think we should recover by pretending that the template literal was a key. Therefore, this assertion should say something like `object(literal, literal)`.

test/test-parse-expression.cpp
```diff
@@ -1442,6 +1442,21 @@ TEST_F(test_parse_expression, malformed_object_literal) {
...
+    EXPECT_THAT(
+        p.errors(),
+        ElementsAre(
+            ERROR_TYPE_FIELD(error_unexpected_template_string, where,
+                             offsets_matcher(p.locator, 6, 6)),
```
strager, line 1454: Blocking: This assertion is saying that the error should occupy no characters (starting at character 6 but ending before character 6). Shouldn't the error span the template literal (starting at character but ending before character 6)?

src/parse.cpp
```diff
@@ -849,6 +849,16 @@ expression_ptr parser::parse_object_literal() {
       QLJS_PARSER_UNIMPLEMENTED();
     }
     switch (this->peek().type) {
+    case token_type::complete_template:
+    case token_type::incomplete_template:
```
strager, line 853: Blocking: If we encounter an `incomplete_template`, we should parse the template literal using e.g. `this->parse_template`. Otherwise, parsing will get out of sync, and eventually we'll treat the end of the template literal as the start of a different template literal.

Your tests don't trigger the `incomplete_template` code path. If they did, I think you'd see the problem I'm talking about. Example:

```js
({ `hello${k}`: v })
// ^^^^^^^^ incomplete_template
```

src/parse.cpp
```diff
@@ -849,6 +849,16 @@ expression_ptr parser::parse_object_literal() {
...
+      this->skip();
+      if (this->peek().type == token_type::colon) {
+        const char8 *backtick_location = this->lexer_.end_of_previous_token();
+        this->error_reporter_->report(error_unexpected_template_string{
+            source_code_span(backtick_location, backtick_location)});
+        this->skip();
+      }
```
strager, line 860: Let's crash if we don't see a colon.
```suggestion
      } else {
        QLJS_PARSER_UNIMPLEMENTED();
      }
```

strager (changes requested):
I think you're moving in the right direction with this PR. There are a few more details to work out, but it looks like you're touching the right code.

I think it'd be easier to talk about this PR in real time on Twitch. Please say hello on my stream! =]

strager:
@alexanderkarlis Do you plan on resuming work on this PR?

PR #158, opened 2021-02-05
Issue #22: Error on invalid hex escape sequences

AngelJA:
Resolves: #22

- add new error: type error_invalid_hex_escape_sequence
- span returned by the error always points to the escape prefix `\x`

(accidentally exploded the original PR and can't reopen, but here it is for reference: #156)


strager:
Landed as commit 9c5493cf6.

Issue #10, opened 2020-09-12
Parse escaped newlines in string and template literals


strager:
JavaScript string literals and template literals can contain backslash-escaped newlines:

```javascript
let s = "hello\
world";
let t = `hello\
world`;
```

Make sure these parse correctly as part of the string/template. Test in test-lex.cpp, fix in lex.cpp.

strager:
Actually, I realized this task isn't done. \r\n doesn't work!

```
$ hexdump -C hello.js
00000000  63 6f 6e 73 6f 6c 65 2e  6c 6f 67 28 22 68 65 6c  |console.log("hel|
00000010  6c 6f 5c 0d 0a 77 6f 72  6c 64 22 29 3b           |lo\..world");|
0000001d
$ node hello.js
helloworld
$ ./build/quick-lint-js hello.js
hello.js:1:13: error: unclosed string literal
../src/parse.cpp:413: internal check failed in parse_expression_remainder: this->peek().type == token_type::right_paren
zsh: illegal hardware instruction  ./build/quick-lint-js hello.js
```

strager:
I knew there was more to this task than met the eye. I just forgot what the gotcha was. =]

AngelJA:
@strager Just wanted to make sure I'm reproducing the problem correctly here. In vscode with end of line sequence set to CRLF.
![image](https://user-images.githubusercontent.com/3516571/107107386-7b601c00-67e5-11eb-8d82-ee2e6eddd3da.png)

This should report no error, right?

AngelJA:
Just to clarify, templates don't require the escaped newline like strings, so the rules are a bit different.
```javascript
# valid
let a = `hello
         world`;
```

strager:
This was implemented by @AngelJA.

PR #159, opened 2021-02-06
Allow escaped newlines with CRLF

AngelJA:
Resolves a lingering issue in #10 

Currently in a multi-line string literal, if the line ends with CRLF, quick-lint incorrectly reports 'unclosed string literal'.
```javascript
let a = 'hello \
         world.'
```
This removes the error.

Also add a test to ensure that all line ending types are allowed in a string literal if escaped with a backslash.

### DIFF
--- a/src/lex.cpp
+++ b/src/lex.cpp
@@ -607,6 +607,13 @@ const char8* lexer::parse_string_literal() noexcept {
           ++c;
           break;
         }
+        break;
+      case '\r':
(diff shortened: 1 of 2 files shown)


AngelJA:
Updated test to check all `line_terminators` as well as single and double quotes. Before this fix, only the CRLF case behaved incorrectly. Now, none of these strings report an error.
image

PR #208, opened 2021-04-06
Report warnings for misuse of --lsp-server

### LINKED ISSUE #202: 7$: Report warnings for misuse of --lsp-server
If `--lsp-server` is combined with options like `--output-format` and `--vim-file-bufnr`, issue a warning explaining that `--output-format` won't be used.

LiraNuna:
We should warn when the option `--lsp-server` and one of the output formats are chosen since the option is incompatible and therefore ignored.
To detect if an option was given, a new enum value `default_format` was added.

There are no tests for this, mainly since I couldn't find a test that targets the `o.files_to_lint` check right below, nor a test for `handle_options` at all.

Feedback about this approach is welcomed.

Fixes #202

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,6 +39,7 @@ class any_error_reporter {
   static any_error_reporter make(output_format format,
                                  compiled_error_list *exit_fail_on) {
     switch (format) {
+    case output_format::default_format:
     case output_format::gnu_like:
       return any_error_reporter(error_tape<text_error_reporter>(
           text_error_reporter(std::cerr), exit_fail_on));
--- a/test/test-options.cpp
+++ b/test/test-options.cpp
@@ -35,7 +35,7 @@ TEST(test_options, default_options_with_no_files) {
   EXPECT_FALSE(o.help);
   EXPECT_FALSE(o.version);
   EXPECT_FALSE(o.lsp_server);
-  EXPECT_EQ(o.output_format, output_format::gnu_like);
+  EXPECT_EQ(o.output_format, output_format::default_format);
   EXPECT_THAT(o.files_to_lint, IsEmpty());
 }
 
@@ -85,6 +85,12 @@ TEST(test_options, debug_parser_visits_shorthand) {
 
 TEST(test_options, output_format) {
   {
+    options o = parse_options({});
+    EXPECT_THAT(o.error_unrecognized_options, IsEmpty());
+    EXPECT_EQ(o.output_format, output_format::default_format);
+  }
+
+  {
     options o = parse_options({"--output-format=gnu-like"});
     EXPECT_THAT(o.error_unrecognized_options, IsEmpty());
     EXPECT_EQ(o.output_format, output_format::gnu_like);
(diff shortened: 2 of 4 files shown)


LiraNuna:

src/main.cpp
```diff
@@ -141,6 +142,11 @@ void handle_options(quick_lint_js::options o) {
     std::exit(EXIT_FAILURE);
   }
   if (o.lsp_server) {
+    if (o.output_format != output_format::default_format) {
+      std::cerr << "warning: output format selection is ignored in "
+                   "--lsp-server mode\n";
+    }
```
strager, line 148: Minor: Put this logic in `options::dump_errors` (called on line 141 above).

strager (approved):
Let me know when I should merge this PR.

LiraNuna:
Ready to merge (I would recommend re-review).

test/test-options.cpp
```diff
@@ -314,6 +320,41 @@ TEST(test_options, dump_errors) {
...
+  }
+
+  {
+    options o;
+    o.lsp_server = true;
+    o.output_format = output_format::gnu_like;
+
+    std::ostringstream dumped_errors;
+    bool have_errors = o.dump_errors(dumped_errors);
+    EXPECT_TRUE(have_errors);
```
strager, line 342: Shouldn't `have_errors` be false, since we only saw warnings?

test/test-options.cpp
```diff
@@ -314,6 +320,41 @@ TEST(test_options, dump_errors) {
...
+    options o;
+    o.lsp_server = true;
+    o.output_format = output_format::vim_qflist_json;
+
+    std::ostringstream dumped_errors;
+    bool have_errors = o.dump_errors(dumped_errors);
+    EXPECT_TRUE(have_errors);
+    EXPECT_EQ(dumped_errors.str(),
+              "warning: output format ignored with --lsp-server\n");
+  }
```
strager, line 357: Minor: Instead of repeating the code, loop over different `output_format`s?

test/test-options.cpp
```diff
@@ -314,6 +320,41 @@ TEST(test_options, dump_errors) {
...
+  {
+    options o;
+    o.lsp_server = true;
+    o.output_format = output_format::gnu_like;
+
+    std::ostringstream dumped_errors;
+    bool have_errors = o.dump_errors(dumped_errors);
+    EXPECT_TRUE(have_errors);
+    EXPECT_EQ(dumped_errors.str(),
+              "warning: output format ignored with --lsp-server\n");
```
strager, line 344: Nit: Reword to: "warning: --output-format ignored with --lsp-server"

strager (changes requested):

test/test-options.cpp
```diff
@@ -314,6 +320,37 @@ TEST(test_options, dump_errors) {
...
+  }
+
+  {
+    const std::array<output_format, 2> output_formats = {
+        /* default_format intentionally left out */
+        output_format::gnu_like,
+        output_format::vim_qflist_json,
+    };
+
+    for (const auto &format : output_formats) {
```
strager, line 342: Nit: You can inline `output_formats`:

```cpp
    for (const auto &format : {
        /* default_format intentionally left out */
        output_format::gnu_like,
        output_format::vim_qflist_json,
    }) {
```

strager (approved):
:heart:

test/test-options.cpp
LiraNuna: did not know that! modern C++ is neat.

LiraNuna:
Ready

strager:
Landed as commit 20c2886abf96950aedc121decbf5804bd8fe9c56.

Issue #179, opened 2021-03-24
7$: Missing syntax error: x += ;


strager:
The following code has a syntax error, but quick-lint-js doesn't emit an error:
```javascript
let x = 3;
x += ;
```

keyehzy:
I'm trying this one out and should send a PR soon.

strager:
@keyehzy fixed this issue in commit 02fce249.

PR #209, opened 2021-04-06
reports error message when missing rhs of compound assignment operator

keyehzy:
Attempts to fix issue #179.

I'm not entirely confident that just reporting an invalid right-hand side should catch all the cases, but for all the compound assignment operators it seems to work and does not conflict with previous tests.

### DIFF
--- a/test/test-parse-expression-statement.cpp
+++ b/test/test-parse-expression-statement.cpp
@@ -132,6 +132,20 @@ TEST(test_parse, parse_invalid_math_expression) {
             ERROR_TYPE_FIELD(error_unmatched_parenthesis, where,
                              offsets_matcher(&code, strlen(u8"2 * "), u8"("))));
   }
+
+  {
+    padded_string code(u8"x += ;"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    p.parse_and_visit_expression(v);
+    EXPECT_THAT(v.visits,
+                ElementsAre("visit_variable_use", "visit_variable_assignment"));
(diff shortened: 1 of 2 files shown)


keyehzy:

test/test-parse-expression-statement.cpp
```diff
@@ -132,6 +132,21 @@ TEST(test_parse, parse_invalid_math_expression) {
...
+    parser p(&code, &v);
+    p.parse_and_visit_expression(v);
+    EXPECT_THAT(v.visits,
+                ElementsAre("visit_variable_use", "visit_variable_assignment"));
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"x"}));
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_missing_operand_for_operator, where,
+                    offsets_matcher(&code, strlen(u8"x += "), u8";"))));
```
strager, line 148: For `x + ;`, we report the error on the `+`. (See line 64 for an example.) Shouldn't we report the error on the `+=` here (not on the `;`)?

strager (changes requested):

test/test-parse-expression-statement.cpp
keyehzy: Yes, absolutely. Using code_span of rhs was wrong, I'm using code_span of the operator instead.

strager (approved):

strager:
Landed to master as commit 02fce249.

Issue #189, opened 2021-03-31
LibreJS support for website


strager:
https://www.gnu.org/software/librejs/free-your-javascript.html

PR #212, opened 2021-04-07
Added LibreJS support to website

Jasper-Hawks:
Specified the license in both quick-lint-js.mjs and editor.mjs
Fixed Issue: #189

### DIFF
--- a/website/demo/editor.mjs
+++ b/website/demo/editor.mjs
@@ -1,3 +1,4 @@
+// @license magnet:?xt=urn:btih:1f739d935676111cfff4b4693e3816e664797050&dn=gpl-3.0.txt GPL-v3-or-Later
 // Copyright (C) 2020  Matthew Glazar
 // See end of file for extended copyright information.
 
@@ -219,3 +220,4 @@ function wrapNodes(wrapperElement, firstChildNode, lastChildNode) {
 //
 // You should have received a copy of the GNU General Public License
 // along with quick-lint-js.  If not, see <https://www.gnu.org/licenses/>.
+// @license-end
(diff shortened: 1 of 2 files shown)


Jasper-Hawks:

strager:
Looks good. Thanks for taking a stab at this task.

Did you test that the LibreJS browser extension does the right thing? Can you screenshot its UI?

What about the .wasm file? Does it need similar annotation?

Jasper-Hawks:
The program list and blocks the scripts that are not compatible with LibreJS, if all the scripts are compatible then no scripts will be listed and they will all be executed.
librejs

Jasper-Hawks:
The .wasm file doesn't appear to be blocked or accepted

strager:
> The program list and blocks the scripts that are not compatible with LibreJS, if all the scripts are compatible then no scripts will be listed and they will all be executed.

What's it look like with your patch?

> The .wasm file doesn't appear to be blocked or accepted

That's probably because the .mjs files load the .wasm files without using `<script>` or `import`.

Jasper-Hawks:
None of the scripts are blocked on any of the pages. I'll add the license tags to the .wasm file.

Jasper-Hawks:
I'm looking through the files and I only see wasm-demo.h and wasm-demo.cpp. Would you like me to add the license tags to both?

strager:
> I'm looking through the files and I only see wasm-demo.h and wasm-demo.cpp. Would you like me to add the license tags to both?

`wasm-demo.h` and `wasm-demo.cpp` are not served to web clients.

See website/demo/README.md for instructions on building the .wasm file.

> None of the scripts are blocked on any of the pages.

Okay. If annotating the .wasm file is unnecessary, I guess we don't need to do it.

I'll try out this patch tomorrow to get a feel for how LibreJS works.

strager:
This patch doesn't seem to fully work. Some scripts are still blocked, so the demo doesn't work with LibreJS installed:

Screenshot from 2021-04-08 15-04-12

strager (changes requested):
.

strager:
@Jasper-Hawks Are you still working on this PR?

Jasper-Hawks:
No. I've given up.

PR #218, opened 2021-04-09
warn when --exit-fail-on is used with --lsp-server

LiraNuna:
Yet another parameter that should warn when `--lsp-server` is used.
A part of https://github.com/quick-lint/quick-lint-js/pull/208 and https://github.com/quick-lint/quick-lint-js/issues/202

### DIFF
--- a/src/error-list.cpp
+++ b/src/error-list.cpp
@@ -207,6 +207,10 @@ bool compiled_error_list::is_present(const char* error_code) const noexcept {
   }
   return present;
 }
+
+bool compiled_error_list::is_user_provided() const noexcept {
+  return !parsed_error_lists_.empty();
+}
 }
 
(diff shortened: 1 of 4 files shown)


src/error-list.cpp
```diff
@@ -207,6 +207,10 @@ bool compiled_error_list::is_present(const char* error_code) const noexcept {
   }
   return present;
 }
+
+bool compiled_error_list::empty() const noexcept {
```
strager, line 211: This name is misleading. An empty `compiled_error_list` would mean that no error code `is_present`.

I'd call this something like `did_add_error_list` or `number_of_added_error_lists`? (I'm very unsure. =S)

strager (changes requested):
Other than the function name, this looks good to me.

strager (approved):

PR #232, opened 2021-04-16
Website: Synchronize demo's pre and textarea sizing

erlliam:
https://user-images.githubusercontent.com/32761424/115054415-2f4cea00-9eae-11eb-9034-0e7e471d6f07.mp4

https://user-images.githubusercontent.com/32761424/115054442-34aa3480-9eae-11eb-9bc6-a98fba9f0a50.mp4

There seems to be some misalignment if you make the height too small. (potential "bug" not worth caring about)

I wasn't sure if the comment from the known bugs section referred to the fix I've implemented so I gave it it's own commit.
If it is referring to the fix, that commit should be fixup'd. If not, that commit should be dropped.

### DIFF
--- a/website/demo/index.html
+++ b/website/demo/index.html
@@ -100,7 +100,6 @@
     * Scrolling in general works poorly when sub-pixel scrolling (e.g. when
       zoomed in).
     * Scrolling horizontally sometimes doesn't scroll the highlights.
-    * Resizing the editor breaks highlights.
     * Browser text search finds text in the shadow code input in addition to
       text in the user code input.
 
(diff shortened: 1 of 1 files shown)


website/demo/index.html
```diff
@@ -100,7 +100,6 @@
     * Scrolling in general works poorly when sub-pixel scrolling (e.g. when
       zoomed in).
     * Scrolling horizontally sometimes doesn't scroll the highlights.
-    * Resizing the editor breaks highlights.
```
strager, line 103: Yeah, this is the issue you fixed.

strager (approved):
Looks great!

erlliam:
I've rebased. Should be good to go.

PR #237, opened 2021-04-18
LSP: Prevent deadlock with synchronous client

strager:
If an LSP client sends a bunch of requests, but does not read from
quick-lint-js' stdout, quick-lint-js can hang in pipe_reader::write.
(This came up when writing some benchmarks. My LSP client naively made a
bunch of requests without reading responses during the requests.)

Prevent this (unlikely) deadlock by switching stdout to non-blocking
mode, buffering data in memory, and resuming reading from stdin as soon
as possible.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,6 +13,8 @@ option(
   FALSE
 )
 
+find_package(Threads REQUIRED)
+
 quick_lint_js_add_executable(
   quick-lint-js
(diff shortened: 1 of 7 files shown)


strager:
I'm going to punt on this PR. Fixing this issue isn't worth the complexity (Linux) or the [theorized] performance hit (Windows). Saved my progress to refs/archive/lsp-deadlock.

Issue #236, opened 2021-04-18
[VS Plugin] vs.windows.onDidChangeTextDocument activates twice on first input


lifeinData:
**OS:** Windows 10 Enterprise 19042.928
**Application:** VS Code 10 v1.55.2
When you open a new or existing js file with quicklintjs as your linter and you type the first character, vs.windows.onDidChangeTextDocument activates twice. We only expect it to be activated 1 time for your first character input.

strager:
This is fixed, right? @lifeinData

PR #239, opened 2021-04-23
Fix for vscode.workspace.onDidChangeTextDocument running twice on first character input

lifeinData:
[issue being addressed](https://github.com/quick-lint/quick-lint-js/issues/236)

Triggering the onDidChangeTextDocument  twice when inputting a character for the first time is a known issue. Maintainer suggests to check whether the contentChange is empty, which is what we did here.

See [scode.workspace.onDidChangeTextDocument is called with no changes](https://github.com/Microsoft/vscode/issues/50344)


plugin/vscode/extension.js
```diff
@@ -373,15 +373,17 @@ async function activateAsync() {
...
-      logAsyncErrors(
-        (async () => {
-          if (isLintable(event.document)) {
-            await linters
-              .getLinter(event.document)
-              .textChangedAsync(event.contentChanges);
-          }
-        })()
-      );
+      if (event.contentChanges.length !== 0){
```
strager, line 376: Do you know why our onDidChangeTextDocument callback is called? Is it possible that modifications were made on the document anyway and we need to sync state by looking at the document's full text? Or is it really just a fluke?

strager (approved):

plugin/vscode/extension.js
lifeinData: Quoting from [official doc](https://code.visualstudio.com/api/references/vscode-api#TextDocumentChangeEvent), "This usually happens when the contents changes but also when other things like the dirty-state changes.". 

I think it may have something to do with the fact that our document in question has changed from the state of "no qljs attached" to "qljf attached" causing a event to occur but it does not record this event in **.contentChanges**

**.contentChanges** only refers to any character deletion or addition events and it is guaranteed to fire one on any of those events (tested)

strager: Should we lint if we get a dirty-state change? What is a dirty-state change anyway?

lifeinData: What is a dirty-state change anyway?
There's an answer about "dirty object" [here](https://stackoverflow.com/questions/1745888/what-is-meant-by-the-term-dirty-object). The last line is the most relevant " It's a typical situation of desynchronization between a cache and a storage." 

From my digging of the [globalState in vscode](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext), the "storage" is a simple db file that is created in AppData. globalState has all states of the vscode process with respect to the extensions that are running. This is initial state is **cached** in RAM (undocumented). When a vscode extension changes it state, it **does not ** immediately update the vscode db file hence a desync between the cache and storage when a vscode extension changes its properties which we can say is a change in "dirty-state" by the definition in the stack overflow post.

lintVisibleEditorsAsync is the thing that lints the file immediately upon opening and therefore, onDidChangeTextDocument _should only_ be concerned with character addition/deletion because your other lovely functions take care of the other cases.

strager (changes requested):
As discussed in person, I'm concerned that this change might break linting in some other case. Let's get to the bottom of why the event is being generated so we have some confidence that this won't have a negative impact anywhere.

plugin/vscode/extension.js
lifeinData: Triggering the onDidChangeTextDocument  twice when inputting a character for the first time is a known issue. Maintainer suggests to check whether the contentChange is empty, which is what we did here.

See [scode.workspace.onDidChangeTextDocument is called with no changes](https://github.com/Microsoft/vscode/issues/50344)

Please resolve conversation if you're good with it 👍

plugin/vscode/extension.js
```diff
@@ -4,6 +4,7 @@
 "use strict";
 
 let assert = require("assert");
+const { debug } = require("console");
```
strager, line 7: Nit: Remove this.
```suggestion
```

plugin/vscode/extension.js
```diff
@@ -373,15 +374,22 @@ async function activateAsync() {
...
-            await linters
-              .getLinter(event.document)
-              .textChangedAsync(event.contentChanges);
-          }
-        })()
-      );
+      /*
+      Event fires twice on first character input event of the open file. 
+      First one is caused by dirty-state change and the secibd is the actual character change.
+      Below is the implementation of a suggested workaround by maintainer: https://github.com/Microsoft/vscode/issues/50344
```
strager, line 380: > Below is the implementation
What is this comment referring to? The next line? The next 10 lines? I suggest being more descriptive. Perhaps something like this:
```js
// comment, yada yada. talks about isBogusEvent.
let isBogusEvent = event.contentChanges.length === 0;
if (!isBogusEvent) {
  ...
}
```

strager (approved):

lifeinData:
Hello strager,

I'd like to claim this issue.

strager:
Landed as commit 1f032d1f61f68e2e517f0c591126f71dda1c9351.

Issue #194, opened 2021-04-04
7$: CI: Detect broken links on website


strager:

backbugkaught:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

backbugkaught:
```python
#!/usr/bin/env python3
import requests
import re
from bs4 import BeautifulSoup
from urllib.parse import urljoin, urlparse
import argparse

"""
(error_message), url, broken_url
"""

mail_whitelist = "whitelist_mail.txt"
hash_urls = []


def is_mailto_link(link) -> bool:
    return True if re.search(r"^mailto:*", link) else False


def readfile(filename) -> list:
    try:
        with open(filename, "r") as f:
            return list(map(lambda x: x.strip("\n"), f.readlines()))
    except BaseException:
        return []


def in_mail_whitelist(link, filename=mail_whitelist) -> bool:
    return True if link in readfile(filename) else False


def in_namespace(namespace, url) -> bool:
    return urlparse(url).netloc == namespace


def get_urls_from_page(url) -> list:
    req = requests.get(url)
    soup = BeautifulSoup(req.text, "html.parser")
    href_tags = soup.find_all(href=True)
    hrefs = [tag.get("href") for tag in href_tags]
    return hrefs


def return_mail_address(link) -> str:
    return link.split(":")[1]


def crawl_and_report(parent_url, site_namespace, urls):
    # report broken urls on parent page
    # if url is in "namespace" then crawl that page
    for i in urls:
        if is_mailto_link(i):
            mail_link = return_mail_address(i)
            if not in_mail_whitelist(mail_link):
                print(f"(unknown mail) {parent_url}, {mail_link}")
        else:
            url = urljoin(parent_url, i)
            if url not in hash_urls:
                hash_urls.append(url)
                result = requests.get(url)
                if result.ok and in_namespace(site_namespace, result.url):
                    urls_from_page = get_urls_from_page(result.url)
                    if len(urls_from_page) != 0:
                        crawl_and_report(
                            result.url, site_namespace, urls_from_page)
                elif not result.ok:
                    print(
                        f"({result.status_code} error) {parent_url}, {url}")


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--site", type=str)
    args = parser.parse_args()
    if args.site:
        site = args.base_url
    else:
        site = "https://quick-lint-js.com/"
[...]

PR #242, opened 2021-04-24
check broken links on website #194

emadflash:


tools/check-links
```diff
@@ -0,0 +1,143 @@
...
+        except requests.exceptions.ConnectionError:
+            print("(error) failed to get", self.site)
+
+    def in_namespace(self, url) -> bool:
+        return urlparse(url).netloc == self.site_namespace
+
+    def get_urls_from_page(self, soup) -> list:
+        href_tags = soup.find_all(href=True)
+        hrefs = [tag.get("href") for tag in href_tags]
+        return hrefs
```
strager, line 70: We should also crawl scripts referenced by `<script src=...>`. See website/benchmarks/index.html for an example.

tools/check-links
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
```
strager, line 1: Include copyright comments at the top and bottom of this file. See other Python code for examples.

tools/check-links
```diff
@@ -0,0 +1,143 @@
...
+                        if soup is not None and not check_fragment(soup, fragment):
+                            print("(fragment missing)", parent_url, url, fragment)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("site", type=str)
+    if len(sys.argv) == 1:
+        parser.print_help()
+        exit(1)
```
strager, line 140: Doesn't line 141 do this checking automatically?

strager (approved):

tools/check-links
```diff
@@ -0,0 +1,217 @@
...
+            result = requests.get(self.site)
+            soup = BeautifulSoup(result.text, "html.parser")
+            self.urls = self.get_urls_from_page(soup)
+        except requests.exceptions.ConnectionError:
+            print("(error) failed to get", self.site)
+            sys.exit(1)
+
+
+
+class Crawler(initialize_crawler):
```
strager, line 93: Merge this class with `initialize_crawler`.

tools/check-links
```diff
@@ -0,0 +1,217 @@
...
+            except UrlNotFound as e:
+                self.report_error(f"{e.response_code} error", url_packet)
+
+        with Pool(max_workers=len(urls)) as executor:
+            future_to_url = {executor.submit(is_url_broken, packet) for packet in urls}
+            for future in future_to_url:
+                try:
+                    data = future.result()
+                except Exception as e:
+                    pass
```
strager, line 154: Almost-blocking: Catching `Exception` will hide a lot of bugs. Why are we catching exceptions here at all? When would we expect `is_url_broken` to raise an exception?

tools/check-links
```diff
@@ -0,0 +1,217 @@
...
+
+
+def is_mailto_link(link) -> bool:
+    return True if re.search(r"^mailto:*", link) else False
+
+
+def check_mail_link(link: str, parent: str) -> None:
+    mail = link.partition(":")[-1]
+    if mail not in allow_mails:
+        print(f"(unknown mail) {parent}, {mail}")
```
strager, line 48: Almost-blocking: If an invalid email is found, this script will still exit successfully (code 0). I think we should call `Crawler.report_error` here instead of `print`ing directly.

emadflash:

tools/check-links
```diff
@@ -87,9 +87,9 @@ class Crawler:
     def get_urls_from_page(self, soup) -> list:
         href_tags = soup.find_all(href=True)
         hrefs = [tag.get("href") for tag in href_tags]
-        scripts = list(map(lambda x: x.get('src'), soup.findAll('script')))
-        if None not in scripts:
-            hrefs.extend(scripts)
+        script_srcs = list(map(lambda x: x.get("src"), soup.findAll("script")))
+        if None not in script_srcs:
+            hrefs.extend(script_srcs)
```
strager, line 92: Almost-blocking: This looks broken. If a document contains one `<script>` with no src and one `<script>` with a src, then this logic will ignore the `<script>` with a src. Why would we want to exclude the `<script>` with a src?

tools/check-links
```diff
@@ -137,15 +137,18 @@ class Crawler:
...
             for future in future_to_url:
                 packet = future_to_url[future]
                 try:
                     future.result()
                 except UrlNotFound as e:
-                    self.report_error(f"{e.response_code} error", packet)
-                except Exception as exc:
-                    print('%r generated an exception: %s' % (packet.url, exc))
+                    if not requests.get(packet.url).ok:
+                        self.report_error(f"{e.response_code} error", packet)
```
strager, line 149: Minor: Why is the retry logic here and not next to where the request is made (line 62)? This seems like an odd place to retry.

tools/check-links
```diff
@@ -105,8 +105,8 @@ class Crawler:
 
     def get_urls_to_crawl(self, packet) -> list:
         urls_from_page = list()
-        if packet.url[-1] == '/':
-            packet.url = urljoin(packet.url, 'index.html')
+        if packet.url[-1] == "/":
+            packet.url = urljoin(packet.url, "index.html")
```
strager, line 109: Almost-blocking: This puzzled me at first glance.

* We're changing `packet.url`, but the URL has already been used to download the page. Therefore, we won't actually request `/foo/index.html`. But the code makes you think that we'd request `/foo/index.html`.
* `url` and `defraged_url` are out of sync after this mutation.

If the goal of these lines is to influence line 110's logic, why not change the `in_allowed_file_soup` function instead? Is there some non-local side effect of this code that I'm not seeing?

strager:
The following can be done after merging this PR:
- [ ] mode to check only internal links
- [ ] allow multiple input URLs
- [ ] run on CI

strager:
Let me know when you want me to merge this. I'm willing to merge it as-is or after you make tweaks.

emadflash:
> I'm willing to merge it as-is or after you make tweaks.

After I make tweaks.

strager (approved):

tools/check-links
```diff
@@ -97,11 +97,11 @@ class Crawler:
             hrefs.extend(script_srcs)
         return hrefs
 
-    def in_allowed_file_soup(self, url) -> bool:
-        if url[-1] == "/":
+    def in_allowed_file_soup(self, url_response) -> bool:
+        if url_response.url[-1] == "/":
             return True
```
strager: This check isn't needed, right? Don't requests to `localhost/benchmarks/` return `content-type: text/html` already?

strager (approved):

Issue #196, opened 2021-04-04
22$: Add performance tracing to VS Code plugin


strager:
Idea for an append-only recoverable format:

```
timestamp file-id file-path-or-duration
```
```
10000 1 index.js
10010 1 12451
10011 1 412
10011 1 41231
10012 1 3121
10012 1 2415
25000 2 other.js
30201 2 12812
30240 1 1412
```

lifeinData:
I'm working on this. Will let you know if I'm completely stuck.

strager:
A simpler format would be something like this:
```
10010 12451 index.js
10011 412 index.js
10011 41231 index.js
10012 3121 index.js
10012 2415 index.js
30201 12812 other.js
30240 1412 index.js
```
I suggested the more complicated format earlier because it'd produce smaller files. (Repeating the file name, especially if it is long, might produce relatively big files.) `file-id` was a way to compress the file name into a number.

lifeinData:
currently tracking on #244

PR #244, opened 2021-04-24
Performance tracing implementation for vscode plugin

lifeinData:
Issue Ref: #196
Implementation details:
Satisfies requirements to trace performance of linter and have the format specified in the original issue. Every new file using qljs is indexed via persistent[ vscode extension property global storage utility method](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalState)

Added a new column called "event-id". This is meant to be a utility column that can store events that are occurring within qljs other than performance of parser. For example, currently we only have 2 different event-id, a "n" which denotes a new file has qljs attached. Another potential one could be if the file name changes, currently it still detects this as a new file but we could make another event name for this. The other event-id we have is "blank" which measure the speed of parser.lint() (i.e. every character addition/deletion + initial parse when file open)

### DIFF
--- a/plugin/vscode/extension.js
+++ b/plugin/vscode/extension.js
@@ -5,12 +5,15 @@
 
 let assert = require("assert");
 let vscode = require("vscode");
+let process = require("process");
+
 let {
   DiagnosticSeverity,
   ProcessCrashed,
   createProcessFactoryAsync,
 } = require("quick-lint-js-wasm/quick-lint-js.js");
 
+let PerformanceWriter = require("./performance-writer");
 // TODO(strager): Allow developers to reload the .wasm file.
 let processFactoryPromise = createProcessFactoryAsync();
 
@@ -48,10 +51,11 @@ class DocumentLinterDisposed extends Error {}
 exports.DocumentLinterDisposed = DocumentLinterDisposed;
 
 class DocumentLinter {
-  constructor(document, diagnosticCollection) {
+  constructor(document, diagnosticCollection, performanceWriter) {
     this._document = document;
     this._diagnosticCollection = diagnosticCollection;
     this._state = DocumentLinterState.NO_PARSER;
+    this._performanceWriter = performanceWriter;
 
     // Used only in states: CREATING_PARSER
     this._parserPromise = null;
@@ -182,8 +186,12 @@ class DocumentLinter {
           }
           this._pendingChanges.length = 0;
           // END CRITICAL SECTION (no awaiting above)
-
+          let startTime = process.hrtime.bigint();
           let diags = this._parser.lint();
+          this._performanceWriter.initiateLogging(
+            process.hrtime.bigint() - startTime,
+            this._document.fileName
+          );
           this._updateDocumentDiagnostics(diags);
         } catch (e) {
           // END CRITICAL SECTION (no awaiting above)
@@ -226,8 +234,7 @@ class DocumentLinter {
       this._pendingChanges.length = 0;
       this._state = DocumentLinterState.PARSER_LOADED;
       // END CRITICAL SECTION (no awaiting above)
-
-      let diags = this._parser.lint();
+      let diags = this._parser.lint(this._performanceWriter);
       this._updateDocumentDiagnostics(diags);
     } catch (e) {
       if (e instanceof ProcessCrashed) {
@@ -322,9 +329,9 @@ class DocumentLinter {
 }
 
 class DocumentLinterCollection {
-  constructor(diagnosticCollection) {
+  constructor(diagnosticCollection, performanceWriter) {
     this._diagnosticCollection = diagnosticCollection;
-
+    this._performanceWriter = performanceWriter;
     // Mapping from URI string to DocumentLinter.
     this._linters = new Map();
   }
@@ -333,7 +340,11 @@ class DocumentLinterCollection {
     let documentURIString = document.uri.toString();
     let linter = this._linters.get(documentURIString);
     if (typeof linter === "undefined") {
-      linter = new DocumentLinter(document, this._diagnosticCollection);
+      linter = new DocumentLinter(
+        document,
+        this._diagnosticCollection,
+        this._performanceWriter
+      );
       this._linters.set(documentURIString, linter);
     }
     return linter;
@@ -358,6 +369,7 @@ class DocumentLinterCollection {
   dispose() {
     logAsyncErrors(async () => {
       await this.disposeAsync();
+      this._performanceWriter.writeToFile();
     });
   }
 }
@@ -365,11 +377,11 @@ exports.DocumentLinterCollection = DocumentLinterCollection;
 
 let toDispose = [];
 
-async function activateAsync() {
+async function activateAsync(ExtensionContext) {
   let diagnostics = vscode.languages.createDiagnosticCollection();
   toDispose.push(diagnostics);
-
-  let linters = new DocumentLinterCollection(diagnostics);
+  let performanceWriter = new PerformanceWriter(ExtensionContext);
+  let linters = new DocumentLinterCollection(diagnostics, performanceWriter);
   toDispose.push(linters);
 
   toDispose.push(
--- a/plugin/vscode/performance-writer.js
+++ b/plugin/vscode/performance-writer.js
@@ -1,0 +1,114 @@
+// Copyright (C) 2020  Matthew Glazar
+// See end of file for extended copyright information.
+
+const fs = require("fs");
+const path = require("path");
+const process = require("process");
+
+class PerformanceWriter {
+  constructor(extensionContext) {
+    this._extensionContext = extensionContext;
+    this._dataAttributes = {
+      columns: ["timestamp", "file-ref", "event-id", "file-name-or-duration"],
+    };
+    this._tmpLogLines = [];
+    this._fileIndex = 0;
+    this._writeToFileTimer = null;
+    this._timerActivated = false;
+    if (!fs.existsSync(this._extensionContext.logUri.fsPath)) {
+      fs.mkdirSync(this._extensionContext.logUri.fsPath);
+      console.log(this._extensionContext.logUri.fsPath);
+      this.writeHeaderLine();
+    }
+    this._performanceTxtPath = path.join(
+      this._extensionContext.logUri.fsPath,
+      "qljsPerformanceTrace.txt"
+    );
+    this._writableStream = fs.createWriteStream(this._performanceTxtPath, {
+      encoding: "utf8",
+      flags: "a",
+    });
+    this._currentSessionMap = {};
+  }
+
+  fileNameOptimizer(fileName) {
+    if (
+      Object.prototype.hasOwnProperty.call(this._currentSessionMap, fileName)
+    ) {
+      this._fileIndex = this._currentSessionMap[fileName];
+    } else {
+      this._fileIndex++;
+      this._currentSessionMap[fileName] = this._fileIndex;
+      this.writeNewFileLine(fileName);
+    }
+  }
+
+  initiateLogging(timeTaken, fileName) {
+    this.fileNameOptimizer(fileName);
+    this.writeExistingFileLine(timeTaken);
+    this.setWriteToFileTimer();
+  }
+
+  writeHeaderLine() {
+    this._tmpLogLines.push(...this._dataAttributes.columns);
+  }
+
+  writeNewFileLine(fileName) {
+    this._tmpLogLines.push(Date.now(), this._fileIndex, "n", fileName);
+  }
+
+  writeExistingFileLine(timeTaken) {
+    this._tmpLogLines.push(Date.now(), this._fileIndex, "p", timeTaken);
+  }
+
+  setWriteToFileTimer() {
+    if (!this._timerActivated) {
+      this._timerActivated = true;
+      this._writeToFileTimer = setInterval(() => {
+        this._armTimer();
+      }, 2000);
+    }
+  }
+
+  _armTimer() {
+    if (this._tmpLogLines.length === 0) {
+      this._timerActivated = false;
+      clearInterval(this._writeToFileTimer);
+    } else {
+      this.writeToFile();
+    }
+  }
+
+  writeToFile() {
+    while (this._tmpLogLines.length > 0) {
+      let tracingLine = this._tmpLogLines.splice(0, 4);
+      if (typeof tracingLine[0] === "string") {
+        tracingLine = tracingLine.join(",");
+      } else {
+        tracingLine = "\n" + tracingLine.join(",");
+      }
+
+      this._writableStream.write(tracingLine);
+    }
+  }
+}
+
+module.exports = PerformanceWriter;
+
+// quick-lint-js finds bugs in JavaScript programs.
+// Copyright (C) 2020  Matthew Glazar
+//
+// This file is part of quick-lint-js.
+//
+// quick-lint-js is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// quick-lint-js is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with quick-lint-js.  If not, see <https://www.gnu.org/licenses/>.


plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+
+class PerformanceWriter {
+    
+    constructor (extensionContext) {
+        this._performanceTxtPath = path.join(process.cwd(), "test_performance.txt");
+        this._dataAttributes = {
+            columns:'timestamp,file-ref,file-name-or-duration,event-id'
+        };
+        this._fileName;
+        this._fileIndex;
```
strager, line 16: What are these doing? Looks like dead code. Did you mean to initialize these to `null`?
```suggestion
        this._fileName = null;
        this._fileIndex = null;
```

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+        this._extensionContext = extensionContext;
+    }
+
+    fileNameOptimizer (){
+        let currentSessionMap = this._extensionContext.globalState.get('qljsFileMap');
```
strager, line 21: Why isn't this an instance variable? What advantage is there to persisting this data in VS Code's globalState?

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+        this._fileName = this._fileName.split(path.sep);
+        this._fileName = this._fileName[this._fileName.length - 1];
+        if (!(currentSessionMap)) {
+            currentSessionMap = {};
+            currentSessionMap['latestIndex'] = 1;
+            currentSessionMap[this._fileName] = 1;
```
strager, line 27: Minor: What if the user opens a JavaScript file called `latestIndex`?

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+            this._fileIndex = 1;
+            this.writeToFile(null,"header");
+            this.writeToFile(null, "newFile");
+            this._extensionContext.globalState.update('qljsFileMap', currentSessionMap);
+        } else {
+            let existingFileIndex = currentSessionMap[this._fileName];
+            if (existingFileIndex) {
```
strager, line 34: Minor: This check will erroneously report that a file called `hasOwnProperty` or `toString`. Use a `Map` or do this instead:
```suggestion
            if (Object.prototype.hasOwnProperty.call(currentSessionMap, this._fileName)) {
                let existingFileIndex = currentSessionMap[this._fileName];
```

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+
+        if (mode === 'header') {
+            tracing_line = this._dataAttributes.columns;
+        } else if (mode === "newFile") {
+            tracing_line = '\n' + [Date.now(), this._fileIndex, this._fileName, 'n'].join(',');
+        } else {
+            tracing_line = '\n' + [Date.now(), this._fileIndex, timeTaken, ''].join(',');
+        }
+
+        fs.appendFile(this._performanceTxtPath, tracing_line, (err) => {
```
strager, line 57: Blocking: This is problematic for two reasons:
1. I assume that I/O is very expensive compared to linting. Here, we're opening a file, writing a tiny piece of data, then closing the file, every time the user types something. It shouldn't be blocking, so it shouldn't hang the editor or preventing linting from happening, but it's still wasteful of energy.
2. This is asynchronous. Are appends guaranteed to happen in order? Are appends guaranteed not to interleave? If not, we'll corrupt the file.

I suggest we do the following instead:
1. When we need to log data, append records to an array.
2. Periodically (e.g. every second), if there are any records in the array, append all the records to the file.

plugin/vscode/quick-lint-js.js
```diff
@@ -160,9 +162,14 @@ class Parser {
     }
   }
 
-  lint() {
+  lint(performanceWriter) {
+    let lint_start;
+    let lint_end;
```
strager, line 167: Nit: Variable names should be `lowerCamelCase`:
```suggestion
    let lintStart;
    let lintEnd;
```

plugin/vscode/quick-lint-js.js
```diff
@@ -160,9 +162,14 @@ class Parser {
...
+    lint_start = process.hrtime.bigint()
     let diagnosticsPointer = this._process._lint(this._parser);
-
+    lint_end = process.hrtime.bigint()
+    performanceWriter.fileNameOptimizer()
```
strager, line 171: Shouldn't we only need to do this when we open the document, not every time we lint?

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,82 @@
...
+// See end of file for extended copyright information.
+
+const fs = require('fs');
+const path = require("path");
+const process = require("process");
+
+class PerformanceWriter {
+    
+    constructor (extensionContext) {
+        this._performanceTxtPath = path.join(process.cwd(), "test_performance.txt");
```
strager, line 11: Blocking: We should pick a better file name. Also, cwd is probably the wrong place to put our files.

Does VS Code have a place for extensions to store log files or similar data? If so, we should use that place. If not, we should discuss a good home.

plugin/vscode/extension.js
```diff
@@ -130,11 +133,13 @@ class DocumentLinter {
     switch (this._state) {
       case DocumentLinterState.NO_PARSER:
         await this._createParser();
-        return await this.editorChangedVisibilityAsync();
+        await this.editorChangedVisibilityAsync();
+        break
```
strager, line 137: Nit: Why these changes? Make a separate PR if it's just code cleanup.

strager (changes requested):
Overall, this looks like the right direction in terms of design. However, see my comments marked "Blocking:".

Ideally, we'd have some tests for PerformanceWriter. (No need to test the changes to plugin/vscode/quick-lint-js.js though.)

Also, try to get Prettier working and run `yarn fmt`.

plugin/vscode/performance-writer.js
lifeinData: Addressed in [this commit](https://github.com/lifeinData/quick-lint-js/4f4bdd46c6cdbe22e9b84ed2d69f9ed4d372fe5a)

plugin/vscode/performance-writer.js
lifeinData: Addressed in [this commit](https://github.com/lifeinData/quick-lint-js/4f4bdd46c6cdbe22e9b84ed2d69f9ed4d372fe5a)

plugin/vscode/performance-writer.js
lifeinData: There is a logging schema from vscode for extension logging files. The schema is as follows: For each new process of vscode, (i.e. you open a new vscode process, this is NOT the same as opening a new window of vscode from an existing process) it adds a folder with a timestamp on to **%AppData%\Code\logs\_timestamped folder_** in windows and **~/.config/Code/logs/_timestamped folder_/** (Untested on linux, saw from [overstack post](https://stackoverflow.com/questions/46373762/where-are-visual-studio-code-log-files)) in linux. For each new workspace that the process has, it spawns a %AppData%\Code\logs\exthost# where # represents a given workspace. Inside of this workspace folder is where vscode wants you to store the extension log data. It automatically generates a URI for you via [extensionContext.logUri](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext).path. This folder does not automatically exist, you must create it. 

Implications: If we were to use vscode's extension schema, your initial example of keeping track of the files using qljs will still work but it may look abit weird in the sense that the initial index of the file and it's associated file name will be in disparate process folders. How would you like to proceed with the design of how you want to store the index and it's linking file name?

strager: > Implications: If we were to use vscode's extension schema, your initial example of keeping track of the files using qljs will still work but it may look abit weird in the sense that the initial index of the file and it's associated file name will be in disparate process folders.

This sounds fine to me. Not at all weird.

> How would you like to proceed with the design of how you want to store the index and it's linking file name?

What's the deletion policy for these folders? Does VS Code delete them automatically? Do we need to delete them ourselves?

plugin/vscode/performance-writer.js
lifeinData: The variable is now called *latestIndex because js file cannot be named with a *

plugin/vscode/quick-lint-js.js
lifeinData: When I asked you initially about logging, you mentioned that we wanted to log every time we lint (i.e. a keystroke is pressed). Could you expand more on what you mean by "when we open the document"?

plugin/vscode/performance-writer.js
lifeinData: resolved in latest commit

plugin/vscode/performance-writer.js
lifeinData: Put it as an instance variable in latest commit

plugin/vscode/performance-writer.js
lifeinData: followed the suggested change in latest commit

plugin/vscode/performance-writer.js
lifeinData: did it according to your suggested way in latest commit

plugin/vscode/quick-lint-js.js
lifeinData: Used your suggested Changes in latest commit

plugin/vscode/performance-writer.js
strager: > The variable is now called *latestIndex because js file cannot be named with a *

It can on Linux!

plugin/vscode/extension.js
```diff
@@ -364,11 +371,14 @@ exports.DocumentLinterCollection = DocumentLinterCollection;
...
 
-async function activateAsync() {
+async function activateAsync(extensionContext = vscode.ExtensionContext) {
   let diagnostics = vscode.languages.createDiagnosticCollection();
   toDispose.push(diagnostics);
-
-  let linters = new DocumentLinterCollection(diagnostics);
+  let performanceWriter = new PerformanceWriter(extensionContext)
+  performanceWriter.createHeader()
+  setInterval(() => {performanceWriter.writeToFile()}, 2000);
```
strager, line 379: Blocking: We don't want to be waking up VS Code while the user is idle. We should only arm timers when we have something to write to the performance log.

plugin/vscode/extension.js
```diff
@@ -426,6 +437,9 @@ async function deactivateAsync() {
   let disposing = toDispose.splice(0, toDispose.length);
   disposing.reverse();
   for (let disposable of disposing) {
+    if (disposable instanceof(DocumentLinterCollection)) {
+      disposable._performanceWriter.writeToFile()
+    }
```
strager, line 442: Why not have DocumentLinterCollection#dispose call `this._performanceWriter.writeToFile()`?

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+        this._dataAttributes = {
+            columns:'timestamp,file-ref,file-name-or-duration,event-id'
+        };
+        this._tmpLogLines = []
+        this._fileName = null;
+        this._fileIndex = null;
+        this._extensionContext = extensionContext;
+        fs.mkdirSync(this._extensionContext.logUri.fsPath)
+        this._performanceTxtPath = path.join(this._extensionContext.logUri.fsPath, "qljsPerformanceTrace.txt");
+        this._currentSessionMap = this._extensionContext.globalState.get('qljsFileMap');
```
strager, line 20: Why does qljsFileMap need to be persisted? Could we create an empty object here instead?

I worry that maintaining the qljsFileMap globalState might be costly. We have no mechanism to clean up old data from it, for example.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    createHeader (){
+        this.writeToTmpArray(null,"header");
+    }
+    
+    fileNameOptimizer (){
+        this._fileName = this._fileName.split(path.sep);
+        this._fileName = this._fileName[this._fileName.length - 1];
+        if (!(this._currentSessionMap)) {
```
strager, line 31: Performance: Move this check into the constructor.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+        this._writableStream = fs.createWriteStream(this._performanceTxtPath, {encoding: 'utf8', flags: 'a'})
+    }
+
+    createHeader (){
+        this.writeToTmpArray(null,"header");
+    }
+    
+    fileNameOptimizer (){
+        this._fileName = this._fileName.split(path.sep);
+        this._fileName = this._fileName[this._fileName.length - 1];
```
strager, line 30: Performance: Let's try to avoid string manipulation (especially allocation) in the hot path. I think we can store the full path instead of just the last component. Or, we can look up by `vscode.Document` instance (with a `Map`) (though this might get confusing with renames... unsure).

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+    
+    writeToTmpArray(timeTaken=null, mode=null) {
+        if (mode === 'header') {
+            this._tmpLogLines.push(this._dataAttributes.columns);
+        } else if (mode === "newFile") {
+            this._tmpLogLines.push('\n' + [Date.now(), this._fileIndex, this._fileName, 'n'].join(','));
+        } else {
+            this._tmpLogLines.push('\n' + [Date.now(), this._fileIndex, timeTaken, ''].join(','));
+        }
```
strager, line 60: Performance: Avoid array and string allocation on the hot path. How about something like this:

```suggestion
        if (mode === 'header') {
            this._tmpLogLines.push(...this._dataAttributes.columns);
        } else if (mode === "newFile") {
            this._tmpLogLines.push(Date.now(), this._fileIndex, this._fileName, 'n');
        } else {
            this._tmpLogLines.push(Date.now(), this._fileIndex, timeTaken, ''));
        }
```

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+                this._currentSessionMap[this._fileName] = this._currentSessionMap['*latestIndex'];
+                this._fileIndex = this._currentSessionMap['*latestIndex'];
+                this._extensionContext.globalState.update('qljsFileMap', this._currentSessionMap);
+                this.writeToTmpArray(null, "newFile", this._currentSessionMap['*latestIndex']);
+            }
+        }
+    }
+    
+    writeToTmpArray(timeTaken=null, mode=null) {
+        if (mode === 'header') {
```
strager, line 54: Performance: Make different methods for the different modes. As written, each call to `writeToTmpArray` involves at least one string comparison. All comparisons can be avoided by making separate methods.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    writeToFile() {
+        let tracingLine = null
+        if (this._tmpLogLines.length !== 0) {
+            for (tracingLine of this._tmpLogLines) {
+                this._writableStream.write(tracingLine)
+            }
+            this._tmpLogLines = []
```
strager, line 69: Performance: I *think* the following is better, but I'm not sure:
```suggestion
            this._tmpLogLines.length = 0;
```

plugin/vscode/extension.js
```diff
@@ -224,8 +230,8 @@ class DocumentLinter {
       this._pendingChanges.length = 0;
       this._state = DocumentLinterState.PARSER_LOADED;
       // END CRITICAL SECTION (no awaiting above)
-
-      let diags = this._parser.lint();
+      this._performanceWriter._fileName = this._document.fileName
+      let diags = this._parser.lint(this._performanceWriter);
```
strager, line 234: Minor: Why not pass the fileName as a parameter? This assignment looks weird, and it might accidentally get moved.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+    
+    constructor (extensionContext) {
+        this._dataAttributes = {
+            columns:'timestamp,file-ref,file-name-or-duration,event-id'
+        };
+        this._tmpLogLines = []
+        this._fileName = null;
+        this._fileIndex = null;
+        this._extensionContext = extensionContext;
+        fs.mkdirSync(this._extensionContext.logUri.fsPath)
```
strager, line 18: Performance: I'd rather this be asynchronous. Perhaps do this in `writeToFile`? (You'd need to lazily execute `fs.createWriteStream` in `writeToFile`.)

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+            return 'newFile'
+        } else {
+            if (Object.prototype.hasOwnProperty.call(this._currentSessionMap, this._fileName)) {
+                this._fileIndex = this._currentSessionMap[this._fileName];
+            } else {
+                this._currentSessionMap['*latestIndex'] = this._currentSessionMap['*latestIndex'] + 1;
+                this._currentSessionMap[this._fileName] = this._currentSessionMap['*latestIndex'];
+                this._fileIndex = this._currentSessionMap['*latestIndex'];
+                this._extensionContext.globalState.update('qljsFileMap', this._currentSessionMap);
+                this.writeToTmpArray(null, "newFile", this._currentSessionMap['*latestIndex']);
```
strager, line 48: Performance: Avoid repeated lookups of `this._currentSessionMap['*latestIndex']`.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+    
+    writeToTmpArray(timeTaken=null, mode=null) {
+        if (mode === 'header') {
+            this._tmpLogLines.push(this._dataAttributes.columns);
+        } else if (mode === "newFile") {
+            this._tmpLogLines.push('\n' + [Date.now(), this._fileIndex, this._fileName, 'n'].join(','));
```
strager, line 57: Almost-blocking: If the file name contains a comma or a newline, our log file gets a bit corrupt.

My original plan for handling commas was to put the file name as the *last* entry. Your schema adds a new event-type column after the file name. I'd move it before the file name column, or have some other mechanism for sanitizing commas.

My original plan for handling newlines was to replace newlines with spaces. (The data doesn't need to be 100% accurate, and who puts newlines in file paths anyway?!)

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,93 @@
+// Copyright (C) 2020  Matthew Glazar
+// See end of file for extended copyright information.
+
+const fs = require('fs');
+const path = require("path");
+const process = require("process");
+
+class PerformanceWriter {
```
strager, line 8: Please document the file format somewhere. Here (in performance-writer.js) is fine. In a separate .md file would be better.

strager (changes requested):

plugin/vscode/performance-writer.js
strager, line 30: This code means that two files named `foo.js` in different folders will get the same file ID/index. This seems wrong. We should key by the full path (or URI) or by `vscode.Document` instance or something more unique.

plugin/vscode/performance-writer.js
lifeinData: There is now an empty object specific to the instance of PerformanceWriter in latest commit

plugin/vscode/quick-lint-js.js
strager: > Could you expand more on what you mean by "when we open the document"?

Basically, the first lint for that file; the first time the quick-lint-js extension becomes aware of a file.

plugin/vscode/extension.js
lifeinData: Resolved in latest commit

plugin/vscode/extension.js
lifeinData: Used suggestion in latest commit

plugin/vscode/performance-writer.js
lifeinData: The PerformanceWriter class is only activated once during activateAsync and then writing performance of a new document starts with fileNameOptimizer method of this one instance of PerformanceWriter. If we were to move the _fileName into the constructor, it would imply that there is a new PerformanceWriter every time a new document is being performance traced which is completely doable, it's just not the way it's working now.

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]

This line is no longer relevant in latest commit

plugin/vscode/performance-writer.js
lifeinData: [Not addressed in latest commit]
Let's discuss this more.

plugin/vscode/performance-writer.js
lifeinData: See comment reply in [here](https://github.com/quick-lint/quick-lint-js/pull/244#discussion_r632243508).

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]

See refactored writeToFile method

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]

All if statements are standalone functions now.

plugin/vscode/performance-writer.js
strager: > If we were to move the _fileName into the constructor, it would imply that there is a new PerformanceWriter every time a new document is being performance traced which is completely doable, it's just not the way it's working now.

I wasn't thinking of doing this. I was thinking that the file path could be passed down as an argument instead of being a property on the PerformanceWriter object.

I think a single PerformanceWriter object, as currently implemented, is the right solution.

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]

the file index of the current vscode session is now called this._fileIndex and no longer a key within a object

plugin/vscode/performance-writer.js
lifeinData: the filepath is now an argument into the initiateLogging class method

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]
this._fileName is now an argument without separating the path to have only the filename

plugin/vscode/extension.js
lifeinData: [Addressed in latest commit]
this._fileName is now an argument without separating the path to have only the file name. The full path is logged now

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]
*latestIndex is no longer relevant

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]
this._fileName is now an argument without separating the path to have only the filename

plugin/vscode/performance-writer.js
lifeinData: [Addressed in latest commit]
event-id column and file-name-or-duration is switched now. Added a "p" event-id tag to a log line related to measuring the lint time.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+    this._currentSessionMap = {};
+  }
+
+  fileNameOptimizer(fileName) {
+    if (
+      Object.prototype.hasOwnProperty.call(this._currentSessionMap, fileName)
+    ) {
+      this._fileIndex = this._currentSessionMap[fileName];
+    } else {
+      this._fileIndex++;
```
strager, line 39: Blocking: This will cause file indexes to be reused. Example:
```js
pw.fileNameOptimizer("a.js");  // this._fileIndex = 1
pw.fileNameOptimizer("b.js");  // this._fileIndex = 2
pw.fileNameOptimizer("a.js");  // this._fileIndex = 1
pw.fileNameOptimizer("c.js");  // this._fileIndex = 2 // but b.js already claimed 2
```

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+      this._currentSessionMap[fileName] = this._fileIndex;
+      this.writeNewFileLine(fileName);
+    }
+  }
+
+  initiateLogging(timeTaken, fileName) {
```
strager, line 45: Nit: Call this `logLintDuration` or something more descriptive or succinct? `initiateLogging` sounds very Enterprise.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  writeExistingFileLine(timeTaken) {
+    this._tmpLogLines.push(Date.now(), this._fileIndex, "p", timeTaken);
+  }
+
+  setWriteToFileTimer() {
+    if (!this._timerActivated) {
+      this._timerActivated = true;
+      this._writeToFileTimer = setInterval(() => {
```
strager, line 66: Why setInterval? I think the logic would be simpler with setTimeout. Maybe not, though.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  writeToFile() {
+    while (this._tmpLogLines.length > 0) {
+      let tracingLine = this._tmpLogLines.splice(0, 4);
+      if (typeof tracingLine[0] === "string") {
+        tracingLine = tracingLine.join(",");
+      } else {
+        tracingLine = "\n" + tracingLine.join(",");
+      }
```
strager, line 88: Let's always add the newline prefix.

Let's say that, for whatever reason, we had a PerformanceWriter for this file which crashed in the middle of writing. The file might look like this (without a trailing newline):
```
timestamp,file-ref,event-id,file-name-or-duration
1234,1,
```
If we start up a PerformanceWriter, then we'll write the header again, but without a leading newline. The file will end up looking like this:
```
timestamp,file-ref,event-id,file-name-or-duration
1234,1,timestamp,file-ref,event-id,file-name-or-duration
4567,1,n,banana.js
4567,1,p,12473
```
The file is corrupted.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+    this._dataAttributes = {
+      columns: ["timestamp", "file-ref", "event-id", "file-name-or-duration"],
+    };
+    this._tmpLogLines = [];
+    this._fileIndex = 0;
+    this._writeToFileTimer = null;
+    this._timerActivated = false;
+    if (!fs.existsSync(this._extensionContext.logUri.fsPath)) {
+      fs.mkdirSync(this._extensionContext.logUri.fsPath);
+      this.writeHeaderLine();
```
strager, line 20: Almost-blocking: Avoid `Sync` APIs. Use the non-Sync or Promise variants.

strager, line 20: I think we should write the header line unconditionally.

plugin/vscode/performance-writer.js
strager: I realize now that you only write the header line if the file didn't already exist. So I guess the situation I mentioned won't cause corruption.

Also, you mentioned in chat that the file wouldn't be reused in practice anyway.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  initiateLogging(timeTaken, fileName) {
+    this.fileNameOptimizer(fileName);
+    this.writeExistingFileLine(timeTaken);
+    this.setWriteToFileTimer();
+  }
+
+  writeHeaderLine() {
+    this._tmpLogLines.push(...this._dataAttributes.columns);
```
strager, line 52: Nit: Let's inline the this._dataAttributes.columns variable. 
```suggestion
    this._tmpLogLines.push("timestamp", "file-ref", "event-id", "file-name-or-duration");
```

plugin/vscode/extension.js
```diff
@@ -358,18 +369,19 @@ class DocumentLinterCollection {
...
+      this._performanceWriter.writeToFile();
     });
   }
 }
 exports.DocumentLinterCollection = DocumentLinterCollection;
 
 let toDispose = [];
 
-async function activateAsync() {
+async function activateAsync(ExtensionContext) {
```
strager, line 380: I think tests need to provide an ExtensionContext in order for this code to work now. Should be easy to whip up a temporary directory; I think tests make a temp dir already for the workspace.

plugin/vscode/extension.js
```diff
@@ -226,8 +234,7 @@ class DocumentLinter {
       this._pendingChanges.length = 0;
       this._state = DocumentLinterState.PARSER_LOADED;
       // END CRITICAL SECTION (no awaiting above)
-
-      let diags = this._parser.lint();
+      let diags = this._parser.lint(this._performanceWriter);
```
strager, line 237: Blocking: This looks sketch.

website/test/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
+# Copyright (C) 2020  Matthew Glazar
```
strager, line 1: Blocking: I don't think you meant to add these files in website/. Remove them.

strager (changes requested):

plugin/vscode/extension.js
lifeinData: Fixed in latest commit:
this._performanceWriter is deleted

website/test/CMakeLists.txt
lifeinData: Fixed in latest commit:
rebased to latest qljs master branch

plugin/vscode/performance-writer.js
lifeinData: Fixed in latest commit:
Did it as you suggested

plugin/vscode/performance-writer.js
lifeinData: Current logic is as follows for logging linting performance:
setWriteToFileTimer is activated every single time a key stroke is pressed => setInterval is activated => keeps going until the tmpLogLines is empty AND also it is locked from being activated again via the "!this._timerActivated" condition. 

If using setTimeout , the logic, in my opinion would be exactly the same because you'll need to activate setTimeout again if there is new tmpLogLines while waiting for the setTimeout to activate. I'll need to build out some sort of logic to know to setTimeout again if there is new tmpLogLines elements while waiting for setTimeout to activate which is essentially setInterval.

On the other hand, you could just activate setTimeout everytime a keystroke is pressed and you'll have _many_ setTimeout functions going on... but I'm sure that's not what you meant with this implementation. 

Conclusion: we should probably just keep setInterval as setTimeout would boil down to the same logic imo.

strager: > If using setTimeout , the logic, in my opinion would be exactly the same because you'll need to activate setTimeout again if there is new tmpLogLines while waiting for the setTimeout to activate.

No, you wouldn't need to activate setTimeout again if there is new tmpLogLines. You can write those tmpLogLines immediately.

lifeinData: Fixed in latest commit:
Did it as you suggested

plugin/vscode/performance-writer.js
lifeinData: Fixed in latest commit:
Did it as you suggested

plugin/vscode/performance-writer.js
lifeinData: Fixed in latest commit

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+      this._tmpLogLines = [];
+      this._currentIndex = null;
+      this._fileIndex = 0;
+      this._writeToFileTimer = null;
+      this._timerActivated = false;
+      fs.mkdir(this._extensionContext.logUri.fsPath, (err) => {
+        if (!err) {
+          this.writeHeaderLine();
+        }
+      });
```
strager, line 21: Must fix: We aren't waiting for `mkdir` to finish before calling `createWriteStream`. I think this code is fragile and might not even work.

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  writeExistingFileLine(timeTaken) {
+    this._tmpLogLines.push(Date.now(), this._currentIndex, "p", timeTaken);
+  }
+
+  setWriteToFileTimer() {
+    if (!this._timerActivated) {
+      setTimeout(() => {
+        this._timerActivated = true;
```
strager, line 74: Shouldn't this go outside the `setTimeout` callback? I don't see the point of the `_timerActivated` property as written. (`writeToFile` will set `_timerActivated` to `false`.)

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+        this.writeToFile();
+      }, 1000);
+    }
+  }
+
+  writeToFile() {
+    while (this._tmpLogLines.length > 0) {
+      let tracingLine = this._tmpLogLines.splice(0, 4);
+      if (typeof tracingLine[0] === "string") {
```
strager, line 83: I think a variable or comment would significantly help with readability:
```suggestion
      let isHeaderLine = typeof tracingLine[0] === "string";
      if (isHeaderLine) {
```

plugin/vscode/performance-writer.js
```diff
@@ -0,0 +1,113 @@
...
+const process = require("process");
+
+class PerformanceWriter {
+  constructor(extensionContext) {
+    if (extensionContext !== undefined) {
+      this._extensionContext = extensionContext;
+      this._tmpLogLines = [];
+      this._currentIndex = null;
+      this._fileIndex = 0;
+      this._writeToFileTimer = null;
```
strager, line 15: Nit: Delete unused property.
```suggestion
```

strager (changes requested):

Issue #234, opened 2021-04-18
7$: E024 mentions 'let' even when using 'var' or 'const'


strager:
```js
const
if (true) {}
let
if (true) {}
var
if (true) {}
```
Observed:
```
hello.js:1:1: error: let with no bindings [E024]
hello.js:5:1: error: let with no bindings [E024]
hello.js:3:1: warning: use of undeclared variable: let [E057]
```
Expected:
```
hello.js:1:1: error: const with no bindings [E024]
hello.js:5:1: error: var with no bindings [E024]
hello.js:3:1: warning: use of undeclared variable: let [E057]
```

LeeWannacott:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.
  
``` cpp
QLJS_ERROR_TYPE(                                                             \
      error_let_with_no_bindings, "E024", { source_code_span where; },         \
-      .error(QLJS_TRANSLATABLE("let with no bindings"), where))               \
+     .error(QLJS_TRANSLATABLE("{0} with no bindings"), where))```           \

I changed this one line and it does the `Expected` output. 
I don't expect payment if it was as simple as changing one line.

LeeWannacott:
I have read and hereby agree to quick-lint-js' Contributor License Agreement (CLA-v1.md).

PR #246, opened 2021-04-27
7$: E024 mentions 'let' even when using 'var' or 'const' #234

LeeWannacott:
Added `{0}` instead of hard-coded `let` therefore const or var with no bindings is reported. 
Not worried about payment `if` this solves it; because it was pretty easy.


LeeWannacott:

strager (approved):
Wow, this is beautiful! Much easier than I thought it would be.

Some tests in plugin/vscode/ are still looking for the "let with no bindings" message. You should update those tests.

strager:
Landed as commit 9d4bc2391f67bfc94d473f1df274b7c5b8077b2f.

Issue #233, opened 2021-04-18
10$: Improve diagnostics for var\nvar\nvar


strager:
quick-lint-js emits confusing diagnostics for the following code:

```js
var
var
var
```
```
hello.js:1:1: error: let with no bindings [E024]
hello.js:3:1: error: cannot declare variable named keyword 'var' [E124]
```

quick-lint-js should report E024 for each line.

LeeWannacott:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

LeeWannacott:
Question :thinking: : Does each line mean? 
```
hello.js:1:1: error: var with no bindings [E024]
hello.js:2:1: error: var with no bindings [E024]
hello.js:3:1: error: var with no bindings [E024]
```
Or:
```
hello.js:1:1: error: var with no bindings [E024]
hello.js:3:1: error: var with no bindings [E024]
```

strager:
> Question 🤔 : Does `each line` mean?
> 
> ```
> hello.js:1:1: error: var with no bindings [E024]
> hello.js:2:1: error: var with no bindings [E024]
> hello.js:3:1: error: var with no bindings [E024]
> ```
> 
> Or:
> 
> ```
> hello.js:1:1: error: var with no bindings [E024]
> hello.js:3:1: error: var with no bindings [E024]
> ```
> 
> Note: After the arduous issue of #234 it now outputs error: var, not let.

The former: 3 errors, one per `var`.

strager:
I think improving this isn't worth the engineering effort.

PR #247, opened 2021-04-28
10$: Improve diagnostics for var\nvar\nvar #233

LeeWannacott:
Removing case `token_type::end_of_file:` from error report E124: `error_cannot_declare_variable_with_keyword_name` gives the expected output. 
```
test.js:1:1: error: var with no bindings [E024]
test.js:2:1: error: var with no bindings [E024]
test.js:3:1: error: var with no bindings [E024]
```
Unless there is cases where you want that error at the end of file?

Note: I have my commits from issue 234 because it's a related issue and it's not merged yet.

### DIFF
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -3037,7 +3037,6 @@ class parser {
 
         switch (this->peek().type) {
         // let switch = 3;  // Invalid.
-        case token_type::end_of_file:
         case token_type::equal:
         case token_type::semicolon:
           this->lexer_.commit_transaction(std::move(transaction));
--- a/plugin/vscode/test/other-tests.js
+++ b/plugin/vscode/test/other-tests.js
@@ -426,6 +426,8 @@ let tests = {
           assert.strictEqual(messages.length, 1, messages);
           assert.ok(
             messages[0] === "let with no bindings" ||
+              messages[0] === "const with no bindings" ||
+              messages[0] === "var with no bindings" ||
               messages[0] === "redeclaration of variable: x",
             messages
           );
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -503,7 +503,7 @@
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
       error_let_with_no_bindings, "E024", { source_code_span where; },         \
-      .error(QLJS_TRANSLATABLE("let with no bindings"), where))                \
+      .error(QLJS_TRANSLATABLE("{0} with no bindings"), where))                \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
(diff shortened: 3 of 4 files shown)


LeeWannacott:
Removing case token_type::end_of_file: from error report E124: error_cannot_declare_variable_with_keyword_name causes this test to fail. I guess it expects a keyword after var but its getting var\nvar\nvar.
``` cpp
for (string8 keyword : disallowed_binding_identifier_keywords) {
    {
      padded_string code(u8"var " + keyword);
      SCOPED_TRACE(code);
      spy_visitor v;
      parser p(&code, &v);
      EXPECT_TRUE(p.parse_and_visit_statement(v));
      EXPECT_THAT(v.variable_declarations, IsEmpty());
      EXPECT_THAT(v.errors,
                  ElementsAre(ERROR_TYPE_FIELD(
                      error_cannot_declare_variable_with_keyword_name, keyword,
                      offsets_matcher(&code, strlen(u8"var "), keyword))));
    }
}
```
I guess it expects a keyword after var but its getting var\nvar\nvar.

Note: The other tests inside the for loop don't fail.

src/quick-lint-js/parse.h
```diff
@@ -3037,7 +3037,6 @@ class parser {
```
strager, line 3037: I think a proper fix would involve checking `this->lexer_.has_leading_newline` on line 3034. I could be wrong though.

src/quick-lint-js/parse.h
```diff
@@ -3037,7 +3037,6 @@ class parser {
 
         switch (this->peek().type) {
         // let switch = 3;  // Invalid.
-        case token_type::end_of_file:
```
strager, line 3040: Blocking: This deletion changes the behavior of `let var` or `var var` at the end of a file, as the failing test points out. We should make `var var` (at end of file) continue to report E124.

strager:
> Removing case `token_type::end_of_file:` from error report E124: `error_cannot_declare_variable_with_keyword_name` gives the expected output.
> 
> ```
> test.js:1:1: error: var with no bindings [E024]
> test.js:2:1: error: var with no bindings [E024]
> test.js:3:1: error: var with no bindings [E024]
> ```

Write a test for this.

> Removing case token_type::end_of_file: from error report E124: error_cannot_declare_variable_with_keyword_name causes the following test to fail in the file test-parse-var.cpp:
>
> [snip]
>
> I guess it expects a keyword after var but its getting var\nvar\nvar.

The test is making the following code:
```cpp
var var
```
(Note the lack of a line break.)

strager (changes requested):

strager:
@LeeWannacott Do you plan on implementing this?

LeeWannacott:
I did try. Feel free to give the issue to someone else; if they want it /close the PR. I might have another punt at some point. I think my time is better spent learning JavaScript and React though.

PR #249, opened 2021-04-29
Update error.h

hanycopes:
Showing a better message with a const field error.

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -1059,7 +1059,10 @@
      and helpful error instead. */                                             \
   QLJS_ERROR_TYPE(                                                             \
       error_unexpected_token, "E054", { source_code_span token; },             \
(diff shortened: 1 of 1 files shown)


src/quick-lint-js/error.h
```diff
@@ -1059,7 +1059,10 @@
      and helpful error instead. */                                             \
   QLJS_ERROR_TYPE(                                                             \
       error_unexpected_token, "E054", { source_code_span token; },             \
-      .error(QLJS_TRANSLATABLE("unexpected token"), token))                    \
+      .error(QLJS_TRANSLATABLE(                                                \
+          "const fields are only allowed in TypeScript,"                       \
+          "not JavaScript [E054]"),                                            \
+          token))                                                              \
```
strager, line 1065: Blocking: This change affects all "unexpected token" errors. We should only show this diagnostic message for the specific case of using const fields.

strager (changes requested):

PR #259, opened 2021-05-01
Report warning for redundant delete statement on variable

### LINKED ISSUE #251: 7$: Warn when deleting variable
In the following code, `delete x;` seems to have no effect:
```js
let x = 3;
delete x;
console.log(x);
```
Have quick-lint-js report a warning saying that `delete` has no effect.

singalhimanshu:
Fixes #251

### DIFF
--- a/docs/errors/E086.md
+++ b/docs/errors/E086.md
@@ -1,6 +1,6 @@
-# E086: redundant delete statement
+# E086: redundant delete statement on variable
 
-The following delete statement is redundant:
+The following delete statement is redundant on variable:
 
     let x = 3;
     delete x;
--- a/docs/errors/E086.md
+++ b/docs/errors/E086.md
@@ -1,0 +1,12 @@
+# E086: redundant delete statement
+
+The following delete statement is redundant:
+
+    let x = 3;
+    delete x;
+    console.log(x);
+
+To fix this warning, remove the delete statement:
+
+    let x = 3;
+    console.log(x);
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -748,6 +748,10 @@
       .error(QLJS_TRANSLATABLE("missing operand for operator"), where))        \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_redundant_delete_statement, "E086", { source_code_span where; },   \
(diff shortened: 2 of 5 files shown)


docs/errors/E086.md
```diff
@@ -0,0 +1,12 @@
+# E086: redundant delete statement
```
strager, line 1: The error should explain *why* the `delete` is redundant.

* Is there two consecutive `delete` statements?
* Does the `delete` statement have no effect because the object is frozen?
* Does the `delete` statement have no effect because the operand is a variable?
* Does the `delete` statement have no effect because the operand is otherwise not a property lookup?

test/test-parse-expression-statement.cpp
```diff
@@ -754,6 +747,19 @@ TEST(test_parse, expression_statement) {
...
+    spy_visitor v;
+    padded_string code(u8"delete x;"_sv);
+    parser p(&code, &v);
+    p.parse_and_visit_expression(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use"));
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"x"}));
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_redundant_delete_statement_on_variable,
+                              where, offsets_matcher(&code, 0, u8"delete"))));
```
strager, line 760: Should we highlight `delete x` instead of just `delete`?

src/quick-lint-js/error.h
```diff
@@ -747,6 +747,12 @@
       error_missing_operand_for_operator, "E026", { source_code_span where; }, \
       .error(QLJS_TRANSLATABLE("missing operand for operator"), where))        \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_redundant_delete_statement_on_variable, "E086",                    \
+      { source_code_span where; },                                             \
```
strager, line 752: `where` is a bad variable name. (I know a lot of other error types use it.) Let's pick something better. `delete_operator`? `delete_expression` (if we change it to cover the entire expression, not just the keyword)?

strager (approved):

Issue #188, opened 2021-03-31
7$: zsh/bash tab completion for CLI


strager:
- [x] macOS: Homebrew
  - [ ] ~~bash~~ -- **broken** (not sourced; crashes)
  - [x] fish
  - [x] zsh (Homebrew)
  - [x] zsh (Apple)
- [ ] macOS: manual install
  - [ ] zsh (Apple) -- **broken** (not sourced)
- [ ] Debian: manual install
  - [ ] bash
  - [ ] fish
  - [ ] zsh
- [x] Debian: .deb install
  - [x] bash
  - [x] fish
  - [x] zsh
- [ ] Debian: Nix install
  - [x] bash
  - [x] fish
  - [ ] zsh
- [x] Arch: manual install
  - [x] bash
  - [x] fish
  - [x] zsh

maniac-en:
I claim this for-hire task. I do not expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
We need to check that completion works when installing in the following ways:

- [ ] Homebrew
- [ ] Nix
- [ ] Debian/apt

Once everything is confirmed working, I think this task is bon.

strager:
Homebrew on macOS: zsh, bash, and fish completion all **do not work** (pressing tab after `quick-lint-js -` completes nothing and shows no menu).

The completion files do seem to be installed:
* /usr/local/share/bash-completion/completions/quick-lint-js.bash
* /usr/local/share/fish/vendor_completions.d/quick-lint-js.fish
* /usr/local/share/zsh/site-functions/_quick-lint-js

I'm not using Homebrew-installed shells. Maybe I need to install the shells with Homebrew.

strager:
If I use Homebrew-installed shells, things work a bit better:
* zsh completion works
* bash completion **does not work**
* fish completion works

strager:
When I uninstalled the Homebrew-installed bash, I saw this message:

```
Uninstalling /usr/local/Cellar/bash/5.1.8... (157 files, 9.7MB)

Warning: The following may be bash configuration files and have not been removed!
If desired, remove them manually with `rm -rf`:
  /usr/local/etc/bash_completion.d
```

/usr/local/etc/bash_completion.d is not where quick-lint-js installed completion files.

I copied quick-lint-js.bash into /usr/local/etc/bash_completion.d/, but completion in bash (Homebrew or macOS-provided) still didn't work.

[...]

PR #260, opened 2021-05-01
Added command line completion for CLI

maniac-en:
I've also added docs/COMPLETIONS.md which explains how to add shell
completion for bash, zsh and fish shell manually!

Note: This commit doesn't affect the working of quick-lint-js.

Fixes issue: #188


maniac-en:

maniac-en:
recheck

docs/COMPLETIONS.md
```diff
@@ -0,0 +1,49 @@
+### Shell completions
```
strager, line 1: Nit: I'd move this file into the `completions` directory and call it `README.md`.

strager (changes requested):
Blocking: At least the zsh completion seems to be broken.

I typed `quick-lint-js ` then pressed tab. I was presented with a menu of *options*, but I expected to see a menu of *files*.

What I see after pressing tab (with my zsh settings):
```
$ quick-lint-js --exit-fail-on
--exit-fail-on           -- Fail with a non-zero exit code if any of these errors are f
--help            -h     -- Print help message
--lsp-server      --lsp  -- Run quick-lint-js in LSP server mode
--output-format          -- Format to print feedback where FORMAT is one of: gnu-like (
--version         -v     -- Print version information
--vim-file-bufnr         -- Select a vim buffer for outputting feedback
```

What I expect after pressing tab:
```
$ quick-lint-js CMakeLists.txt
CMakeLists.txt  build-debug/    copyrite        hello.js        test/
COPYING         build-fuzz/     default.nix     hello.mjs       tools/
Formula/        build-profile/  dist/           make*           vendor/
README.md       cmake/          docs/           plugin/         website/
benchmark/      completions/    fuzz/           po/
build/          configure*      fuzz-corpus/    src/
```

Blocking: `--output-format` doesn't seem to complete in zsh:

1. If I type `quick-lint-js --out` then press tab, zsh fills in `quick-lint-js --output-format `. This is fine.
2. If I then press tab again, zsh rings a bell and shows no completion options. This is wrong. I expect zsh to present me with a menu of choices.

(Also, ideally, after pressing tab in step #1, zsh would fill in `quick-lint-js --output-format=` and present a menu of choices. Maybe this isn't possible or typical, so I'm fine with the user needing to press tab again.)

completions/_quick-lint-js
```diff
@@ -0,0 +1,22 @@
+#compdef _quick-lint-js quick-lint-js
```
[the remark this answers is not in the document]
maniac-en: This is not a comment, though I don't see the difference in working of script even if we redact `#`.

I've two reasons to use it as it is:
1. [zsh docs](http://zsh.sourceforge.net/Doc/Release/Completion-System.html)
> If the number of completion files changes, compinit will recognise this and produce a new dump file. However, if the name of a function or the arguments in the first line of a #compdef function (as described below) change, it is easiest to delete the dump file by hand so that the next time compinit will re-create it.

2. [zsh-users HowTo](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command)
> The first line of a completion function file can look something like this:
> 
> #compdef foobar
> 
> This tells zsh that the file contains code for completing the foobar command. This is the format that you will use most often for the first line, but you can also use the same file for completing several different functions [..SNIP..]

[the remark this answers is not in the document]
maniac-en: It can be named `quick-lint-js.zsh` but it's suggested to use `_quick-lint-js` according to  [zsh-users HowTo](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command). I'm ok with changing it if you are!

> Completion functions for commands are stored in files with names beginning with an underscore _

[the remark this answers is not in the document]
maniac-en: I've added the copyright comments in all completion scripts.

Please update `tools/check-file-copyrights`

completions/_quick-lint-js
```diff
@@ -0,0 +1,22 @@
...
+typeset -A opt_args
+
+function _quick-lint-js() {
+	integer ret=1
```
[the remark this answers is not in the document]
maniac-en: `Changes implemented!`

completions/quick-lint-js.bash
```diff
@@ -0,0 +1,19 @@
...
+
+	if [[ $cur == -* ]]; then
+		COMPREPLY=($(compgen -W '${opts}' -- "$cur"))
+		[[ ${COMPREPLY-} == *= ]] && compopt -o nospace
+		return 0
+	fi
+} &&
+	complete -F _quick-lint-js quick-lint-js
+
+# vim: ft=bash:
```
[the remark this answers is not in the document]
maniac-en: No it's not necessary for this file, the file extension is clear enough already. I added it earlier as it was necessary in other two files, but I've changed it now.

`Changes implemented!`

docs/COMPLETIONS.md
```diff
@@ -0,0 +1,49 @@
...
+
+To get automatic completions for quick-lint-js's options you can install the provided shell completions.
```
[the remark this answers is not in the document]
maniac-en: `TODO`

maniac-en:
Fixed
> I typed quick-lint-js then pressed tab. I was presented with a menu of options, but I expected to see a menu of files.

Also implemented
> (Also, ideally, after pressing tab in step `#1`, zsh would fill in quick-lint-js --output-format= and present a menu of choices. Maybe this isn't possible or typical, so I'm fine with the user needing to press tab again.)

completions/_quick-lint-js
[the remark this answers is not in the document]
strager: Unless you want to maintain copyright for your contribution, the copyright comments should say Matthew Glazar, not Shivam Mehta. They should also say 2020 and have two spaces between the year and my name, exactly like other files.

strager (commented):
> Also implemented
> 
> > (Also, ideally, after pressing tab in step `#1`, zsh would fill in quick-lint-js --output-format= and present a menu of choices. Maybe this isn't possible or typical, so I'm fine with the user needing to press tab again.)

This doesn't seem to be implemented, at least on zsh version 5.8. Tab completes `--output-format=` with no menu of choices. Pressing tab again has no effect (i.e. no menu).

completions/_quick-lint-js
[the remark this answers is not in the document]
maniac-en: I don't plan to maintain copyright for my contribution, so I'll change it as specified!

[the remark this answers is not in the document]
maniac-en: Filename kept as `_quick-lint-js` and the file header has now following comments:
> This script is used to provide command line completion for
> quick-lint-js when using zsh shell

maniac-en:
> This doesn't seem to be implemented, at least on zsh version 5.8. Tab completes `--output-format=` with no menu of choices. Pressing tab again has no effect (i.e. no menu).

Latest commit should fix this.

Along with that I've also added functionality to ignore incompatible parameters:
>Example:
>	`--lsp or --lsp-server` won't be suggested if `--exit-fail-on`
>	or `--output-format` is present

completions/.gitignore
```diff
@@ -0,0 +1,5 @@
+CMakeCache.txt
```
strager, line 1: Delete this .gitignore file. (See my blocking comment in completions/README.md.)

completions/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+# Copyright (C) 2020  Matthew Glazar
+# See end of file for extended copyright information.
+
+cmake_minimum_required(VERSION 3.10)
+project(quick-lint-js-completions)
```
strager, line 5: Why make a `project`? Let's inherit the `project(quick-lint-js)` from `/CMakeLists.txt`.

completions/README.md
```diff
@@ -1,12 +1,18 @@
...
+# quick-lint-js completions
 
-To get automatic completions for quick-lint-js's options you can install the provided shell completions.
+## Using cmake
 
-## Zsh
+```sh
+cmake .
+make
+sudo make install
```
strager, line 8: Blocking: We shouldn't provide instructions on *only* building and installing completion files. We should install completion files as part of the CLI build. This is how the manual (`quick-lint-js.1`) is installed, for example.

(The manual installation instructions below are good to have.)

completions/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+install(
+  FILES _quick-lint-js
+  DESTINATION "/usr/share/zsh/site-functions"
```
strager, line 9: Blocking: We should use `${DATAROOTDIR}` from `GNUInstallDirs` instead of hard-coding `/usr/share`. https://cmake.org/cmake/help/v3.10/module/GNUInstallDirs.html

See docs/CMakeLists.txt for a similar example.

strager (changes requested):

completions/_quick-lint-js
```diff
@@ -0,0 +1,58 @@
...
+
+        args+=(
+        "(- *)"{-h,--help}"[Print help message]"
+        "(- *)"{-v,--version}"[Print version information]"
+        "($I1 $I2 --lsp --lsp-server)"{--lsp,--lsp-server}"[Run quick-lint-js in LSP server mode]"
+        "($I1 $I3)--exit-fail-on=[Fail with a non-zero exit code if any of these errors are found (default\: \"all\")]"
+        "($I1 $I3)--output-format=[Format to print feedback where FORMAT is one of\: gnu-like (default if omitted), vim-qflist-json]:format:->format"
+        "($I1)--vim-file-bufnr=[Select a vim buffer for outputting feedback]")
+
+        _arguments -C -S -s "${args[@]}" '*:file:_files' && ret=0
```
strager, line 29: quick-lint-js' command line option parser does not support option stacking. Therefore, `-s` shouldn't be given to `_arguments`.

completions/_quick-lint-js
```diff
@@ -0,0 +1,58 @@
...
+        local I2="--output-format"
+        local I3="--lsp --lsp-server"
+
+        args+=(
+        "(- *)"{-h,--help}"[Print help message]"
+        "(- *)"{-v,--version}"[Print version information]"
+        "($I1 $I2 --lsp --lsp-server)"{--lsp,--lsp-server}"[Run quick-lint-js in LSP server mode]"
+        "($I1 $I3)--exit-fail-on=[Fail with a non-zero exit code if any of these errors are found (default\: \"all\")]"
+        "($I1 $I3)--output-format=[Format to print feedback where FORMAT is one of\: gnu-like (default if omitted), vim-qflist-json]:format:->format"
+        "($I1)--vim-file-bufnr=[Select a vim buffer for outputting feedback]")
```
strager, line 27: The outcome of the exclusions seems a bit weird:

* `quick-lint-js --lsp --<TAB>` will suggest `--exit-fail-on` and `--vim-file-bufnr`. Neither of these options work with `--lsp`.
* `quick-lint-js --output-format=gnu-like --<TAB>` will suggest  `--exit-fail-on` and `--vim-file-bufnr`. `--help` and `--version` also make sense to suggest (in my opinion).

I think the following changes should make it complete how I expect:
```suggestion
        "(--exit-fail-on --lsp --lsp-server --vim-file-bufnr)"{--lsp,--lsp-server}"[Run quick-lint-js in LSP server mode]"
        "(--lsp --lsp-server)--exit-fail-on=[Fail with a non-zero exit code if any of these errors are found (default\: \"all\")]"
        "(--lsp --lsp-server --output-format)--output-format=[Format to print feedback where FORMAT is one of\: gnu-like (default if omitted), vim-qflist-json]:format:->format"
        "(--lsp --lsp-server)--vim-file-bufnr=[Select a vim buffer for outputting feedback]")
```

maniac-en: `Implemented`

completions/CMakeLists.txt
maniac-en: `Implemented`

completions/.gitignore
maniac-en: `File deleted`

docs/COMPLETIONS.md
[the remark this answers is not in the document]
maniac-en: `CMakeLists` implemented!

completions/_quick-lint-js
```diff
@@ -0,0 +1,53 @@
...
+        local -a args
+
+        args+=(
+        "(- *)"{-h,--help}"[Print help message]"
+        "(- *)"{-v,--version}"[Print version information]"
+	"(--exit-fail-on --lsp --lsp-server --vim-file-bufnr)"{--lsp,--lsp-server}"[Run quick-lint-js in LSP server mode]"
+        "(--lsp --lsp-server)--exit-fail-on=[Fail with a non-zero exit code if any of these errors are found (default\: \"all\")]"
+        "(--lsp --lsp-server --output-format)--output-format=[Format to print feedback where FORMAT is one of\: gnu-like (default if omitted), vim-qflist-json]:format:->format"
+        "(--lsp --lsp-server)--vim-file-bufnr=[Select a vim buffer for outputting feedback]")
+
```
strager, line 23: If `--lsp-server` is on the command line, files shouldn't be completed.

Similarly, if a file is on the command line, `--lsp-server` shouldn't be completed.

completions/_quick-lint-js
```diff
@@ -0,0 +1,53 @@
...
+function _quick-lint-js() {
+        local curcontext="$curcontext" state line
+        local -i ret=1
+        local -a args
+
+        args+=(
+        "(- *)"{-h,--help}"[Print help message]"
+        "(- *)"{-v,--version}"[Print version information]"
+	"(--exit-fail-on --lsp --lsp-server --vim-file-bufnr)"{--lsp,--lsp-server}"[Run quick-lint-js in LSP server mode]"
```
strager, line 19: If `--lsp-server` is on the command line, `--output-format` should not be completed.

completions/quick-lint-js.bash
```diff
@@ -0,0 +1,39 @@
...
+# Copyright (C) 2020  Matthew Glazar
+# See end of file for extended copyright information.
+
+_quick-lint-js () {
+        local cur prev opts
+
+        COMPREPLY=()
+        cur="${COMP_WORDS[COMP_CWORD]}"
+        prev="${COMP_WORDS[COMP_CWORD-1]}"
+        opts="-h --help -v --version --lsp --lsp-server --exit-fail-on --output-format --vim-file-bufnr"
```
strager, line 12: Instead of suggesting both `-h` and `--help`, let's suggest only `--help`. I don't want to clutter the completion menu.

Similarly with `-v` and `--version`, and `--lsp` and `--lsp-server`.

strager, line 12: Is it possible to include a `=` after `--exit-fail-on` and `--vim-file-bufnr` and `--output-format`?

completions/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+)
+
+install(
+  FILES quick-lint-js.bash
+  DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/bash-completion/completions"
+)
+
+install(
+  FILES quick-lint-js.fish
+  DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/fish/completions"
```
strager, line 19: This looks like the wrong installation location for fish completions, according to my reading of this document:

> The completions shipped with fish, usually installed in /usr/share/fish/completions
> https://fishshell.com/docs/current/completions.html

I think fish/vendor_completions.d/ is a better installation location:

> A directory for third-party software vendors to ship their own completions for their software, usually /usr/share/fish/vendor_completions.d

strager: fish completions were autogenerated from the `man` page on my machine. This is because my fish installation didn't look in /usr/share/fish/completions; it instead looked in /usr/share/fish/vendor_completions.d.

strager (changes requested):
Blocking: CMakeLists.txt needs `add_subdirectory(completions)`. Otherwise, completions/CMakeLists.txt is unused.

completions/quick-lint-js.fish
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
maniac-en: I agree, shebang's not needed! Next commit removes shebangs in all completions!

completions/quick-lint-js.fish
```diff
@@ -0,0 +1,31 @@
...
+#
+# quick-lint-js is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with quick-lint-js.  If not, see <https://www.gnu.org/licenses/>.
+
+# vim: ft=sh:
```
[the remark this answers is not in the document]
maniac-en: Keeping the modeline, shebang removed!

completions/CMakeLists.txt
maniac-en: Changed in next commit

completions/_quick-lint-js
```diff
@@ -0,0 +1,53 @@
...
+#
+# quick-lint-js is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with quick-lint-js.  If not, see <https://www.gnu.org/licenses/>.
+
+# vim: ft=zsh:
```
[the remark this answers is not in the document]
maniac-en: Removed

completions/_quick-lint-js
maniac-en: implemented in next commit!

maniac-en:
As per my findings from fish docs and other fish completions, it behaves differently when it comes to appending `=` to an option unlike bash and zsh. Adding a submenu for `--output-format` in case of fish gives a completion like this (notice the space after `=`)



It looks ugly in my opinion, and I'm not sure even functional/valid. On the other hand, it works properly for bash and zsh.
It's your call now if we should keep it or not. :)

strager:
> As per my findings from fish docs and other fish completions, it behaves differently when it comes to appending `=` to an option unlike bash and zsh. Adding a submenu for `--output-format` in case of fish gives a completion like this (notice the space after `=`)
> 
> 
> 
> It looks ugly in my opinion, and I'm not sure even functional/valid. On the other hand, it works properly for bash and zsh.
> It's your call now ~if we should keep it or not.~ about where to proceed with fish completion regarding this concern. :)

Inserting the `=` is incorrect in this case. Remove it.

completions/quick-lint-js.fish
```diff
@@ -1,13 +1,11 @@
...
-complete -c quick-lint-js -s h -l help -d 'Print help message' -f
-complete -c quick-lint-js -s v -l version -d 'Print version information' -f
-complete -c quick-lint-js -l lsp -l lsp-server -d 'Run quick-lint-js in LSP server mode' -r
-complete -c quick-lint-js -l exit-fail-on -d 'Fail with a non-zero exit code if any of these errors are found (default: "all")' -r
-complete -c quick-lint-js -l output-format -d 'Format to print feedback where FORMAT is one of: gnu-like (default if omitted), vim-qflist-json' -r
+complete -c quick-lint-js -l help -d 'Print help message' -f
+complete -c quick-lint-js -l version -d 'Print version information' -f
+complete -c quick-lint-js -l lsp-server= -d 'Run quick-lint-js in LSP server mode' -r
+complete -c quick-lint-js -l exit-fail-on= -d 'Fail with a non-zero exit code if any of these errors are found (default: "all")' -r
+complete -c quick-lint-js -l output-format= -d 'Format to print feedback where FORMAT is one of: gnu-like (default if omitted), vim-qflist-json' -xa 'gnu-like vim-qflist-json'
```
strager: Blocking: Please fix completion by removing the `=`s.

strager (changes requested):
Aside from the `Blocking` comment, this looks good to me!

strager (approved):

Issue #263, opened 2021-05-04
add support for reading from standard input


wgrr:
quick lint js can only get data from files, reading from stdin streams is valuable for some use cases, like the shell:
```
# verify if we didn't screw up an automated rewrite
sed 's/foo/bar/' file.js | quick-lint-js
```
curretnly without shell specific extension it's required to create temporaries files for achieving such use case.

PR #264, opened 2021-05-04
plugins: initial emacs support

wgrr:
List of noticiable missing that I want to have working in emacs.
  - flycheck errors explainer give documentation pointer for [EN] errors, I noticed there's in docs/errors/ with a few documented, does this installs along with quick lint js somewhere in disk? or we use network? e.g: https://quick-lint-js.com/ERRORN?
  - server-client alike protocol, some inspiration might be [rtags](https://github.com/Andersbakken/rtags).
    - Notes: recent versions of emacs includes json parsing library and [jsonrpc](https://www.jsonrpc.org/) protocol builtin, those were added in emacs 27.1, the problem (for now) of relying on those is that popular distros like current debian stable (buster), current LTS ubuntu (focal) and latest linux mint 20.1 (Ulyssa) is using emacs 26.3, not to mention olders but with notiable userbase like ubuntu 18.04
  - [MELPA](https://melpa.org/), it's by far the most popular emacs packages repository, and makes far easier to install this plugin, adding a package there is trivial, but i think the code here needs a few adaptions to meet their [criteria](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org), if im not mistaken, this also requires a new repository for plugins/emacs folder.
  - `--standard-input`` support, needs #263 fixed
  - Windows tests, currently i don't have an easy way to do this, also i don't know how to automate this with github CI

Maybe worthwhile:
  - more realible error pattern detection instead of regexp is desireable, currently it works well enough.
  - Flymake backend, flymake is emacs builtin, but its hard to deal with for users and us, but it's standard, it might be good to have one. A high level comparion with flycheck can be found here https://www.flycheck.org/en/latest/user/flycheck-versus-flymake.html


wgrr:
I'm not familiar with github actions, i bet i made mistakes in workflows/test-emacs-plugin.yml, double check that =)

wgrr:

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+
+(defgroup flycheck-quicklintjs nil
+  "Quick Lint JS Flycheck integration."
+  :prefix "flycheck-"
+  :group 'flycheck
+  :group 'quicklintjs
+  :link '(url-link :tag "Website" "https://quick-lint-js.com"))
+
+(flycheck-def-args-var flycheck-quicklintjs-args javascript-quicklintjs)
+(flycheck-def-executable-var javascript-quicklintjs "quick-lint-js")
```
strager, line 55: If quick-lint-js is installed locally in a project via npm, ideally we should use the npm-installed version instead of the globally-installed version. (This would mean that some buffers might use the npm-installed version and others might not. So maybe this kind of setup isn't possible with flycheck.)

Vim's ALE has a mechanism to dynamically search for the linter per-buffer. Does Flycheck have a similar mechanism? If not, I think we should mention this limitation in the documentation. Specifically, something like "If you installed quick-lint-js in your project using npm, you need to manually configure the Emacs plugin to use your project's version. Here's how: ...".

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+
+;; This flycheck exetension configures flycheck to run Quick Lint JS
+;; to run as a checker.
+;;
+;; Example usage in your init.el:
+;;
+;; (require 'flycheck-quicklintjs)
+;;
+;; ;; Optionally you can setup flycheck with the expected
+;; ;; quick lint js experience:
```
strager, line 13: *quick-lint-js

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
+;;; flycheck-quicklintjs --- Quick Lint JS flycheck support   -*- lexical-binding: t; -*-
+
+;;; Commentary:
+
+;; This flycheck exetension configures flycheck to run Quick Lint JS
```
strager, line 5: Is it "Flycheck" or "flycheck"

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+;;
+;;   ;; Use quick-lint-js by default when in 'js-mode`
+;;   (flycheck-select-checker 'javascript-quicklintjs)
+;;
+;;   ;; Remove any delay after a change in buffer to run checkers.
+;;   ;; The default is 0.5 (500ms)
+;;   (setq-local flycheck-idle-change-delay 0)
+;;
+;;   ;; Arguments to quick-lint-js program
+;;   ;; (setq-local flycheck-quicklintjs-args "--output-format=gnu-like")
```
strager, line 29: Blocking: Don't suggest that the user change the `--output-format`. That'll break our plugin's parsing!

quick-lint-js doesn't yet have any options which makes sense for a user to customize. Therefore, let's not suggest any options.

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+;;
+;;   ;; Change the quick-lint-js program to run.
+;;   ;; If not set, Emacs executable-find function is used, which searches in
+;;   ;; exec-path variable, probably you don't need to set this if quick-lint-js
+;;   ;; is installed in your system.
+;;   ;; (setq-local flycheck-javascript-quicklintjs-executable "~/quick-lint-js")
+;;
+;;   ;; Run quick-lint-js program when the buffer is changed and when 'js-mode` is
+;;   ;; loaded
+;;   (setq-local flycheck-check-syntax-automatically '(mode-enabled idle-change)));
```
strager, line 39: A trailing semicolon? In Elisp?!?! 😱

.github/workflows/test-emacs-plugin.yml
```diff
@@ -0,0 +1,47 @@
...
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: version
+        run: emacs --version
+
+      - name: test
+        run: ./plugin/emacs/test-emacs.sh
```
strager, line 29: This won't work because quick-lint-js isn't installed. You'll need to do something like .github/workflows/test-install.yml does to install quick-lint-js first.

.github/workflows/test-emacs-plugin.yml
```diff
@@ -0,0 +1,47 @@
...
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  build:
+    name: Emacs
+    runs-on: ubuntu-latest
+    container: ghcr.io/quick-lint/quick-lint-js-github-builder:v1
```
strager, line 14: This version of the container doesn't include your changes. I'll upload a version for you to test.

strager (changes requested):
This looks pretty good. Though I'm not an Emacs user, so I don't know if it *feels* good to a veteran.

Can you include a screenshot showcasing the plugin in action? This example might be interesting:
```javascript
const occupation = language + " coder";
const language = "JavaScript";
occupation = "Senior " + occupation;

console.log(`Welcome, ${ocupation}!`);
```

> flycheck errors explainer give documentation pointer for [EN] errors, I noticed there's in docs/errors/ with a few documented, does this installs along with quick lint js somewhere in disk? or we use network? e.g: https://quick-lint-js.com/ERRORN?

That's a cool feature! quick-lint-js doesn't have anything yet though. The files in docs/errors/ are a work in progress and aren't published or installed anywhere.

---

Does Flycheck have an LSP mode? If so, let's try using it. It might simplify the plugin.

.github/workflows/test-emacs-plugin.yml
strager: I published `ghcr.io/quick-lint/quick-lint-js-github-builder:v2-wip` which has emacs installed (per your Dockerfile).

plugin/emacs/flycheck-quicklintjs.el
wgrr: I see, but in fact this isn't how we define the program to run, we define below as a literal string "quick-lint-js",  in fact this is a macro which its expansion will define flycheck-SYMBOL-executable, in this case flycheck-javascript-quicklintjs-executable, it's a variable meant for the user to change, not us, it's here so users inspect what the default value is with emacs builtin help system (C-h v).

Flycheck is a emacs minor mode, as far as im concern, every minor mode is **buffer-local**, meaning, it's possible to change dynamically which linter to run, the problem here is that emacs as vim is a text editor, it's doesnt have a notion of a "project", just open text files, some emacs users use a popular package named projectile for that.  But i see that we can change :command to be a generic lisp function, which traverses back the parent directories until it founds a node_modules with quick-lint-js? im not sure if it's what general users expect, but im sure it's what a npm user would expect.

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+                    (zero-or-more whitespace) "error:" (zero-or-more whitespace)
+                    (message) line-end)
+                   (warning
+                    line-start (file-name) ":" line ":" column ":"
+                    (zero-or-more whitespace) "warning:" (zero-or-more whitespace)
+                    (message) line-end)
+                   (info
+                    line-start (file-name) ":" line ":" column ":"
+                    (zero-or-more whitespace) "note:" (zero-or-more whitespace)
+                    (message) line-end))
```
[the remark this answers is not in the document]
wgrr: notes does underscores as green in the match (errors are red, warning orange), i dont think has any association with previous warning or error.

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -0,0 +1,101 @@
...
+
+(flycheck-define-checker javascript-quicklintjs
+  "Quick Lint JS finds bugs in JavaScript programs.
+
+https://quick-lint-js.com"
+  :command	("quick-lint-js"
+			 (eval flycheck-quicklintjs-args)
+			 source)
+  :standard-input 't
+  :error-patterns ((error
```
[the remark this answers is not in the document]
wgrr: Im not sure if i understood correctly, i think it's highling the full error, see the screenshoot and correct me i missed something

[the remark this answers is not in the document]
wgrr: lgtm as of the screenshots, most of character counting stuff is handled by emacs, not flycheck, so i suppose it should "just" "work".

plugin/emacs/flycheck-quicklintjs.el
wgrr: In their website, documentation and github they refer with a upper case F, so i tried to maintain that, i missed this one.
ps: but not in lisp code, because gnu lisp style never uses upper cases letters, at least i never saw one, since this is a comment it should be upper case indeed.

plugin/emacs/flycheck-quicklintjs.el
strager: >  But i see that we can change :command to be a generic lisp function, which traverses back the parent directories until it founds a node_modules with quick-lint-js? im not sure if it's what general users expect, but im sure it's what a npm user would expect.

That solution makes sense to me.

What does ESLint's Emacs plugin do?

plugin/emacs/flycheck-quicklintjs.el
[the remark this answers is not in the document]
strager: > Im not sure if i understood correctly, i think it's highling the full error, see the screenshoot and correct me i missed something

What screenshot? Did you share it in a different comment?

wgrr:
> Can you include a screenshot showcasing the plugin in action? This example might be interesting:

You meant for the readme like vscode has? i attached some here for reviewing purposes, might be good enough for a presentation.

2021-05-04-052218_898x675_scrot

2021-05-04-051459_898x675_scrot

2021-05-04-051312_898x675_scrot


> Does Flycheck have an LSP mode? If so, let's try using it. It might simplify the plugin.

No, but the most popular LSP implementation does intregates out of the box with flycheck, and recommends it over flymake: https://github.com/emacs-lsp/lsp-mode#overview

To me a better looking LSP implementation is https://github.com/joaotavora/eglot ; This doesn't use flycheck instead it populates the builtin flymake with errors sent from the server, also its owner Joao is a well known contribuitor to emacs, he maintains lots of builtin lisp in emacs (like flymake) and also authored many, another pro is that it's on the gnu official ELPA plugin repository, its a very, very restrict list.
ps: (if im not mistaken I recall emails from him in emacs-devel that he tried to push eglot as a builtin to emacs, but the only piece that got into it was its jsonrpc library)

strager:
> You meant for the readme like vscode has? i attached some here for reviewing purposes, might be good enough for a presentation.

Nah, I just wanted to see for myself. Your screenshots were what I wanted, thanks.

> No, but the most popular LSP implementation does intregates out of the box with flycheck, and recommends it over flymake: https://github.com/emacs-lsp/lsp-mode#overview
> 
> To me a better looking LSP implementation is https://github.com/joaotavora/eglot

Could we integrate with all of them? =] (That's what I plan on doing for Vim: supporting Syntastic, ALE, coc.nvim, and anything else.) (Doesn't have to be in this PR.)

wgrr:
@strager Just to clarify some emacs specific context things so we can do better decisions here

MELPA is the "comunnity driven" version of GNU ELPA, both are plain tarballs fetched from a central repo by the builtin package manager named "package.el", MELPA builds tarballs from recipes which pulls from git/hg/etc daily and make available in https://stable.melpa.org/packages (if you dont have the emacs user-agent it redirects you, lame).
ELPA (https://elpa.gnu.org/packages/) is the GNU official repo with very few packages, it's unrealistic to be accepted there, first off all packages there are considered part of emacs, so it must follow GNU standards and a copyright transfer to the FSF is required, but thats easy part, only one package proving X feature can live there, and i think packages accepted there extended emacs to do X, not extension to Y. (even popular packages like flycheck aren't in ELPA because there's flymake already)

Another good option to consider is merging our flycheck extension upstream, i just looked up flycheck website and they look like to be ok with community extension (https://www.flycheck.org/en/latest/community/extensions.html), we can indeed try to merge our checker there, so it works for users of flycheck out-of-the box. The only main concern flycheck maintainer have is that the checker arbitrary executing in-buffer code, apparently one of exlir checkers were removed because of that, not a problem for quick-lint-js

There's also the community "straight.el" package manager, (https://github.com/raxod502/straight.el) that we might want to support, it's easier because it's vim-sh like, it just pulls right from github, but it does also pull from melpa, so getting merged into melpa we already support this one

Shouldn't be hard to support all of them, some numbers of interest sourced from here: https://emacssurvey.org/2020
How do you manage third-party lisp?
  46% `use-package` (it's a wrapper on top o package.el)
  28% `package.el`
  15% `straight.el`

How do you get Emacs packages
  melpa 72%
  elpa 47%
  melpa-stable 30%
  direct (includes straight.el) 26%

What package do you use for error checking?
  flycheck 60%
  flymake 13%

Which language you program in:
  javascript 43%

Do you use a language server with lsp-mode or eglot? With what languages? (% ommited because its is hard to reason since it's a free text questions and spread all over the place)
  `lsp-mode` appears 35 times, `eglot` 4
  `javascript` appears in the list with `10`, but its spread all over the place with co-languages like typescript, go and python, i didn't bother to sum up but indeed its on top 5 for lsp users

plugin/emacs/flycheck-quicklintjs.el
wgrr: @strager here's what flycheck.el does for eslint:
```
  :command ("eslint" "--format=json"
            (option-list "--rulesdir" flycheck-eslint-rules-directories) ;; <- a customizablie list of directories that contains eslint rules, what that means.
            (eval flycheck-eslint-args)
            "--stdin" "--stdin-filename" source-original)
  :enabled (lambda () (flycheck-eslint-config-exists-p)) ;; <- checks for .eslintrc, apparently
```
it doesnt look like to make anything more than that, probably it's leftover for users to overwrite this with a js-mode hook that does that.

wgrr: self correcting me:
> it's doesnt have a notion of a "project"

I just learned from the survey i linked below about the builtin "package.el", it was introduced in emacs 26.1, i'll need investigate that if it's of any value for us.

.github/workflows/test-emacs-plugin.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: build
+        run: cmake --build . --config Debug
+      - name: install
+        run: sudo cmake --install . --config Debug
+      - name: test executable
+        run: quick-lint-js --help
```
strager, line 30: Nit: I wouldn't bother checking that `quick-lint-js` installed. But I'm not super against it.

strager (approved):
Let me know when I should land. (When I land, I'll create a :v2 Docker image and I'll update .github/workflows/test-emacs-plugin.yml to refer to it. I'll also squash your commits; let me know if this is a problem.)

wgrr:
@strager squashed the commits, feel free to merge if think it's good

strager:
Landed as commit ef642dfed11f8a931a19a176627af65ece67fae2. Thanks!

PR #268, opened 2021-05-05
Report error on unicode BOM before #!

### LINKED ISSUE #205: 7$: Report error on BOM before #!
BOM: Byte Order Mark
#!: Shebang

Linux (and I assume macOS and other POSIX-like platforms) only interpret shebangs if the first byte in the file is `#` and the second byte in the file is `!` (I think; please confirm). This means that the presence of a BOM (UTF-8-encoded) will cause the shebang characters to not be interpreted by the operating system. This is almost definitely a mistake.

Report an error if this mistake occurs.

singalhimanshu:
Fixes #205

### DIFF
--- a/docs/errors/E095.md
+++ b/docs/errors/E095.md
@@ -1,0 +1,13 @@
+# E095: unicode byte order mark (BOM) cannot appear before #! at beginning of script
+
+A script cannot have unicode byte order mark (BOM) before `#!`
+
+    #!/usr/bin/env node
+    let x = 3;
+    console.log(x);
+
+To fix this error, remove the unicode BOM before `#!`
+
+    #!/usr/bin/env node
+    let x = 3;
+    console.log(x);
--- a/docs/errors/E095.md
+++ b/docs/errors/E095.md
@@ -1,12 +1,12 @@
-# E095: unicode byte order mark (BOM) cannot appear before #! at beginning of script
+# E095: Unicode byte order mark (BOM) cannot appear before #! at beginning of script
 
-A script cannot have unicode byte order mark (BOM) before `#!`
+A script cannot have Unicode byte order mark (BOM) before `#!`
 
     #!/usr/bin/env node
     let x = 3;
     console.log(x);
 
-To fix this error, remove the unicode BOM before `#!`
+To fix this error, remove the Unicode BOM before `#!`
 
     #!/usr/bin/env node
     let x = 3;
--- a/test/test-lex.cpp
+++ b/test/test-lex.cpp
@@ -1876,18 +1876,6 @@ TEST_F(test_lex, lex_not_shebang) {
                               offsets_matcher(&input, 2, 3))));
   }
 
-  // BOM must not appear before '#!'.
-  {
-    error_collector v;
-    padded_string input(u8"\ufeff#!notashebang\n"_sv);
-    lexer l(&input, &v);
-    EXPECT_EQ(l.peek().type, token_type::end_of_file) << "# should be skipped";
-
-    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
-                              error_unexpected_bom_before_shebang, where,
-                              offsets_matcher(&input, 0, 2))));
-  }
-
   {
     error_collector v;
     padded_string input(u8"#\\u{21}\n"_sv);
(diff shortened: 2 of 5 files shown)


test/test-lex.cpp
```diff
@@ -1876,6 +1876,18 @@ TEST_F(test_lex, lex_not_shebang) {
...
 
+  // BOM must not appear before '#!'.
+  {
+    error_collector v;
+    padded_string input(u8"\ufeff#!notashebang\n"_sv);
+    lexer l(&input, &v);
+    EXPECT_EQ(l.peek().type, token_type::end_of_file) << "# should be skipped";
+
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_unexpected_bom_before_shebang, where,
```
strager, line 1887: Rename `where` to `bom`.

test/test-lex.cpp
```diff
@@ -1876,6 +1876,18 @@ TEST_F(test_lex, lex_not_shebang) {
                               offsets_matcher(&input, 2, 3))));
   }
 
+  // BOM must not appear before '#!'.
```
strager, line 1879: Make this a separate `TEST` and put this documentation in the test's name.

test/test-lex.cpp
```diff
@@ -1876,6 +1876,18 @@ TEST_F(test_lex, lex_not_shebang) {
...
+  {
```
strager, line 1880: Also test BOM without `#!`. (It should be ignored, producing no error.)

test/test-lex.cpp
```diff
@@ -1876,6 +1876,18 @@ TEST_F(test_lex, lex_not_shebang) {
...
+    error_collector v;
+    padded_string input(u8"\ufeff#!notashebang\n"_sv);
+    lexer l(&input, &v);
+    EXPECT_EQ(l.peek().type, token_type::end_of_file) << "# should be skipped";
+
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_FIELD(
+                              error_unexpected_bom_before_shebang, where,
+                              offsets_matcher(&input, 0, 2))));
```
strager, line 1888: Blocking: Shouldn't the error cover all three bytes of the BOM, not just the first two bytes?

Write the test like this, which should make the issue more obvious:
```suggestion
                              offsets_matcher(&input, 0, u8"\ufeff"))));
```

strager (changes requested):
Looks good aside from the weird error span.

test/test-lex.cpp
singalhimanshu: Done

test/test-lex.cpp
singalhimanshu: Shouldn't this belong to - https://github.com/quick-lint/quick-lint-js/blob/6c54dc042d691bbbe6d0c83115b5cd6e907c92bd/test/test-lex.cpp#L1842 as the name suggest `lex_not_shebang`. Any particular reason to move it as a separate `TEST`?

test/test-lex.cpp
singalhimanshu: I guess there it already a test for that here - https://github.com/quick-lint/quick-lint-js/blob/6c54dc042d691bbbe6d0c83115b5cd6e907c92bd/test/test-lex.cpp#L1836

test/test-lex.cpp
singalhimanshu: yeah makes sense

test/test-lex.cpp
strager: > Any particular reason to move it as a separate `TEST`?

Yes. You thought the test was different enough to write a comment, so I think it's different enough to put into its own `TEST` block.

Also, all of the other tests are expecting an `error_unexpected_hash_character`, implying that the `#!` was not parsed as a comment. Your test is parsing the `#!` as a comment.

test/test-lex.cpp
strager: I meant test `\ufeffStuff`, not `#!Stuff`.

singalhimanshu: oh okay, I misread that.

singalhimanshu: > I meant test `\ufeffStuff`, not `#!Stuff`.

Isn't this function testing that already - https://github.com/quick-lint/quick-lint-js/blob/6c54dc042d691bbbe6d0c83115b5cd6e907c92bd/test/test-lex.cpp#L1781

Let me know if I'm wrong.

docs/errors/E095.md
```diff
@@ -0,0 +1,13 @@
+# E095: unicode byte order mark (BOM) cannot appear before #! at beginning of script
+
+A script cannot have unicode byte order mark (BOM) before `#!`
+
+    #!/usr/bin/env node
+    let x = 3;
+    console.log(x);
+
+To fix this error, remove the unicode BOM before `#!`
```
strager, line 9: Nit: "Unicode" is capitalized:
```suggestion
A script cannot have Unicode byte order mark (BOM) before `#!`

    #!/usr/bin/env node
    let x = 3;
    console.log(x);

To fix this error, remove the Unicode BOM before `#!`
```

strager (approved):

docs/errors/E095.md
singalhimanshu: Done!

singalhimanshu:
I think it is ready to merge

PR #287, opened 2021-05-12
Disable variable lookup in presence of eval

### LINKED ISSUE #95: 45$: Disable variable lookups in the presense of eval

singalhimanshu:
Fixes #95 

The following code doesn't report an error now
```js
eval("var x = 42;")
console.log(x);
```


src/lint.cpp
```diff
@@ -280,11 +280,23 @@ void linter::visit_exit_for_scope() {
...
+  }
+  return false;
+}
+
 void linter::visit_exit_function_scope() {
   QLJS_ASSERT(!this->scopes_.empty());
-  this->propagate_variable_uses_to_parent_scope(
-      /*allow_variable_use_before_declaration=*/true,
-      /*consume_arguments=*/true);
+  if (!is_eval_used()) {
```
strager, line 295: I think this approach will cause false negatives for code like the following:
```js
let eval = console.log;
function g() {
  eval("hello");
  undeclaredVariable;
}
g();
```
False negatives (i.e. not reporting an error) are better than false positives (i.e. reporting a bogus error), so I'm fine with this tradeoff if we can't find a better solution.

singalhimanshu: I have tried to come up with some logic to handle this, let me know if it fails in any other condition.

singalhimanshu: I just realized that it has a bug in the following code:

```js
function f() {
  eval("var x = 42;")
}
f();

function g() {
  undeclaredVariable;    // quick-lint-js should report an error here but it doesn't
}
```

I guess we will have to make a trade-off here, with the current system it doesn't seem that we can handle all the cases.

strager:
Is this ready for review, or are you still working on it?

If the former, you should add a lot of tests.

singalhimanshu:
> Is this ready for review, or are you still working on it?
> 
Yes it is ready now.

> If the former, you should add a lot of tests.

I have added test cases let me know if some important conditions are left untested.

test/test-lint.cpp
```diff
@@ -2811,6 +2811,201 @@ TEST(
...
+  }
+
+  {
+    // function f() {
+    //   x;
+    //   x = 10;
+    //   eval("var x = 42;");
+    //   x;
+    //   x = 10;
+    // }
```
strager, line 2895: I think this test deserves some explanation. For example:

> In this case, we do know that the first references to `x` will error. However, the current algorithm is conservative, because variable references before `eval` might be executed after the `eval` (e.g. if there is a loop and the earlier references are conditional).

test/test-lint.cpp
```diff
@@ -2811,6 +2811,201 @@ TEST(
...
+
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+  {
+    // function f() {
+    //   eval("var x = 42;");
+    // }
+    // function g() {
+    //   x;
+    // }
```
strager, line 2943: I'd put this test (and a few following tests) in a different `TEST`. `disable_variable_lookup_in_presence_of_eval` doesn't describe the behavior being tested by this example.

test/test-lint.cpp
```diff
@@ -2811,6 +2811,201 @@ TEST(
...
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(error_use_of_undeclared_variable, name,
+                                     span_matcher(use)),
+                    ERROR_TYPE_FIELD(error_assignment_to_undeclared_variable,
+                                     assignment, span_matcher(use))));
+  }
+}
 }
```
strager, line 3009: More silencing test cases I thought of:
```js
function f(a = eval('var x = 42;')) {
  console.log(x); // no error
}
f();
console.log(x); // ERROR (use of undeclared variable)
```

More erroring test cases I thought of:
```js
function f() {
  const x = 42;
  {
    eval("var x = 0;");
    x = 3;  // ERROR (assignment to const variable)
  }
}
f();
```
```js
let eval = console.log;
eval("var x = 42");
x;  // ERROR (use of undeclared variable)
x = 10;  // ERROR (assignment to undeclared variable)
```
```js
let eval = console.log;
function f() {
  eval("var x = 42");
  x;  // ERROR (use of undeclared variable)
  x = 10;  // ERROR (assignment to undeclared variable)
}
f();
```
```js
function f() {
  eval = console.log;
  eval("var x = 42;");
  x;  // ERROR (use of undeclared variable)
  x = 10;  // ERROR (assignment to undeclared variable)
  {
    var eval;
  }
}
f();
```

Even if these erroring tests don't work as desired (i.e. even if the errors are silenced), I think we should still include them to document the current behavior. In the future, when we want to fix the false negatives, we can tweak the already-written tests. (The test on line 2889 falls in this bucket IMO.) If we do include them, I'd create another `TEST` so their false-negative-ness is more apparent.

test/test-lint.cpp
```diff
@@ -2811,6 +2811,201 @@ TEST(
...
+    l.visit_end_of_module();
+
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+  {
+    // function f() {
+    //   eval("var x = 42;");
+    // }
+    // function g() {
+    //   x;
```
strager, line 2942: For tests with errors, I like to include a comment pointing out where the error is:
```suggestion
    //   x;  // ERROR
```

src/lint.cpp
```diff
@@ -500,11 +519,21 @@ void linter::propagate_variable_uses_to_parent_scope(
...
+    if (used_var.name.normalized_name() == u8"eval") {
+      if (propagate_eval_use) {
+        (allow_variable_use_before_declaration
+             ? parent_scope.variables_used_in_descendant_scope
+             : parent_scope.variables_used)
+            .emplace_back(used_var);
+      }
+      break;
+    } else if (has_eval) {
+    } else if (var) {
```
strager, line 536: This is a bit odd. `this->is_eval_used` loops through `current_scope.variables_used` to find `"eval"`. Then this code loops through `current_scope.variables_used` again to find `"eval"`, skipping other variables if we found `"eval"` previously.

I think you meant to write this code differently, e.g. by skipping the second loop entirely if we find `"eval"` in the first loop:
```suggestion
bool has_eval = false;
for (const used_variable &used_var : current_scope.variables_used) {
      if (used_var.name.normalized_name() == u8"eval") {
          (allow_variable_use_before_declaration
             ? parent_scope.variables_used_in_descendant_scope
             : parent_scope.variables_used)
            .emplace_back(used_var);
          has_eval = true;
          break;
      }
}
if (!has_eval) {
  for (const used_variable &used_var : current_scope.variables_used) {
    QLJS_ASSERT(!current_scope.declared_variables.find(used_var.name));
    const declared_variable *var =
        parent_scope.declared_variables.find(used_var.name);
    if (var) {
```

src/lint.cpp
```diff
@@ -500,11 +519,21 @@ void linter::propagate_variable_uses_to_parent_scope(
...
 
+  bool has_eval = this->is_eval_used();
   for (const used_variable &used_var : current_scope.variables_used) {
     QLJS_ASSERT(!current_scope.declared_variables.find(used_var.name));
     const declared_variable *var =
         parent_scope.declared_variables.find(used_var.name);
-    if (var) {
+    if (used_var.name.normalized_name() == u8"eval") {
+      if (propagate_eval_use) {
+        (allow_variable_use_before_declaration
```
strager, line 529: I think `allow_variable_use_before_declaration` is always `false` if `propagate_eval_use` is `true`. Do you think that is significant? Should we assert it?
```suggestion
        QLJS_ASSERT(allow_variable_use_before_declaration);
        (allow_variable_use_before_declaration
```

If we assert that `allow_variable_use_before_declaration` is always `true` when propagating `eval`, then I think we can delete the changes below in the `current_scope.variables_used_in_descendant_scope` loop (lines 560-569). But maybe not. (The logic is complicated. xD)

test/test-lint.cpp
```diff
@@ -2811,6 +2811,201 @@ TEST(
...
+    l.visit_variable_use(identifier_of(use_eval));
+    l.visit_exit_block_scope();
+    l.visit_variable_use(identifier_of(use));
+    l.visit_variable_assignment(identifier_of(use));
+    l.visit_exit_function_scope();
+    l.visit_end_of_module();
+
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+  {
```
strager, line 2937: Nit: Add a blank line.
```suggestion
  }
  
  {
```

strager (approved):
Looks good, but a bit rough. I think some of the roughness is inherit in the problem, though. =S

I think your test coverage is good, but this deserves more tests (at the very least, the example with the `eval` call in the parameter list).

src/lint.cpp
singalhimanshu: > QLJS_ASSERT(allow_variable_use_before_declaration);

did you mean this -
```suggestion
        (!allow_variable_use_before_declaration);
```

singalhimanshu: > I think `allow_variable_use_before_declaration` is always `false` if `propagate_eval_use` is `true`. Do you think that is significant? Should we assert it?

For `visit_enter_function_scope_body` this doesn't follow now, I realized this after writing the test for that condition.

src/lint.cpp
```diff
@@ -519,63 +509,68 @@ void linter::propagate_variable_uses_to_parent_scope(
...
-    } else if (has_eval) {
-    } else if (var) {
-      // This variable was declared in the parent scope. Don't propagate.
-      if (used_var.kind == used_variable_kind::assignment) {
-        this->report_error_if_assignment_is_illegal(
-            var, used_var.name, /*is_assigned_before_declaration=*/false);
+  bool has_eval_in_descendant_scope =
+      has_eval_in_current_scope
+          ? true
+          : has_eval(current_scope.variables_used_in_descendant_scope);
```
strager, line 557: Nit:
```suggestion
  bool has_eval_in_descendant_scope =
      has_eval_in_current_scope
          || has_eval(current_scope.variables_used_in_descendant_scope);
```

test/test-lint.cpp
```diff
@@ -3005,6 +3013,146 @@ TEST(test_lint_eval, disable_variable_lookup_in_presence_of_eval) {
...
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(error_use_of_undeclared_variable, name,
+                                     span_matcher(use)),
+                    ERROR_TYPE_FIELD(error_assignment_to_undeclared_variable,
+                                     assignment, span_matcher(use))));
+  }
+}
+
+TEST(test_lint_eval, false_negatives_on_redeclaration_of_eval) {
```
strager, line 3102: Add a TODO comment to fix these (eventually...)

test/test-lint.cpp
```diff
@@ -3005,6 +3013,146 @@ TEST(test_lint_eval, disable_variable_lookup_in_presence_of_eval) {
...
+  const char8 use[] = u8"x";
+
+  {
+    const char8 eval_declaration[] = u8"eval";
+
+    // let eval = console.log;
+    // (function() {
+    //   eval("var x = 42;");
+    //   x;  // should be ERROR (use of undeclared variable)
+    //   x = 10;  // should be ERROR (assignment to undeclared variable)
```
strager, line 3113: Nit: s/should be ERROR/TODO: ERROR/ ?

test/test-lint.cpp
```diff
@@ -3005,6 +3013,146 @@ TEST(test_lint_eval, disable_variable_lookup_in_presence_of_eval) {
...
+}
+
+TEST(test_lint_eval, false_negatives_on_redeclaration_of_eval) {
+  const char8 use_eval[] = u8"eval";
+  const char8 use[] = u8"x";
+
+  {
+    const char8 eval_declaration[] = u8"eval";
+
+    // let eval = console.log;
```
strager, line 3109: Nit: Visits don't match the code.
```suggestion
    // let eval = () => {};
```

strager (approved):

test/test-lint.cpp
singalhimanshu: Should I put a comment for this test?

singalhimanshu:
This is ready to merge from my side. Let me know if there are any required changes.

test/test-lint.cpp
strager: Sure

strager (approved):

PR #288, opened 2021-05-12
Support reading file from standard input

wgrr:


src/file.cpp
```diff
@@ -224,6 +229,11 @@ read_file_result read_file(const char *path) {
   posix_fd_file file(fd);
   return read_file(path, file.ref());
 }
+
+read_file_result read_stdin() {
+  posix_fd_file file(STDIN_FILENO);
```
strager, line 234: Almost-blocking: `posix_fd_file` is the wrong class to use. `posix_fd_file::~posix_fd_file` will close the file descriptor, which we probably don't want to do. Use `posix_fd_file_ref` instead.

(Same comment applies to `windows_handle_file`.)

src/main.cpp
```diff
@@ -150,13 +150,20 @@ void handle_options(quick_lint_js::options o) {
...
+  if (o.files_to_lint.empty() && !o.stdinput) {
     std::cerr << "error: expected file name\n";
     std::exit(EXIT_FAILURE);
   }
 
   quick_lint_js::any_error_reporter reporter =
       quick_lint_js::any_error_reporter::make(o.output_format, &o.exit_fail_on);
+  if (o.stdinput) {
+    quick_lint_js::read_file_result source = quick_lint_js::read_stdin();
+    file_to_lint fstdin{.path = "<stdin>", .vim_bufnr = std::nullopt};
```
strager, line 162: Almost-blocking: I would expect `quick-lint-js --vim-file-bufnr=42 --stdin` to have `--vim-file-bufnr=42` apply to stdin. I think we need to process `--stdin` as if it was a file in `options::files_to_lint`.

src/main.cpp
```diff
@@ -394,6 +401,7 @@ void print_help_message() {
   print_option("--exit-fail-on=[CODES]",
                "Fail with a non-zero exit code if any of these");
   print_option("", "errors are found (default: \"all\")");
+  print_option("--s, --stdin", "Read standard input as a javascript file");
```
strager, line 404: Nit: Capitalize "JavaScript" correctly.
```suggestion
  print_option("--s, --stdin", "Read standard input as a JavaScript file");
```

strager, line 404: Let's also update docs/cli.adoc.

src/options.cpp
```diff
@@ -99,6 +99,10 @@ class arg_parser {
       this->is_ignoring_options_ = true;
       this->option_ = std::nullopt;
     } else if (this->current_arg()[0] == '-') {
+      if (std::strcmp(this->current_arg(), "-") == 0) {
+        this->option_ = std::nullopt;
+        return;
+      }
```
strager, line 105: I think checking for `-` like on line 94 would look better:
```suggestion
    } else if (std::strcmp(this->current_arg(), "-") == 0) {
      this->option_ = std::nullopt;
    } else if (this->current_arg()[0] == '-') {
```

src/options.cpp
```diff
@@ -99,6 +99,10 @@ class arg_parser {
       this->is_ignoring_options_ = true;
       this->option_ = std::nullopt;
     } else if (this->current_arg()[0] == '-') {
+      if (std::strcmp(this->current_arg(), "-") == 0) {
```
strager, line 102: Nit: `this->current_arg()` is a string_view, so we can compare with `==`:
```suggestion
      if (this->current_arg() == "-"sv) {
```

src/options.cpp
```diff
@@ -188,6 +196,8 @@ options parse_options(int argc, char** argv) {
       o.version = true;
     } else if (parser.match_flag_option("--lsp-server"sv, "--lsp"sv)) {
       o.lsp_server = true;
+    } else if (parser.match_flag_option("--stdin"sv, "--s"sv)) {
```
strager, line 199: Nit: I don't like the `--s` option. I'd get rid of it, and only allow `--stdin`:
```suggestion
    } else if (parser.match_flag_option("--stdin"sv, "--stdin"sv)) {
```

src/main.cpp
```diff
@@ -150,23 +150,20 @@ void handle_options(quick_lint_js::options o) {
...
-  }
   for (const quick_lint_js::file_to_lint &file : o.files_to_lint) {
-    quick_lint_js::read_file_result source =
-        quick_lint_js::read_file(file.path);
+    quick_lint_js::read_file_result source;
+    if (file.is_stdin) {
+      source = quick_lint_js::read_stdin();
+    } else {
+      source = quick_lint_js::read_file(file.path);
+    }
```
strager, line 166: :+1:

src/options.cpp
```diff
@@ -69,6 +69,18 @@ class arg_parser {
     return matches;
   }
 
+  bool match_flag_option(std::string_view option_name) noexcept {
```
strager, line 72: Nit: I'd be lazy and delegate to the overload on line 59. Option parsing isn't performance-critical.

What you have now is perfectly fine too.

src/options.cpp
```diff
@@ -152,16 +162,27 @@ options parse_options(int argc, char** argv) {
...
 
   arg_parser parser(argc, argv);
   while (!parser.done()) {
     if (const char* argument = parser.match_argument()) {
-      if (std::strcmp(argument, "-") == 0) {
-        o.stdinput = true;
-        continue;
+      if (argument == "-"sv) {
+        if (has_stdin) {
+          continue;
```
strager, line 172: Hmm, perhaps we should issue a warning if `-` or `--stdin` is given multiple times.

website/public/cli/index.html
```diff
@@ -3,444 +3,343 @@
 <!-- See end of file for extended copyright information. -->
 <!-- This file was generated using generate-cli-docs. -->
 <html lang="en">
-  <head>
```
strager, line 6: Run `cd website && yarn && yarn fmt` to reformat this code and minimize the diff.

strager (approved):
Looks good! Thanks for making the adjustments.

Let me know when I should squash 'n' merge.

src/options.cpp
wgrr: i thought it would be unnecessary noisy, done.

website/public/cli/index.html
wgrr: thats lemme that i can't run that with debian sid nodejs version (12.something), it's been 1 hour compiling node, done.

wgrr:
@strager i had to force push, i've messed up upstream/origin rebases, sorry about that
If current version looks good to you, feel free to squash the commits and merge

strager (approved):

Issue #265, opened 2021-05-05
9$: Warn on semicolon in if-else chain


strager:
The following code likely has a mistake:

```javascript
if (x) {
  f();
} else; {  // mistake: stray semicolon
  g();
}
```

produce a warning for code like this.

strager:
@someoneigna made a PR for this task. (GitHub won't let me assign them this task, though, because they haven't made a comment.) #290

PR #290, opened 2021-05-13
#265 Warn on redundant semicolon after else

someoneigna:
### Summary
Warn when a semicolon is found after else.

After this change a warning is reported for the semicolons at `7:7` `12:15` and `16:2`
```
let x = 0;
function f() {}
function g() {}

if (x) {
    f();
} else; {
    g();
}


if (x) {} else;
{ }

if (x) {} else
;{ }
 ```
 
 Output in VIM plugin format is:
```
{"qflist": [{"col": 7, "lnum": 7, "end_col": 7, "end_lnum": 7, "type": "W", "nr": "E202", "vcol": 0, "text": "redundant semicolon after else", "filename": "invalid.js"},
{"col": 15, "lnum": 12, "end_col": 15, "end_lnum": 12, "type": "W", "nr": "E202", "vcol": 0, "text": "redundant semicolon after else", "filename": "invalid.js"},
{"col": 1, "lnum": 16, "end_col": 1, "end_lnum": 16, "type": "W", "nr": "E202", "vcol": 0, "text": "redundant semicolon after else", "filename": "invalid.js"}]}
```

`else;;; {}` is also valid code but was left out for simplicity. (loop of peek and skip required).

### CODE REVIEW

**Comment by someoneigna** [comment]

File: src/quick-lint-js/parse.h [hunk 1]
```diff
@@ -2515,6 +2515,13 @@ class parser {
 
     if (this->peek().type == token_type::kw_else) {
       this->skip();
+      if (this->peek().type == token_type::semicolon) {
+        source_code_span semicolon = this->peek().span();
```
someoneigna, line 2519: Is there somewhere to move this out (`this->warn_on_redundant_semicolon();` ). I don't like the inner if.

src/quick-lint-js/error.h
```diff
@@ -1139,6 +1139,12 @@
       { source_code_span continue_statement; },                                \
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
+      error_redundant_semicolon_after_else, "E202",                            \
```
someoneigna, line 1144: Is there a convention for error/warning order and naming?

strager: > Is there a convention for error/warning order

Yes. Sort lexicographically by the struct name (`error_redundant_semicolon_after_else`)

> Is there a convention for [...] naming?

Aside from the `error_` prefix and snake case, no.

src/quick-lint-js/parse.h
```diff
@@ -2515,6 +2515,13 @@ class parser {
...
+        this->error_reporter_->report(error_redundant_semicolon_after_else{
+          .semicolon = source_code_span(
+            semicolon.begin(), semicolon.end()),
```
strager, line 2522:
```suggestion
          .semicolon = semicolon,
```

src/quick-lint-js/parse.h
strager: > Is there somewhere to move this out (`this->warn_on_redundant_semicolon();` ). I don't like the inner if.

I understand your sentiment. You can certainly make another member function in `parser`.

Normally I only make functions in this class to avoid code duplication or for major syntax elements. Therefore, what you wrote is typical style.

src/quick-lint-js/error.h
```diff
@@ -1139,6 +1139,12 @@
...
+      { source_code_span semicolon; },                                         \
+      .warning(QLJS_TRANSLATABLE("redundant semicolon after else"),            \
```
strager, line 1146: Blocking: This error message is incorrect. The extra semicolon is *not* redundant; it changes behavior.

Example:
```js
for (let b of [true, false]) {
  if (b) {
    console.log('t');
  } else /* Maybe add a semicolon here */ {
    console.log('f');
  }
  console.log('---');
}
```
Without semicolon:
```
t
---
f
---
```
With semicolon:
```
t
f
---
f
---
```

We should communicate that the semicolon is likely unintended, but also imply that removing the semicolon would change behavior.

Perhaps this message is better: "else has empty body; consider removing the semicolon or removing the 'else'" The more I think about it, the more I think we should have two different error messages:

```js
if (b) {
} else; {  // else has empty body; consider removing the semicolon to make it not empty
}
```
```js
if (b) {
} else;   // else has empty body; consider removing the redundant 'else'
{
}
```
(The first example has no line break between `;` and `{`. The second example does.)

(Perhaps the second case doesn't deserve a diagnostic at all though. Or perhaps it should be of a lesser severity than "warning".)

src/quick-lint-js/parse.h
```diff
@@ -2515,6 +2515,13 @@ class parser {
 
     if (this->peek().type == token_type::kw_else) {
       this->skip();
+      if (this->peek().type == token_type::semicolon) {
+        source_code_span semicolon = this->peek().span();
+        this->error_reporter_->report(error_redundant_semicolon_after_else{
```
strager, line 2520: Almost-blocking: I think this code reports a warning when it should report no warnings. For example, reporting this warning on the following code is very misleading:

```js
{
  if (b)
    console.log("yes");
  else
    ;
}
```

I think we should only report a diagnostic if the `;` is followed by a `{`.

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,22 @@ TEST(test_parse, else_without_if) {
...
+TEST(test_parse, else_redundant_semicolon) {
+  {
+    spy_visitor v;
+    padded_string code(u8"if (cond) { body; } else; { body; }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",      // cond
+                                      "visit_enter_block_scope",  // (if)
+                                      "visit_variable_use",       // body
+                                      "visit_exit_block_scope")); // (else)
```
strager, line 590: Nit: The comment is misleading. The `visit_exit_block_scope` is for the `if` body, not the `else` body.
```suggestion
                                      "visit_exit_block_scope")); // (if)
```

strager (changes requested):
Minor: It'd also be great if you could add documentation to docs/errors/ for this new error type. Not necessary, though.

src/quick-lint-js/parse.h
```diff
@@ -2516,7 +2516,18 @@ class parser {
 
     if (this->peek().type == token_type::kw_else) {
       this->skip();
+
+      bool semicolon_after_else = this->peek().type == token_type::semicolon;
+      const char8 *token_after_else_begin = this->peek().begin;
```
someoneigna, line 2521: Wanted to avoid copying the token. Maybe its premature optimization on my part.

docs/errors/E220.md
```diff
@@ -0,0 +1,13 @@
...
+
+A semicolon next to the `else` keyword can cause its body to be executed even when previous `if` statement condition is true.
+
+    if (true) {
+        console.log("true");
+    } else; {
+        console.log("& false");
+    }
+    > true
+    > & false
```
strager, line 11: Don't include command output in the example like this. It'll be interpreted as JavaScript code.

If you want to include the command output, make it a separate quotation paragraph:

```suggestion

> ```
> true
> & false
> ```
```

or make it a comment:

```suggestion
    // true
    // & false
```

docs/errors/E220.md
```diff
@@ -0,0 +1,13 @@
...
+
+To avoid this behavior, remove the semicolon between the else keyword and the opening brace of the else block.
```
strager, line 13: Include an example of what working code looks like.

docs/errors/E220.md
```diff
@@ -0,0 +1,13 @@
+# E220: Semicolon after else may cause unexpected execution of its body.
```
strager, line 1: Nit: Title should match the error description from error.h:
```suggestion
# E220: semicolon after else may be causing unexpected behavior
```

strager, line 1: Blocking: The error code is E202, not E220. Fix the file name and the title.

src/quick-lint-js/parse.h
```diff
@@ -2516,7 +2516,18 @@ class parser {
...
+
       parse_and_visit_body();
+      if (semicolon_after_else &&
+          this->peek().type == token_type::left_curly &&
+          !this->peek().has_leading_newline) {
```
strager, line 2526: :+1:

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,23 @@ TEST(test_parse, else_without_if) {
...
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",        // cond
+                                      "visit_enter_block_scope",   // (if)
+                                      "visit_variable_use",        // body
+                                      "visit_exit_block_scope"));  // (if)
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_unexpected_semicolon_after_else, semicolon,
+                    offsets_matcher(&code, strlen(u8"if (cond) { body; } else"), u8";"))));
+  }
+}
```
strager, line 596: Also add a test for where there is a newline after the `;`:
```js
if (cond) { body; } else;
{ body; }
```

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,23 @@ TEST(test_parse, else_without_if) {
...
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",        // cond
+                                      "visit_enter_block_scope",   // (if)
+                                      "visit_variable_use",        // body
+                                      "visit_exit_block_scope"));  // (if)
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_unexpected_semicolon_after_else, semicolon,
+                    offsets_matcher(&code, strlen(u8"if (cond) { body; } else"), u8";"))));
```
strager, line 594: :+1:

src/quick-lint-js/parse.h
```diff
@@ -2516,7 +2516,18 @@ class parser {
...
+        this->error_reporter_->report(error_unexpected_semicolon_after_else{
+            .semicolon = source_code_span(token_after_else_begin, token_after_else_begin+1)
+        });
```
strager, line 2529: Nit: If we store the whole token, I think the code is more readable:
```suggestion
      token token_after_else = this->peek();

      parse_and_visit_body();
      if (token_after_else.type == token_type::semicolon &&
          this->peek().type == token_type::left_curly &&
          !this->peek().has_leading_newline) {
        this->error_reporter_->report(error_unexpected_semicolon_after_else{
            .semicolon = source_code_span(token_after_else.begin, token_after_else.end)
        });
```

Nit: I'd rather store the token's `end` rather than assuming a semicolon ends 1 after its begin. (Suggested diff above incorporates this change.)

strager (changes requested):

src/quick-lint-js/parse.h
strager: Copying a `token` should be cheap thanks for Scalar Replacement of Aggregates optimizations. (After optimization, only fields which are used are copied, which basically has the same effect as your manually-optimized code.)

docs/errors/E202.md
```diff
@@ -0,0 +1,30 @@
...
+        console.log("else");
+    }
+
+Example output
+```
+> true
+> else
+```
+
+To avoid this behavior, remove the semicolon between the else keyword and the opening brace of the else block.
```
someoneigna, line 17: Maybe:
`In case this is impacting your code, you should remove the semicolon next to the else keyword`

src/quick-lint-js/parse.h
```diff
@@ -2516,7 +2516,27 @@ class parser {
...
+      const token expect_left_curly = this->peek();
+
+      lexer_transaction transaction = this->lexer_.begin_transaction();
+      this->lexer_.skip();
+
+      if (this->peek().type == token_type::right_curly) {
+        this->error_reporter_->report(error_else_has_empty_body{
+            .where = source_code_span(expect_left_curly.begin, this->peek().end)
+        });
+      } else if (token_after_else.type == token_type::semicolon &&
```
someoneigna, line 2531: This is the check for an empty block on the else case. Now that I read it again, I see it should only be triggered if the else has a semicolon 🤦‍♂️
https://github.com/quick-lint/quick-lint-js/pull/290#discussion_r631650388

Is creating a transaction on the lexer and then rolling it back the correct way to peek forward?

strager: > Is creating a transaction on the lexer and then rolling it back the correct way to peek forward?

We only need a transaction if you want to look ahead multiple tokens.

Can you give an example JS program where using `this->peek()` won't work?

strager:
Did you want me to review your latest changes, or are you still hacking?

src/quick-lint-js/error.h
```diff
@@ -1028,6 +1034,13 @@
           QLJS_TRANSLATABLE("for-of loop expression cannot have semicolons"),  \
           semicolon))                                                          \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_unexpected_semicolon_after_else, "E202",                           \
```
strager, line 1038: Blocking: Someone else snagged E202. Pick a different error code.

strager (changes requested):

src/quick-lint-js/error.h
someoneigna: Moved the warning about unexpected behavior due to semicolon to E204 and else has empty body to E205.

src/quick-lint-js/parse.h
someoneigna: I had gone through a rabbit hole of actually checking the body of the next scope to the else. That's why I was advancing the cursor and parsing the next scope after the else using a transaction 🤦‍♂️ . I've rolled the code back to only check for the newline and the opening brace.
Commits are now squashed, let me know what you think.

docs/errors/E204.md
```diff
@@ -0,0 +1,30 @@
...
+    } else {
+        console.log("else");
+    }
+
+Or if the else body is not required remove it completely.
+
+    if (true) {
+        console.log("true");
+    }
+    console.log("always");
```
strager, line 30: Lovely docs!

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,44 @@ TEST(test_parse, else_without_if) {
...
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",        // cond
+                                      "visit_enter_block_scope",   // (if)
+                                      "visit_variable_use",        // body
+                                      "visit_exit_block_scope"));  // (if)
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_else_has_empty_body, where,
+            offsets_matcher(&code, strlen(u8"if (cond) { body; } else {"),
+                            u8"}"))));
```
strager, line 615: Blocking: This strlen string doesn't match the code. The `u8"}"` looks wrong too. Did you mean the following instead?
```suggestion
            offsets_matcher(&code, strlen(u8"if (cond) { body; } else;\n"),
                            u8"{"))));
```

If so, the error looks misplaced. It should point at the `else` token or the `;`, not at the following unrelated `{`.

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,44 @@ TEST(test_parse, else_without_if) {
...
+                    error_unexpected_semicolon_after_else, semicolon,
+                    offsets_matcher(&code, strlen(u8"if (cond) { body; } else"),
+                                    u8";"))));
+  }
+}
+
+TEST(test_parse, else_has_empty_body) {
+  {
+    spy_visitor v;
+    padded_string code(u8"if (cond) { body; } else;\n{ }"_sv);
```
strager, line 603: Nit: I don't think it's worth having a diagnostic for this code. If there's a newline after the `else;`, I think the empty body is intentional.

strager (changes requested):
Thanks for the update. The logic looks much cleaner now!

test/test-parse-statement.cpp
someoneigna: Right now it detects `else;{}` (semicolon after else) and `else;\n{}` (empty body), the following in [parse.h:2534](https://github.com/quick-lint/quick-lint-js/pull/290/files#diff-660547dad34c381d590ccfcc6343d98eb7c4b235bec2dc324fecd1cb62f71662R2523) should change to only trigger **_else has empty body_** when it is not preceded by a newline:

``` diff 
-   if (token_after_else.type == token_type::semicolon &&
-          this->peek().type == token_type::left_curly) {
+   if (token_after_else.type == token_type::semicolon &&
+          this->peek().type == token_type::left_curly &&
+          !this->peek().has_leading_newline) {

-     if (this->peek().has_leading_newline) {
+     if (body is empty) {
              error_else_has_empty_body{.where = this->peek().span()});

       } else {
          this->error_reporter_->report(error_unexpected_semicolon_after_else{
              .semicolon = token_after_else.span()});
        }
    }
```

Should I omit the else has empty body warning altogether, or should I detect the empty body?
I do not know how to do it without making use of a transaction, advancing the peek and then doing a rollback afterwards.

strager: > Should I omit the else has empty body warning altogether

Yeah, let's only emit error_unexpected_semicolon_after_else in this PR.

> I do not know how to do it without making use of a transaction, advancing the peek and then doing a rollback afterwards.

Would something like this work? Or am I missing an edge case?
```c++
      if (token_after_else.type == token_type::semicolon &&
          this->peek().type == token_type::left_curly &&
          !this->peek().has_leading_newline) {
        this->error_reporter_->report(error_unexpected_semicolon_after_else{
          .semicolon = token_after_else.span(),
        });
      }
```

docs/errors/E204.md
```diff
@@ -0,0 +1,30 @@
+# E204: semicolon after else may be causing unexpected behavior.
```
strager, line 1: Someone else took error code E204. Choose a different error code.

Issue #280, opened 2021-05-10
5$: Span for E064 is too much


strager:
In the following code, quick-lint-js has a very big span for the E064 error:
```js
{
    if (isNaN(port))
//  ^^^^^^^^^^^^^^^^ E064 missing body for 'if' statement
}
```
Make the error point to after the ')' instead (i.e. where the body would be expected).

Perhaps include a note of where the `if` statement begins.

AidenThing:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.
Hey I would like to do this already started a little bit so a question:
do you want squiggles after the 'if(a)' statement if there is no '\n' after?
wouldn't it imply that the user is still typing. also, it still throws error if you call quick-lint-js exampltxt.mjs. just no squiggles

strager:
>  do you want squiggles after the 'if(a)' statement if there is no '\n' after?

If there's no newline in `{if(a)}`, I want quick-lint-js to report an empty span between the `)` and the `{`.

> wouldn't it imply that the user is still typing.

An error is an error.

I do want to tailor the error messaging to be as helpful as possible, which is the point of this task. I think the current behavior (highlighting the whole `if(condition)`) is noisy and unhelpful.

> also, it still throws error if you call quick-lint-js exampltxt.mjs. just no squiggles

The command line interface doesn't show squigglies right now. You'd need to use the Vim plugin or VS Code extension (or the [web demo](https://quick-lint-js.com/demo/), which is kind of jank right now) to see any squigglies.

AidenThing:
Hey, is the the kind of formatting you are looking for?
Would you want, in case d for it to not continue to the next line between the if and the else? just want it to stop on the line of the if?

Also, I do not know what to do about quick-lint calling error 150 instead on line 6. I think that that is a separate problem.
 
![quick-lint](https://user-images.githubusercontent.com/60998375/118177964-01998900-b3f9-11eb-9d1b-c68a158b09a3.PNG)

strager:
> Hey, is the the kind of formatting you are looking for?
>
> ![quick-lint](https://user-images.githubusercontent.com/60998375/118177964-01998900-b3f9-11eb-9d1b-c68a158b09a3.PNG)

* First **a** looks good assuming only line 2 has a squiggly (not like 3).
* Second **a** looks good.
* **b** looks good, but it'd be better for the squiggly to not cover the `}`.
* **c** looks good.
[...]

PR #296, opened 2021-05-14
Span for E064 is too much #280

AidenThing:
changes in parse.h and test-parse-statement.cpp


test/test-parse-statement.cpp
```diff
@@ -459,7 +459,7 @@ TEST(test_parse, if_without_body) {
     EXPECT_THAT(v.errors,
                 ElementsAre(ERROR_TYPE_FIELD(
                     error_missing_body_for_if_statement, if_and_condition,
-                    offsets_matcher(&code, 0, u8"if (a)"))));
+                    offsets_matcher(&code, 6, 7))));
```
strager, line 462: Almost-blocking: Use `strlen` and a string to define offset ranges, improving readability:
```suggestion
                    offsets_matcher(&code, strlen(u8"if (a)"), u8"\n"))));
```

strager, line 462: Why is the span covering the newline character?

src/quick-lint-js/parse.h
```diff
@@ -2506,9 +2506,10 @@ class parser {
     case token_type::end_of_file:
     case token_type::kw_else:
     case token_type::right_curly:
+      const char8* end_prev_token = this->lexer_.end_of_previous_token();
       this->error_reporter_->report(error_missing_body_for_if_statement{
-          .if_and_condition = source_code_span(
-              if_token_span.begin(), this->lexer_.end_of_previous_token()),
+          .if_and_condition = 
+              source_code_span(end_prev_token, end_prev_token + 1),
```
strager, line 2512: Blocking: After this change, `if_and_condition` is a misnomer. Rename it.

strager (changes requested):
I think a better solution would be to change how the diagnostic is printed in error.h. We can use `source_code_span(if_and_condition.end(), if_and_condition.end())` instead of using `if_and_condition`. (See error_missing_expression_between_parentheses for a similar example.) What do you think?

AidenThing:

src/quick-lint-js/error.h
```diff
@@ -572,9 +572,10 @@
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
       error_missing_body_for_if_statement, "E064",                             \
-      { source_code_span if_and_condition; },                                  \
+      { source_code_span after_if_condition; },                                \
       .error(QLJS_TRANSLATABLE("missing body for 'if' statement"),             \
-             if_and_condition))                                                \
+             source_code_span(after_if_condition.end(),                        \
+                              after_if_condition.end())))                      \
```
strager, line 578: I think it's weird to make `after_if_condition` an empty span (in parse.h), then unpack it and recreate the span here.

Either use `after_if_condition` as-is here, or leave `after_if_condition` as it was before in parse.h.

src/quick-lint-js/error.h
```diff
@@ -572,9 +572,10 @@
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
       error_missing_body_for_if_statement, "E064",                             \
-      { source_code_span if_and_condition; },                                  \
+      { source_code_span after_if_condition; },                                \
```
strager, line 575: Nit: I think `expected_body` is a better name.

test/test-parse-statement.cpp
```diff
@@ -472,8 +473,9 @@ TEST(test_parse, if_without_body) {
                                       "visit_exit_block_scope"));
     EXPECT_THAT(v.errors,
                 ElementsAre(ERROR_TYPE_FIELD(
-                    error_missing_body_for_if_statement, if_and_condition,
-                    offsets_matcher(&code, strlen(u8"{\n"), u8"if (a)"))));
+                    error_missing_body_for_if_statement, after_if_condition,
+                    offsets_matcher(&code, strlen(u8"{\nif (a)"),
+                                    strlen(u8"{\nif (a)")))));
```
strager, line 478: Instead of `strlen` for the second offset, you can write the span of characters. In this case, the span is empty:
```suggestion
                    offsets_matcher(&code, strlen(u8"{\nif (a)"),
                                   u8""))));
```

strager (approved):

src/quick-lint-js/error.h
AidenThing: Are you saying a change like this?  

` QLJS_ERROR_TYPE(                                                             
      error_missing_body_for_if_statement, "E064",                             
      { source_code_span expected_body; },                                     
      .error(QLJS_TRANSLATABLE("missing body for 'if' statement"),             
             expected_body))       
`

`const char8 *end_of_if_condition = this->lexer_.end_of_previous_token();
      this->error_reporter_->report(error_missing_body_for_if_statement{
          .expected_body =
              source_code_span(end_of_if_condition, end_of_if_condition),`

strager (approved):
Looks good!

AidenThing:
When ever works for you works for me!

PR #302, opened 2021-05-14
Improve error on 'async static' in class

### LINKED ISSUE #291: 10$: Improve error on 'async static' in class
quick-lint-js gives a crappy error message for the following code:
```js
class C {
  async static foo() {}
}
```
Improve the message. Something like: "'async static' is not allowed; write 'static async' instead"

erlliam:
Resolves #291.

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -67,6 +67,13 @@
              await_operator))                                                  \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_async_static_method, "E269", { source_code_span async_static; },   \
+      .error(                                                                  \
+          QLJS_TRANSLATABLE(                                                   \
+              "'async static' is not allowed; write 'static async' instead"),  \
+          async_static))                                                       \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_big_int_literal_contains_decimal_point, "E005",                    \
       { source_code_span where; },                                             \
       .error(QLJS_TRANSLATABLE("BigInt literal contains decimal point"),       \
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -1441,12 +1441,43 @@ class parser {
     QLJS_WARNING_IGNORE_GCC("-Wshadow-local")
 
     std::optional<identifier> last_ident;
+    const char8 *async_token_begin = nullptr;
     function_attributes method_attributes = function_attributes::normal;
+    bool async_static = false;
 
     auto parse_and_visit_field_or_method_impl =
-        [this, &v](std::optional<identifier> property_name,
-                   source_code_span property_name_span,
-                   function_attributes method_attributes) -> void {
+        [this, &v, &async_static, &async_token_begin](
+            std::optional<identifier> property_name,
+            source_code_span property_name_span,
+            function_attributes method_attributes) -> void {
+      if (async_static) {
+        if (this->peek().type == token_type::star) {
+          // async static *m() {}  // Invalid.
+          method_attributes = function_attributes::async_generator;
+          this->skip();
+        }
+        switch (this->peek().type) {
+        QLJS_CASE_RESERVED_KEYWORD_EXCEPT_FUNCTION:
+        QLJS_CASE_CONTEXTUAL_KEYWORD:
+        case token_type::identifier:
+        case token_type::private_identifier:
+        case token_type::reserved_keyword_with_escape_sequence:
+          // async static method() {}             // Invalid
+          // async static *myAsyncGenerator() {}  // Invalid
+          this->error_reporter_->report(error_async_static_method{
+              .async_static =
+                  source_code_span(async_token_begin, property_name_span.end()),
+          });
+          property_name = this->peek().identifier_name();
+          property_name_span = property_name->span();
+          this->skip();
+          break;
+        // async static() {}
+        default:
+          break;
+        }
+      }
+
       switch (this->peek().type) {
       // method() { }
       // method { }    // Invalid (missing parameter list).
@@ -1540,14 +1571,41 @@ class parser {
     // async f() {}
     case token_type::kw_async:
       last_ident = this->peek().identifier_name();
+      async_token_begin = this->peek().begin;
       this->skip();
-      if (this->peek().type != token_type::left_paren) {
-        method_attributes = function_attributes::async;
-      }
-      if (this->peek().type == token_type::star) {
-        // async *g() {}
-        method_attributes = function_attributes::async_generator;
-        this->skip();
+      if (this->peek().has_leading_newline) {
+        switch (this->peek().type) {
+        // 'async' is a field name:
+        // class {
+        //   async
+        //   method() {}
+        // }
+        QLJS_CASE_KEYWORD:
+        case token_type::left_square:
+        case token_type::number:
+        case token_type::string:
+        case token_type::identifier:
+        case token_type::private_identifier:
+        case token_type::star:
+          v.visit_property_declaration(last_ident);
+          return;
+        default:
+          break;
+        }
+      } else {
+        if (this->peek().type != token_type::left_paren) {
+          method_attributes = function_attributes::async;
+        }
+        if (this->peek().type == token_type::star) {
+          // async *g() {}
+          method_attributes = function_attributes::async_generator;
+          this->skip();
+        }
+        if (this->peek().type == token_type::kw_static) {
+          // async static method() {}  // Invalid
+          // async static() {}
+          async_static = true;
+        }
       }
       break;
 
@@ -1576,12 +1634,8 @@ class parser {
     // field = initialValue;
     // #field = initialValue;
     QLJS_CASE_RESERVED_KEYWORD_EXCEPT_FUNCTION:
-    QLJS_CASE_CONTEXTUAL_KEYWORD_EXCEPT_ASYNC_AND_GET_AND_SET_AND_STATIC:
+    QLJS_CASE_CONTEXTUAL_KEYWORD:
     case token_type::identifier:
-    case token_type::kw_async:
-    case token_type::kw_get:
-    case token_type::kw_set:
-    case token_type::kw_static:
     case token_type::private_identifier:
     case token_type::reserved_keyword_with_escape_sequence: {
       identifier property_name = this->peek().identifier_name();
--- a/test/test-parse-class.cpp
+++ b/test/test-parse-class.cpp
@@ -1078,6 +1078,110 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
             offsets_matcher(&code, strlen(u8"with (obj) "), u8"class"))));
   }
 }
+
+TEST(test_parse, async_static_method_is_disallowed) {
+  {
+    spy_visitor v;
+    padded_string code(
+        u8"class C { async static m() { await myPromise; } }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"myPromise"}));
+
+    EXPECT_THAT(v.visits,
+                ElementsAre("visit_variable_declaration",       // C
+                            "visit_enter_class_scope",          //
+                            "visit_property_declaration",       // m
+                            "visit_enter_function_scope",       // m
+                            "visit_enter_function_scope_body",  // m
+                            "visit_variable_use",               // myPromise
+                            "visit_exit_function_scope",        // m
+                            "visit_exit_class_scope"));
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_async_static_method, async_static,
+            offsets_matcher(&code, strlen(u8"class C { "), u8"async static"))));
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"class C { async static static() { } }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_EQ(v.property_declarations[0].name, u8"static");
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_async_static_method, async_static,
+            offsets_matcher(&code, strlen(u8"class C { "), u8"async static"))));
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"class C { async static *m() { } }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_async_static_method, async_static,
+            offsets_matcher(&code, strlen(u8"class C { "), u8"async static"))));
+  }
+}
+
+TEST(test_parse, static_method_allows_newline_after_static_keyword) {
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"class C { static\n m() { } }"_sv);
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+  }
+
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"class C { static\n *m() { } }"_sv);
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+  }
+
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"class C { static\n async *m() { } }"_sv);
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+  }
+
+  {
+    spy_visitor v =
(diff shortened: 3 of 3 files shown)


erlliam:
Ignore my request for review.

erlliam:
It should be good to review now.

Locally the string comparison works just fine, I'm not sure what the issue is (I thought it was lack of `_sv`).

My current solution is sort of hacky... I'm sure you'd prefer the `source_code_span` to point to `async static` and not the identifier. Besides that, this solution gets the job done without having to store extra information or state.

erlliam:
`property_name->normalized_name() == u8"static"`
I realize this will not work. It will incorrectly assume `\u0073tatic` a keyword...

erlliam:
The third attempt is still silly. What I'm thinking:

* See async
* Store the start of the token
* Skip
* See static
* Set a boolean that says potentially_async_static
* Store the end of the token
* Once we are inside `parse_and_visit_field_or_method_impl`, if static isn't a method, report the error

The main issue I'd deal with the lambda captures. Should I just add another parameter to the lambdas? Perhaps a struct? No clue.

src/quick-lint-js/parse.h
```diff
@@ -1549,6 +1555,9 @@ class parser {
         method_attributes = function_attributes::async_generator;
         this->skip();
       }
+      if (this->peek().type == token_type::kw_static) {
+        async_static = true;
```
strager, line 1559: I'd report the error here, and highlight the `async static` keywords instead of the method name. I think that'd simplify the logic and also be a better user experience.

src/quick-lint-js/error.h
```diff
@@ -1139,6 +1139,14 @@
       { source_code_span continue_statement; },                                \
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
+      error_async_static_method, "E202",                                       \
```
strager, line 1144: Nit: Sort error types alphabetically by struct name.

src/quick-lint-js/parse.h
erlliam: That's what I tried originally, but then it triggers a false positive for an async method named static. (`async static() {}`)

Perhaps I should use the transaction stuff and just check for a left paren right then and there?

strager: > That's what I tried originally, but then it triggers a false positive for an async method named static. (`async static() {}`)

I realized that soon after I made my comment. =X

src/quick-lint-js/parse.h
```diff
@@ -1482,6 +1484,10 @@ class parser {
           //   method() {}
           // }
           v.visit_property_declaration(property_name);
+        } else if (async_static) {
+          this->error_reporter_->report(error_async_static_method{
```
strager, line 1488: Document with a comment:
```suggestion
        } else if (async_static) {
          // async static method() {}             // Invalid.
          // async static *myAsyncGenerator() {}  // Invalid.
          this->error_reporter_->report(error_async_static_method{
```

src/quick-lint-js/parse.h
strager, line 1559: Document with a comment:
```suggestion
      if (this->peek().type == token_type::kw_static) {
        // async static method() {}  // Invalid.
        // async static() {}
        async_static = true;
```

test/test-parse-class.cpp
```diff
@@ -212,6 +212,18 @@ TEST(test_parse, class_statement_with_methods) {
                 ElementsAre(spy_visitor::visited_property_declaration{u8"m"}));
   }
 
+  {
```
strager, line 215: I'd put this test into its own `TEST`. All other tests in this `TEST` assert no error.

strager (approved):

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,20 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
             offsets_matcher(&code, strlen(u8"with (obj) "), u8"class"))));
   }
 }
+
+TEST(test_parse, async_static_method_is_disallowed) {
+  {
+    spy_visitor v;
+    padded_string code(u8"class C { async static m() {} }"_sv);
```
strager, line 1085: Verify that error recovery works by checking visits. In particular, a class property called `m` should be declared, and its body should be visited, and `await` should be treated as an operator inside the body.

strager (commented):
Looks good, but I'm skeptical that it's parsing correctly. I'd expect lines 1458-1460 (or similar) to execute. Adding to the tests should give me confidence.

erlliam:
Actually, this won't work. What I would like to do is use goto next (the original next, not the one I added), but I don't know how to do that. I guess the original next doesn't exist inside of the lambda. I'm fighting C++.

strager (commented):
Idea to avoid `goto`:

```diff
diff --git a/src/quick-lint-js/parse.h b/src/quick-lint-js/parse.h
index 32a64bfe..14ead941 100644
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -1450,7 +1450,27 @@ class parser {
             std::optional<identifier> property_name,
             source_code_span property_name_span,
             function_attributes method_attributes) -> void {
-      next:
+        if (async_static && !this->peek().has_leading_newline) {
+          switch (this->peek().type) {
+          case token_type::identifier:
+          case token_type::private_identifier:
+          case token_type::star:
+            // async static method() {}             // Invalid
+            // async static *myAsyncGenerator() {}  // Invalid
+            this->error_reporter_->report(error_async_static_method{
+                .async_static = source_code_span(async_token_begin,
+                                                 property_name_span.end()),
+            });
+            property_name = this->peek().identifier_name();
+            property_name_span = property_name->span();
+            this->skip();
+            break;
+
+          default:
+            break;
+          }
+        }
+
         switch (this->peek().type) {
         // method() { }
         // method { }    // Invalid (missing parameter list).
@@ -1486,17 +1506,6 @@ class parser {
             //   method() {}
             // }
             v.visit_property_declaration(property_name);
-          } else if (async_static) {
-            // async static method() {}             // Invalid
-            // async static *myAsyncGenerator() {}  // Invalid
-            this->error_reporter_->report(error_async_static_method{
-                .async_static = source_code_span(async_token_begin,
-                                                 property_name_span.end()),
-            });
-            property_name = this->peek().identifier_name();
-            property_name_span = property_name->span();
-            this->skip();
-            goto next;
           } else {
             this->error_reporter_->report(error_unexpected_token{
                 .token = property_name_span,
```

erlliam:
> Actually, this won't work. What I would like to do is use goto next (the original next, not the one I added), but I don't know how to do that. I guess the original next doesn't exist inside of the lambda. I'm fighting C++.

It turns out what I wanted was just to use the cases from the switch statement where we enter the `parse_and_visit_field_or_method`. No goto needed (especially after your suggestion).

Something sketchy:
I've refactored the cases used (where we enter `parse_and_visit_field...`), perhaps that should land as it's own commit?
Also the test at the could just use parse_and_visit_statement instead of manually creating the spy visitor, etc.

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,44 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
...
+            offsets_matcher(&code, strlen(u8"class C { "), u8"async static"))));
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"class C { async static static() { } }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_EQ(v.property_declarations[0].name, u8"static");
+  }
```
strager, line 1117: Should we be checking for `error_async_static_method` here?

src/quick-lint-js/parse.h
```diff
@@ -1549,6 +1576,11 @@ class parser {
         method_attributes = function_attributes::async_generator;
         this->skip();
       }
+      if (this->peek().type == token_type::kw_static) {
+        // async static method() {}  // Invalid
+        // async static() {}
```
strager, line 1581: Is `class { async static() {} }` already tested by an existing test? Or should we add a test for it?

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,44 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
...
+}
```
strager, line 1118: What about these:
```js
class C {
  async    // V8 treats this as a field declaration, I think.
  static   // V8 treats this as part of the method 'm'.
  m() { await(myPromise); }  // V8 makes a non-async static method called 'm'.
}
```
```js
class C {
  async  // V8 treats this as a field declaration, I think.
  static m() { }  // V8 makes a non-async static method called 'm'.
}
```

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,38 @@ class parser {
...
+        case token_type::private_identifier:
+        case token_type::reserved_keyword_with_escape_sequence:
+        case token_type::star:
+          // async static method() {}             // Invalid
+          // async static *myAsyncGenerator() {}  // Invalid
+          this->error_reporter_->report(error_async_static_method{
+              .async_static =
+                  source_code_span(async_token_begin, property_name_span.end()),
+          });
+          property_name = this->peek().identifier_name();
```
strager, line 1467: Blocking: This is invalid for `token_type::star`.

strager (changes requested):
Looks good except for `async static *m() {}`. (Write a test!)

src/quick-lint-js/parse.h
erlliam: I believe this test `class_statement_with_keyword_property` handles this.

src/quick-lint-js/parse.h
```diff
@@ -1540,14 +1569,36 @@ class parser {
...
-        this->skip();
+      if (this->peek().has_leading_newline) {
+        switch (this->peek().type) {
+        QLJS_CASE_KEYWORD:
+        case token_type::left_square:
+        case token_type::number:
+        case token_type::string:
+        case token_type::identifier:
+        case token_type::private_identifier:
+        case token_type::star:
```
erlliam, line 1582: This seems wrong (star case)

src/quick-lint-js/parse.h
```diff
@@ -1576,12 +1627,8 @@ class parser {
     // field = initialValue;
     // #field = initialValue;
     QLJS_CASE_RESERVED_KEYWORD_EXCEPT_FUNCTION:
-    QLJS_CASE_CONTEXTUAL_KEYWORD_EXCEPT_ASYNC_AND_GET_AND_SET_AND_STATIC:
+    QLJS_CASE_CONTEXTUAL_KEYWORD:
```
erlliam, line 1630: This change is probably out of scope.

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,91 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
...
+    EXPECT_THAT(v.variable_uses,
+                ElementsAre(spy_visitor::visited_variable_use{u8"myPromise"}));
+
+    EXPECT_THAT(v.visits,
+                ElementsAre("visit_variable_declaration",       // C
+                            "visit_enter_class_scope",          //
+                            "visit_property_declaration",       // m
+                            "visit_enter_function_scope",       // m
+                            "visit_enter_function_scope_body",  // m
+                            "visit_variable_use",         // await myPromise
```
strager, line 1100: Nit: Comment should say just `myPromise`:
```suggestion
                            "visit_variable_use",         // myPromise
```

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,91 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
...
+  }
+}
+
+TEST(test_parse, static_method_allows_newline_after_static_keyword) {
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"class C { static\n m() { } }"_sv);
+    EXPECT_EQ(v.property_declarations[0].name, u8"m");
+  }
+}
```
strager, line 1144: Nit: Are `static async m(){}` or `static *m(){}` parsed differently enough to warrant tests?

test/test-parse-class.cpp
```diff
@@ -1078,6 +1078,91 @@ TEST(test_parse, class_statement_as_with_statement_body_is_disallowed) {
...
+    EXPECT_EQ(v.property_declarations[0].name, u8"async");
+    EXPECT_EQ(v.property_declarations[1].name, u8"m");
+  }
+
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"class C { async\n static m() { } }"_sv);
+    EXPECT_EQ(v.property_declarations[0].name, u8"async");
+    EXPECT_EQ(v.property_declarations[1].name, u8"m");
+  }
```
strager, line 1159: Love it!

src/quick-lint-js/parse.h
```diff
@@ -1540,14 +1569,36 @@ class parser {
     // async f() {}
     case token_type::kw_async:
       last_ident = this->peek().identifier_name();
+      async_token_begin = this->peek().span().begin();
```
strager, line 1572: Nit:
```suggestion
      async_token_begin = this->peek().begin;
```

src/quick-lint-js/parse.h
```diff
@@ -1540,14 +1569,36 @@ class parser {
...
+        switch (this->peek().type) {
+        QLJS_CASE_KEYWORD:
+        case token_type::left_square:
+        case token_type::number:
+        case token_type::string:
+        case token_type::identifier:
+        case token_type::private_identifier:
+        case token_type::star:
+          v.visit_property_declaration(last_ident);
+          return;
```
strager, line 1584: Add a comment:
```suggestion
        switch (this->peek().type) {
        // 'async' is a field name:
        // class {
        //   async
        //   method() {}
        // }
        QLJS_CASE_KEYWORD:
        case token_type::left_square:
        case token_type::number:
        case token_type::string:
        case token_type::identifier:
        case token_type::private_identifier:
        case token_type::star:
          v.visit_property_declaration(last_ident);
          return;
```

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,41 @@ class parser {
...
 
     auto parse_and_visit_field_or_method_impl =
-        [this, &v](std::optional<identifier> property_name,
-                   source_code_span property_name_span,
-                   function_attributes method_attributes) -> void {
+        [this, &v, &async_static, &async_token_begin](
+            std::optional<identifier> property_name,
+            source_code_span property_name_span,
+            function_attributes method_attributes) -> void {
+      if (async_static && !this->peek().has_leading_newline) {
```
strager, line 1453: What is the purpose of `!this->peek().has_leading_newline`? It looks like `!this->peek().has_leading_newline` would cause the following examples to parse differently. Am I correct?
```js
class C {
  async static m() {}
}
```
```js
class C {
  async static m
    () {}
}
```

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,41 @@ class parser {
...
+        switch (this->peek().type) {
+        case token_type::star:
+          method_attributes = function_attributes::generator;
+          this->skip();
+          [[fallthrough]];
```
strager, line 1458: Blocking: I think fallthrough after skipping is dangerous. What if `*` is followed by `+`, for example? I think that'd cause a crash in `identifier_name()` on line 1470.

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,41 @@ class parser {
...
-                   source_code_span property_name_span,
-                   function_attributes method_attributes) -> void {
+        [this, &v, &async_static, &async_token_begin](
+            std::optional<identifier> property_name,
+            source_code_span property_name_span,
+            function_attributes method_attributes) -> void {
+      if (async_static && !this->peek().has_leading_newline) {
+        switch (this->peek().type) {
+        case token_type::star:
+          method_attributes = function_attributes::generator;
```
strager, line 1456: Shouldn't this be an async generator?

strager (changes requested):

src/quick-lint-js/parse.h
erlliam: I'm not sure what the purpose of this new line check is anymore.

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,41 @@ class parser {
...
+          // async static *myAsyncGenerator() {}  // Invalid
+          this->error_reporter_->report(error_async_static_method{
+              .async_static =
+                  source_code_span(async_token_begin, property_name_span.end()),
+          });
+          property_name = this->peek().identifier_name();
+          property_name_span = property_name->span();
+          this->skip();
+          break;
+        default:
```
strager, line 1474: Maybe this deserves a comment?
```suggestion
        // async static() {}
        default:
```

src/quick-lint-js/parse.h
```diff
@@ -1441,12 +1441,41 @@ class parser {
...
-        [this, &v](std::optional<identifier> property_name,
-                   source_code_span property_name_span,
-                   function_attributes method_attributes) -> void {
+        [this, &v, &async_static, &async_token_begin](
+            std::optional<identifier> property_name,
+            source_code_span property_name_span,
+            function_attributes method_attributes) -> void {
+      if (async_static) {
+        if (this->peek().type == token_type::star) {
+          method_attributes = function_attributes::async_generator;
```
strager, line 1455: Nit: Comment?
```suggestion
          // async static *m() {}  // Invalid.
          method_attributes = function_attributes::async_generator;
```

strager (approved):

Issue #96, opened 2020-12-21
26$: Parse {async=3, get=4, set=5} variable declarations


strager:
The following code crashes when given to quick-lint-js:
```js
let {async=3, get=4, set=5} = {};
```
Error:
```
../src/parse.cpp:1299: fatal: token not implemented in parse_object_literal: equal on line 1 column 11
zsh: illegal hardware instruction  ./build/quick-lint-js hello.js
```

strager:
I think we should rewrite the binding parser to not parse an expression.

strager:
> I think we should rewrite the binding parser to not parse an expression.

I think this might be reused by the function parameter parser. Hopefully they have similar or identical rules.

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
> I think we should rewrite the binding parser to not parse an expression.

Actually, this is a bad idea. How would we parse the following code otherwise:
```js
let f = ({k = true} = {}) => {};
```

PR #308, opened 2021-05-15
Fix crash on destructuring assignment with async

singalhimanshu:
The following code doesn't make quick-lint-js to crash now:

    let { async = 3, get = 4, set = 5 } = {};

Fix issue: #96

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -1381,6 +1381,32 @@ expression* parser::parse_object_literal() {
     this->skip();
     return property_name;
   };
+  auto parse_equal = [&](token key_token, expression* key) {
+    expression* lhs;
+    bool missing_key;
+    switch (key_token.type) {
+    case token_type::number:
+    case token_type::string:
+      lhs = this->make_expression<expression::literal>(key_token.span());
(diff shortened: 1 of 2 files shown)


src/parse.cpp
```diff
@@ -1771,6 +1772,36 @@ expression* parser::parse_object_literal() {
         break;
       }
 
+      case token_type::equal: {
```
strager, line 1775: Blocking: This code looks like a copy-paste from line 1564. This is a lot of duplicated code. Let's try to avoid the copy-paste.

strager (changes requested):
Blocking: Add a test case. Perhaps add it to `variables_can_be_named_contextual_keywords` so many variable names are tested.

singalhimanshu:
I have made the requested changes. Let me know if there are more changes that have to be made.

strager (approved):

strager:
Sweet!

Shall I merge now?

singalhimanshu:
> Sweet!
> 
> Shall I merge now?

Yes sure.

Issue #271, opened 2021-05-07
Website: "BOM" looks ugly in light mode


strager:
https://quick-lint-js.com/errors/#E095

JanPoonthong:
I'm working now it

PR #309, opened 2021-05-16
Fix issue #271: "BOM" looking ugly on light mode

JanPoonthong:
Change the background-color to different color.

### DIFF
--- a/website/public/errors/index.ejs.html
+++ b/website/public/errors/index.ejs.html
@@ -47,23 +47,17 @@
         text-decoration: none;
       }
 
-      @media (prefers-color-scheme: dark) {
-        .unicode-bom:before {
-            content: "BOM";
-            border: 1px solid red;
-            background-color: #884444;
-            border-radius: 5px;
-            padding: 1px 3px;
-          }
+      .unicode-bom:before {
+        content: "BOM";
+        background-color: #edbf66;
+        border: 1px solid red;
+        border-radius: 5px;
+        padding: 1px 3px;
       }
 
-      @media (prefers-color-scheme: light) {
+      @media (prefers-color-scheme: dark) {
         .unicode-bom:before {
-          content: "BOM";
-          border: 1px solid red;
-          background-color: #edbf66;
-          border-radius: 5px;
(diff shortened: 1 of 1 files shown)


JanPoonthong:
I have read and hereby agree to quick-lint-js

JanPoonthong:

strager (changes requested):
This PR fixes light mode, but it makes dark mode worse (low contrast):
Screenshot from 2021-05-15 19-00-50
Screenshot from 2021-05-15 19-00-59

For reference, here's what dark mode used to look like (high contrast):
Screenshot from 2021-05-15 19-02-29

I think we need to pick different background colors depending on whether light mode or dark mode is active.

website/public/errors/index.ejs.html
```diff
@@ -47,12 +47,24 @@
...
-        background-color: #884444;
-        border-radius: 5px;
-        padding: 1px 3px;
+      @media (prefers-color-scheme: dark) {
+        .unicode-bom:before {
+            content: "BOM";
+            border: 1px solid red;
+            background-color: #884444;
+            border-radius: 5px;
+            padding: 1px 3px;
```
strager, line 56: Blocking: Most of these properties are duplicated. Factor the duplication.

website/public/errors/index.ejs.html
```diff
@@ -47,12 +47,24 @@
...
+
+      @media (prefers-color-scheme: light) {
+        .unicode-bom:before {
+          content: "BOM";
+          border: 1px solid red;
+          background-color: #edbf66;
+          border-radius: 5px;
+          padding: 1px 3px;
+        }
       }
```
strager, line 68: Assume that light mode is the default, in case `prefers-color-scheme` is not implemented:
```suggestion
        .unicode-bom:before {
          content: "BOM";
          border: 1px solid red;
          background-color: #edbf66;
          border-radius: 5px;
          padding: 1px 3px;
        }
```

strager (changes requested):
Site looks good. Code needs some work.

strager (approved):

Issue #306, opened 2021-05-15
12$: wrong precedence: 'undefined' === typeof foo ? 3 : 4


strager:
The following code has no errors, but quick-lint-js reports E057:
```js
 'undefined' === typeof foo ? 3 : 4;
```
It looks like precedence is the issue. quick-lint-js is happy with the following code:
```js
 'undefined' === (typeof foo) ? 3 : 4;
```

---

Possibly related: The following emits E020 but shouldn't:
```js
typeof true ? true : b = a
```

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

singalhimanshu:
I think the issue is coming for here - https://github.com/quick-lint/quick-lint-js/blob/2ec61fbf9f28f8df7503d14e7fce81a0b58baac7/src/parse.cpp#L220-L223

```js
'undefined' === typeof foo ? 3 : 4; // use of undeclared variable: foo [E057]
//                     ^^^^^^^^^^^ this is considered as an expression_kind::conditional operator.
```
Whereas the expected behavior is that `foo` is considered as an `expression_kind::variable`, but currently it is `expression_kind::conditional` which later causes problem here - https://github.com/quick-lint/quick-lint-js/blob/385b32e0c373859805b30be2620a4a958ad00e3e/src/quick-lint-js/parse.h#L744-L748 
A possible solution can be making `binary_operators=false` here - https://github.com/quick-lint/quick-lint-js/blob/2ec61fbf9f28f8df7503d14e7fce81a0b58baac7/src/parse.cpp#L220-L223 but that won't fix it because a code like this will be interpreted incorrectly -
```js
typeof foo.bar; 
```

strager:
> A possible solution can be making `binary_operators=false` here -
> 
> https://github.com/quick-lint/quick-lint-js/blob/2ec61fbf9f28f8df7503d14e7fce81a0b58baac7/src/parse.cpp#L220-L223
> but that won't fix it because a code like this will be interpreted incorrectly -
> 
> ```js
> typeof foo.bar; 
> ```

Sounds like we need a new precedence bool then.

(Our precedence parsing is crap. I'd welcome something better. ;P)

singalhimanshu:
> Sounds like we need a new precedence bool then.

I don't understand how precedence will help here. Can you explain with this example? 
```js
'undefined' === typeof foo ? 3 : 4;
```

[...]

PR #310, opened 2021-05-16
Fix conditional operator after typeof

singalhimanshu:
The following code should not produce a warning (E057):
     'undefined' === typeof foo ? 3 : 4;

Fix: #306

### DIFF
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -741,9 +741,18 @@ class parser {
     }
     case expression_kind::_typeof: {
       expression *child = ast->child_0();
-      if (child->kind() == expression_kind::variable) {
+      switch (child->kind()) {
+      case expression_kind::conditional: {
+        if (child->child_0()->kind() == expression_kind::variable) {
+          v.visit_variable_typeof_use(child->child_0()->variable_identifier());
+        }
+        break;
+      }
+      case expression_kind::variable: {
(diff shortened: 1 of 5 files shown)


singalhimanshu:
Let me know if this fix was okay. If so I will write the required test cases for this condition.

src/quick-lint-js/parse.h
```diff
@@ -741,9 +741,18 @@ class parser {
     }
     case expression_kind::_typeof: {
       expression *child = ast->child_0();
-      if (child->kind() == expression_kind::variable) {
+      switch (child->kind()) {
+      case expression_kind::conditional: {
+        if (child->child_0()->kind() == expression_kind::variable) {
```
strager, line 746: Blocking: This causes false negatives for other cases, like `"undefined" === typeof foo.x ? 3 : 4;`.

strager (changes requested):
Blocking: I think editing `visit_expression` is the wrong approach. We should fix the parsing.

> If so I will write the required test cases for this condition.

We'd want the test cases regardless of what the solution is, no?

src/quick-lint-js/parse.h
singalhimanshu: > We'd want the test cases regardless of what the solution is, no?

Yes

singalhimanshu:
> We'd want the test cases regardless of what the solution is, no?

Yes =S

strager (approved):

Issue #279, opened 2021-05-10
8$: Error on await () => {}


strager:
Within async functions, `await () => {}` is invalid syntax, at least according to Node.js v15.5.1:
```javascript
async function g() {
  f(await () => {});
}
g();
```
```
hello.js:2
  f(await () => {});
    ^^^^^^^

SyntaxError: Malformed arrow function parameter list
```

quick-lint-js reports no error for this code. Assuming that Node.js is standards-conforming, quick-lint-js should report an error for an attempted await for an arrow function.

strager:
#278 is a related issue. #278 is about `async()=>{}` outside async functions. This task is about `async()=>{}` inside async functions. Perhaps the end result should be the same (i.e. same diagnostic and error recovery), but the parsing code seems to be different, so I thought they deserved separate tasks.

keyehzy:
Will try this one, PR'ing soon.

Issue #278, opened 2021-05-10
7$: Improve error message for 'await () => {}'


strager:
Often, I accidentally type 'await' instead of 'async' for my async arrow functions:
```javascript
function g() {
  f(await () => {
    await x;
  });
}
```
quick-lint-js reports errors for both `await(` and `await x`:
```
hello.js:3:5: error: 'await' is only allowed in async functions [E162]
hello.js:2:5: error: missing operator between expression and arrow function [E063]
hello.js:2:5: warning: use of undeclared variable: await [E057]
```
quick-lint-js should infer that `async` was intended, and provide a better error message (see below) and better error recovery (by assuming that the arrow function is `async`):
```
hello.js:2:5: error: 'await' does not create an async arrow function; use 'async' instead [E???]
```

Coldtrain:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
This seems to be implemented already as E0178.

PR #314, opened 2021-05-17
Suggest `async` when `await` is followed by arrow function.

keyehzy:
Attempts to fix issues #279 and #278.

In both cases we can check if `await (` is followed by an arrow function, suggest changing it to `async` and then rollback. 

For the expression beginning with `await (` inside an `async` function the expression following it is parsed twice, this happens because we have no way to know beforehand if we want that expression to be parsed and `await` might be an identifier or something else.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -683,37 +683,11 @@ expression* parser::parse_await_expression(token await_token, precedence prec) {
         [[fallthrough]];
       case token_type::complete_template:
       case token_type::incomplete_template:
+      case token_type::left_paren:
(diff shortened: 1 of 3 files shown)


keyehzy:
Should be good for review.

strager (changes requested):
See comments marked 'blocking'.

I think we should punt on fixing #278 in this PR; I think we should focus on #279. The two issues are related, but I don't think we can kill two birds with one stone.

strager:
@keyehzy reposted this as #488.

PR #315, opened 2021-05-18
Add how error message would works

matttheus:
I added a file called website/poc/index.html, so you can see how would work

I don't know is not working in the real website/demo, but I added there too.

don't have style yet.


strager (changes requested):
Feedback given in person.

matttheus:

matttheus:
I continued this work at #323

Issue #266, opened 2021-05-05
9$: Warn on side-effect-free expression in `else`


strager:
The following code snippets likely have a mistake:
```javascript
if (x) {
  f();
} else (y)  // missing 'if' keyword
{
  g();
}
```

Produce a warning in this case, suggesting `if`.

Abdelazizmotiaa2021:
if (x) {
   f();
}else{ //No (y) else means if non x
}

Abdelazizmotiaa2021:
else doesn't need a condition

keyehzy:
I have a PR for this issue.

PR #317, opened 2021-05-18
Attempts to fix issue #266

keyehzy:
We do this by treating `else (` separately. If `else` is followed by
a (conditional + block) pattern, maybe the intention was to use an
`else if` instead and a warning is emitted.

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -215,6 +215,13 @@
              else_token))                                                      \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_else_with_conditional_missing_if, "E???",                          \
+      { source_code_span else_token; },                                        \
+      .warning(QLJS_TRANSLATABLE("'else' with condition followed by block; "   \
+                                 "maybe 'else if' was intended"),              \
+               else_token))                                                    \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_escaped_character_disallowed_in_identifiers, "E012",               \
       { source_code_span escape_sequence; },                                   \
       .error(QLJS_TRANSLATABLE(                                                \
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -2515,8 +2515,26 @@ class parser {
     }
 
     if (this->peek().type == token_type::kw_else) {
+      source_code_span else_span = this->peek().span();
       this->skip();
-      parse_and_visit_body();
+
+      switch (this->peek().type) {
+      default:
+        parse_and_visit_body();
+        break;
+
+      case token_type::left_paren:
+        expression *ast = this->parse_expression(precedence{});
+        this->visit_expression(ast, v, variable_context::rhs);
+
+        if (this->peek().type == token_type::left_curly) {
+          this->error_reporter_->report(error_else_with_conditional_missing_if{
+              .else_token = else_span,
+          });
+          parse_and_visit_body();
+        }
+        break;
+      }
     }
   }
 
--- a/test/test-parse-statement.cpp
+++ b/test/test-parse-statement.cpp
@@ -578,6 +578,42 @@ TEST(test_parse, else_without_if) {
   }
 }
 
+TEST(test_parse, else_if) {
+  {
+    spy_visitor v =
+        parse_and_visit_statement(u8"if (a) { b; } else if (c) { d; }"_sv);
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",        // a
+                                      "visit_enter_block_scope",   // <if>
+                                      "visit_variable_use",        // b
+                                      "visit_exit_block_scope",    // </if>
+                                      "visit_variable_use",        // c
+                                      "visit_enter_block_scope",   // <else>
+                                      "visit_variable_use",        // d
+                                      "visit_exit_block_scope"));  // </else>
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"if (a) { b; } else (c) { d; }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_use",        // a
+                                      "visit_enter_block_scope",   // <if>
+                                      "visit_variable_use",        // b
+                                      "visit_exit_block_scope",    // </if>
+                                      "visit_variable_use",        // c
+                                      "visit_enter_block_scope",   // <block>
+                                      "visit_variable_use",        // d
+                                      "visit_exit_block_scope"));  // </block>
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_else_with_conditional_missing_if, else_token,
+            offsets_matcher(&code, strlen(u8"if (a) { b; } "), u8"else"))));
+  }
+}
+
 TEST(test_parse, block_statement) {
   {
     spy_visitor v = parse_and_visit_statement(u8"{ }"_sv);


src/quick-lint-js/error.h
```diff
@@ -214,6 +214,13 @@
       .error(QLJS_TRANSLATABLE("'else' has no corresponding 'if'"),            \
              else_token))                                                      \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_else_with_conditional_missing_if, "E???",                          \
```
strager, line 218: Blocking: Pick a valid error code. The `???` should be three decimal digits. (I should write a test to ensure error codes are well-formed.)

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,42 @@ TEST(test_parse, else_without_if) {
...
+                                      "visit_variable_use",        // d
+                                      "visit_exit_block_scope"));  // </block>
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_else_with_conditional_missing_if, else_token,
+            offsets_matcher(&code, strlen(u8"if (a) { b; } "), u8"else"))));
+  }
+}
```
strager, line 615: Blocking: In the following examples, we should *not* emit a diagnostic, but I think your implementation does produce a false-positive diagnostic:
```js
let a = false;
if (a) {
} else (() => console.log("b"))()  // inserting 'if' would be a syntax error
{ console.log("c"); }
```
```js
let a = false;
let o = {};
if (a) {
} else () => {}  // inserting 'if' would be a syntax error
{ console.log("c"); }
```
```js
let a = false;
let o = someExternalVariable;
if (a) {
} else (o.m())  // body of 'else' has potential side effects; inserting 'if' might be unintended
{ console.log("c"); }
```

src/quick-lint-js/parse.h
```diff
@@ -2515,8 +2515,26 @@ class parser {
...
+      switch (this->peek().type) {
+      default:
+        parse_and_visit_body();
+        break;
+
+      case token_type::left_paren:
+        expression *ast = this->parse_expression(precedence{});
+        this->visit_expression(ast, v, variable_context::rhs);
+
+        if (this->peek().type == token_type::left_curly) {
```
strager, line 2530: I think we also need to check the kind of `ast`. See examples in my 'blocking' comment.

In #266, I wrote "side-effect-free expression". I think this is conservative but probably still helpful. "side-effect-free" would be any combination of:
* `binary_operator` (I think?) (if children are side-effect free)
* `cond`
* `_typeof`
* `_class`
* `template` (if children are side-effect free)
* `array` (if children are side-effect free)
* `object` (if children are side-effect free)
* `conditional` (if children are side-effect free)
* ...

In particular:
* `dot` is disallowed because property access can have arbitrary side effects

A `bool is_side_effect_free(expression*)` function would be a welcome addition!

src/quick-lint-js/parse.h
```diff
@@ -2515,8 +2515,26 @@ class parser {
...
+      source_code_span else_span = this->peek().span();
       this->skip();
-      parse_and_visit_body();
+
+      switch (this->peek().type) {
+      default:
+        parse_and_visit_body();
+        break;
+
+      case token_type::left_paren:
```
strager, line 2526: Hmm, this is simpler than I thought. I like your idea of special-casing expression statements beginning with `(`.

strager (changes requested):

src/quick-lint-js/parse.h
keyehzy: I'm not sure yet about all the base cases, see body of `bool is_side_effect_free(expression* ast)`.

src/quick-lint-js/error.h
keyehzy: Done.

src/quick-lint-js/parse.h
keyehzy: As for checking the kind of the expression, right now the only invalidating kinds are arrow functions. I'm trying to come up with other examples where it's not caught by the side-effect-free condition.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
     v.visit_exit_with_scope();
   }
 
+  bool is_side_effect_free(expression *ast) {
```
strager, line 2463: Nit: Maybe it'd be better to invert the meaning of this function and call it `has_potential_side_effects`? `!is_side_effect_free(ast)` sounds like a double negative to me.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+    switch (ast->kind()) {
+    case expression_kind::_class:
```
strager, line 2465: Blocking: Class expressions might have side effects. For example:
```js
function f() { console.log("yo"); }
let C = class { [f()]() {} };
```

We don't have enough information in the AST to know, so let's be conservative and assume that side effects might happen.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+    case expression_kind::_invalid:
+    case expression_kind::_typeof:
```
strager, line 2467: Blocking: `typeof` might have a side effect, depending on its argument. For example:
```js
typeof 42; // no side effect
typeof f(); // side effect
```

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+    case expression_kind::await:
+    case expression_kind::import:
+    case expression_kind::literal:
+    case expression_kind::new_target:
+    case expression_kind::private_variable:
+    case expression_kind::rw_unary_prefix:
+    case expression_kind::rw_unary_suffix:
+    case expression_kind::spread:
```
strager, line 2475: Blocking: `spread` might have side effects, because a spread would call an iterator's `next` function.

src/quick-lint-js/parse.h
strager, line 2463: I think this function could easily be tested. Let me know if you want some help writing tests. (I think the tests would go in test-parse.cpp or test-parse-expression.cpp.) I provided some non-obvious/tricky examples in some comments.

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,69 @@ TEST(test_parse, else_without_if) {
...
+    padded_string code(u8"if (a) { b; } else () => {} { c; }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(
+        u8"if (a) { b; } else (() => console.log(c))() { d; }"_sv);
```
strager, line 629: Blocking: This should report a missing-semicolon error.

Perhaps you mean to write this code?
```js
if (a) { b; } else (() => console.log(c))()
{ d; }
```

src/quick-lint-js/parse.h
```diff
@@ -2515,8 +2580,42 @@ class parser {
...
+        if (this->peek().type == token_type::left_curly) {
+          if (!is_invalidating_if && this->is_side_effect_free(ast)) {
+            this->error_reporter_->report(
+                error_else_with_conditional_missing_if{
+                    .else_token = else_span,
+                });
+          }
+          parse_and_visit_body();
+        }
+        break;
```
strager, line 2617: Blocking: After parsing the expression statement, the language says we need a semicolon. (A semicolon might be inserted by ASI if a newline or `}` is found instead of a semicolon.)

In other words, the following code should produce a syntax error (and quick-lint-js does report a syntax error before your commit):
```js
if (a) { b; } else (() => console.log(c))() { d; }
```

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,69 @@ TEST(test_parse, else_without_if) {
...
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_FIELD(
+            error_else_with_conditional_missing_if, else_token,
+            offsets_matcher(&code, strlen(u8"if (a) { b; } "), u8"else"))));
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"if (a) { b; } else () => {} { c; }"_sv);
```
strager, line 620: Blocking: This should report a missing-semicolon error.

Perhaps you mean to write this code?
```js
if (a) { b; } else () => {}
{ c; }
```

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,69 @@ TEST(test_parse, else_without_if) {
...
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"if (a) { b; } else (o.m()) { c; }"_sv);
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.errors, IsEmpty());
```
strager, line 640: Blocking: This should report a missing-semicolon error.

strager (changes requested):
Hmm, this issue is more complicated than I thought! Lots of edge cases.

src/parse.cpp
```diff
@@ -2065,6 +2065,73 @@ expression* parser::maybe_wrap_erroneous_arrow_function(
...
+    return this->has_potential_side_effects(ast->child_0()) ||
+           this->has_potential_side_effects(ast->child_1()) ||
+           this->has_potential_side_effects(ast->child_2());
+
+  case expression_kind::object: {
+    for (int i = 0; i < ast->object_entry_count(); i++) {
+      auto entry = ast->object_entry(i);
+      if (entry.property.has_value()) {
+        if (this->has_potential_side_effects(*entry.property) ||
+            this->has_potential_side_effects(entry.value))
```
strager, line 2125: Shouldn't we check `entry.value` regardless of `entry.property.has_value()`?

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
   }
 }
 
+TEST(test_parse, test_expression_for_potential_side_effects) {
+  {
+    test_parser p(u8"class {};"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
+  }
```
strager, line 3170: This really should be `false`, but I know we can't implement that right now. Perhaps the following examples are better, because they do have side effects:
```js
class { [f()]() { } }
```
or:
```js
class { static pi = calculatePi(); }
```

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
...
+
+  {
+    test_parser p(u8"await foo;"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
+  }
+
+  {
+    test_parser p(u8"import foo;"_sv);
```
strager, line 3194: Blocking: This has syntax errors. I think you meant this:
```js
import(foo);
```

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
...
+
+  {
+    test_parser p(u8"x++;"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
+  }
+
+  {
+    test_parser p(u8"...foo"_sv);
```
strager, line 3215: Blocking: This is a syntax error. Spread isn't allowed on its own. I think you meant one of the following:
```js
{...foo}
```
```js
[...foo]
```
```js
f(...foo)
```

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
...
+    test_parser p(u8"new foo;"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
+  }
+
+  {
+    // TODO(keyehzh): Need more specific tests as some binary operators might
+    // have side-effects (such as '+', '==', ...) and some don't
+    // (such as '===' and '!==')
```
strager, line 3311: :+1:

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
...
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
+  }
+
+  {
+    test_parser p(u8"{[f()]: true}"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
```
strager, line 3427: Nit: I think we can save on lines of code:
```suggestion
    EXPECT_TRUE(p.parser().has_potential_side_effects(ast));
```

test/test-parse-expression.cpp
```diff
@@ -3161,6 +3161,273 @@ TEST_F(test_parse_expression,
...
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_FALSE(side_effects);
+  }
+
+  {
+    test_parser p(u8"{key: f()}"_sv);
+    expression* ast = p.parse_expression();
+    bool side_effects = p.parser().has_potential_side_effects(ast);
+    EXPECT_TRUE(side_effects);
```
strager, line 3420: Nit: I think we can safe a line of code by making `has_potential_side_effects` `static`:
```suggestion
    expression* ast = this->parse_expression(u8"{key: f()}"_sv);
    bool side_effects = parser::has_potential_side_effects(ast);
    EXPECT_TRUE(side_effects);
```

strager (changes requested):
Blocking: `if (a) { b; } else (c) { d; }` is still an issue (false negative).

Other than that, this commit looks bueno!

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
     v.visit_exit_with_scope();
   }
 
+  bool is_side_effect_free(expression *ast) {
+    switch (ast->kind()) {
+    case expression_kind::_class:
+    case expression_kind::_invalid:
+    case expression_kind::_typeof:
+    case expression_kind::await:
+    case expression_kind::import:
```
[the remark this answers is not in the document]
keyehzy: Fixed.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+    case expression_kind::literal:
+    case expression_kind::new_target:
+    case expression_kind::private_variable:
+    case expression_kind::rw_unary_prefix:
+    case expression_kind::rw_unary_suffix:
+    case expression_kind::spread:
+    case expression_kind::super:
+    case expression_kind::unary_operator:
```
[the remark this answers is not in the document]
keyehzy: Fixed.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+    case expression_kind::variable:
+    case expression_kind::yield_many:
+    case expression_kind::yield_none:
+    case expression_kind::yield_one:
```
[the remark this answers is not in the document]
keyehzy: Fixed.

src/quick-lint-js/parse.h
```diff
@@ -2460,6 +2460,71 @@ class parser {
...
+      return true;
+
+    case expression_kind::dot:
+    case expression_kind::index:
+    case expression_kind::function:
+    case expression_kind::named_function:
+      return false;
+
+    case expression_kind::_new:
```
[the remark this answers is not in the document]
keyehzy: Fixed.

src/parse.cpp
```diff
@@ -2065,6 +2065,73 @@ expression* parser::maybe_wrap_erroneous_arrow_function(
...
+    return has_potential_side_effects(ast->child_0()) ||
+           has_potential_side_effects(ast->child_1()) ||
+           has_potential_side_effects(ast->child_2());
+
+  case expression_kind::object: {
+    for (int i = 0; i < ast->object_entry_count(); i++) {
+      auto entry = ast->object_entry(i);
+      if (entry.property.has_value()) {
+        if (has_potential_side_effects(*entry.property) ||
+            has_potential_side_effects(entry.value))
```
strager, line 2125: Shouldn't we check `entry.value` regardless of `entry.property.has_value()`?

src/parse.cpp
```diff
@@ -2065,6 +2065,73 @@ expression* parser::maybe_wrap_erroneous_arrow_function(
...
+          return true;
+      }
+    }
+    return false;
+  }
+  }
+  return false;
```
strager, line 2132: Nit:
```suggestion
  QLJS_UNREACHABLE();
```

src/quick-lint-js/parse.h
```diff
@@ -2515,8 +2517,43 @@ class parser {
...
+        default:
+          break;
+
+        case expression_kind::arrow_function_with_expression:
+        case expression_kind::arrow_function_with_statements:
+          is_invalidating_if = true;
+          break;
+        }
+
+        this->consume_semicolon();
```
strager, line 2544: Blocking: This looks like it's in the wrong spot. Won't skipping a semicolon here cause the following code to have the error_else_with_conditional_missing_if warning?
```js
if (a) { b; } else (c);
{ d; }
```

I think we need to consume a semicolon after the check on line 2546.

test/test-parse-statement.cpp
```diff
@@ -578,6 +578,135 @@ TEST(test_parse, else_without_if) {
...
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(
+            ERROR_TYPE_FIELD(
+                error_missing_semicolon_after_statement, where,
+                offsets_matcher(&code, strlen(u8"if (a) { b; } else (c)"),
+                                u8"")),
+            ERROR_TYPE_FIELD(
+                error_else_with_conditional_missing_if, else_token,
+                offsets_matcher(&code, strlen(u8"if (a) { b; } "), u8"else"))));
```
strager, line 619: :+1:

strager (changes requested):

test/test-parse-expression.cpp
keyehzy: This one is resolved.

test/test-parse-expression.cpp
keyehzy: This one is resolved.

src/quick-lint-js/parse.h
keyehzy: Agreed!

src/parse.cpp
keyehzy: Thanks!

src/parse.cpp
```diff
@@ -2121,15 +2121,14 @@ bool parser::has_potential_side_effects(expression* ast) {
     for (int i = 0; i < ast->object_entry_count(); i++) {
       auto entry = ast->object_entry(i);
       if (entry.property.has_value()) {
-        if (has_potential_side_effects(*entry.property) ||
-            has_potential_side_effects(entry.value))
-          return true;
+        return has_potential_side_effects(*entry.property) ||
+               has_potential_side_effects(entry.value);
```
strager, line 2125: Blocking: You're only checking the first entry which has a value! We should check all entries.

src/quick-lint-js/parse.h
strager: Please add a test for this example.

strager:
There are merge conflicts because `parser::parse_and_visit_if`'s body has been moved to `src/quick-lint-js/parse-statement-inl.h`.

Issue #307, opened 2021-05-15
7$: E003 false positive for ++a[0]


strager:
quick-lint-js erroneously reports E003 for the following code:
```js
const a = [42];
++a[0];
```

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

PR #321, opened 2021-05-20
 Fix: False positive on const increment

singalhimanshu:
The following code should not report an error now:
    const a = [42];
    ++a[0];

Fix: #307

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -247,8 +247,10 @@ expression* parser::parse_primary_expression(precedence prec) {
   case token_type::plus_plus: {
     source_code_span operator_span = this->peek().span();
     this->skip();
-    expression* child = this->parse_expression(precedence{
-        .binary_operators = false, .math_or_logical_or_assignment=false, .commas = false});
+    expression* child = this->parse_expression(
+        precedence{.binary_operators = false,
+                   .math_or_logical_or_assignment = false,
+                   .commas = false});
     if (child->kind() == expression_kind::_invalid) {
       this->error_reporter_->report(error_missing_operand_for_operator{
           .where = operator_span,
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -79,7 +79,7 @@ parser::loop_guard parser::enter_loop() {
 
 expression* parser::parse_expression(precedence prec) {
   expression* ast = this->parse_primary_expression(prec);
-  if (!prec.binary_operators) {
+  if (!prec.binary_operators && !prec.is_pre_inc_dec) {
     return ast;
   }
   return this->parse_expression_remainder(ast, prec);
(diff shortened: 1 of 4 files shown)


src/parse.cpp
```diff
@@ -248,7 +248,7 @@ expression* parser::parse_primary_expression(precedence prec) {
     source_code_span operator_span = this->peek().span();
     this->skip();
     expression* child = this->parse_expression(
-        precedence{.binary_operators = false, .commas = false});
+        precedence{.binary_operators = true, .commas = false});
```
strager, line 251: This change doesn't seem right. Won't it break code like this:
```js
++x + y

(++x) + y  // before your commit
++(x + y)  // after your commit
```

strager (changes requested):
Blocking: Add tests.

src/parse.cpp
```diff
@@ -791,6 +791,9 @@ expression* parser::parse_expression_remainder(expression* ast,
     if (!prec.math_or_logical_or_assignment) {
```
strager, line 791: Could we reuse `math_or_logical_or_assignment`?

strager (changes requested):
Blocking: Add tests.

src/parse.cpp
singalhimanshu: I guess we can.

src/parse.cpp
singalhimanshu: > This change doesn't seem right. Won't it break code like this:

Yup it was breaking.

singalhimanshu:
I have added some tests, let me know if there are still more conditions that are to be tested.

test/test-parse-expression.cpp
```diff
@@ -1419,6 +1419,24 @@ TEST_F(test_parse_expression, parse_prefix_plusplus_minusminus) {
...
+  }
+
+  {
+    test_parser p(u8"--x\n+\ny"_sv);
+
+    expression* ast = p.parse_expression();
+    EXPECT_EQ(summarize(ast), "binary(rwunary(var x), var y)");
+    EXPECT_THAT(p.errors(), IsEmpty());
+  }
+}
```
strager, line 1438: I'd like tests for `++x.y` and `++x[y]` in test-parse-express.cpp too.

src/parse.cpp
```diff
@@ -79,7 +79,7 @@ parser::loop_guard parser::enter_loop() {
 
 expression* parser::parse_expression(precedence prec) {
   expression* ast = this->parse_primary_expression(prec);
-  if (!prec.binary_operators) {
+  if (!prec.binary_operators && prec.math_or_logical_or_assignment) {
```
strager, line 82: This looks weird... But it works... 🤷‍♀️

strager (approved):

singalhimanshu:
I have added the test. I think we can ship it.

strager (approved):

Issue #304, opened 2021-05-15
7$: Improve diagnostic for: let x new Array();


strager:
In the following code, the user probably meant to put an `=` between `x` and `new`:
```js
let x new Array();
```
quick-lint-js provides a suboptimal message:
```
error: missing semicolon after statement [E027]
```
Improve this message for the case where a `=` would work well.

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

singalhimanshu:
> Improve this message for the case where a `=` would work well.

Should we show both errors (`missing semicolon` and `missing equal`) or just (`missing equal`).

strager:
> > Improve this message for the case where a `=` would work well.
> 
> Should we show both errors (`missing semicolon` and `missing equal`) or just (`missing equal`).

Only `missing equal`.

PR #329, opened 2021-05-24
Improve diagnostic for variable assignment

singalhimanshu:
Now we report a missing equal error for the following code:

    let x new Array();

Fixes: #304

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -1158,8 +1158,7 @@
   QLJS_ERROR_TYPE(                                                             \
       error_missing_equal_after_variable, "E151",                              \
       { source_code_span expected_equal; },                                    \
-      .error(QLJS_TRANSLATABLE("missing '=' after variable"),                  \
-             expected_equal))                                                  \
+      .error(QLJS_TRANSLATABLE("missing '=' after variable"), expected_equal)) \
                                                                                \
   /* END */
 
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -1154,6 +1154,13 @@
       { source_code_span continue_statement; },                                \
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
+      error_missing_equal_after_variable, "E151",                              \
+      { source_code_span expected_equal; },                                    \
+      .error(QLJS_TRANSLATABLE("missing '=' after variable"),                  \
+             expected_equal))                                                  \
+                                                                               \
   /* END */
 
 namespace quick_lint_js {
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -661,6 +661,11 @@
              switch_keyword))                                                  \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_missing_equal_after_variable, "E202",                              \
+      { source_code_span expected_equal; },                                    \
+      .error(QLJS_TRANSLATABLE("missing '=' after variable"), expected_equal)) \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_missing_expression_between_parentheses, "E078",                    \
       {                                                                        \
         source_code_span left_paren;                                           \
@@ -1154,11 +1159,6 @@
       { source_code_span continue_statement; },                                \
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
-                                                                               \
-  QLJS_ERROR_TYPE(                                                             \
-      error_missing_equal_after_variable, "E151",                              \
-      { source_code_span expected_equal; },                                    \
-      .error(QLJS_TRANSLATABLE("missing '=' after variable"), expected_equal)) \
                                                                                \
   /* END */
 
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -3006,21 +3006,6 @@ class parser {
           break;
         }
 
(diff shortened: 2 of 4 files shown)


singalhimanshu:
I forgot to add the docs. I will do that till tomorrow.

src/quick-lint-js/error.h
```diff
@@ -1139,6 +1139,12 @@
       { source_code_span continue_statement; },                                \
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
+      error_missing_equal_after_variable, "E151",                              \
+      { source_code_span expected_equal; },                                    \
+      .error(QLJS_TRANSLATABLE("missing '=' after variable"), expected_equal)) \
```
strager, line 1146: Nit: Sort alphabetically by `error_missing_equal_after_variable`.

src/quick-lint-js/parse.h
```diff
@@ -3006,6 +3006,19 @@ class parser {
           break;
         }
 
+        case token_type::kw_new: {
+          const char8 *here = this->lexer_.end_of_previous_token();
+          this->error_reporter_->report(error_missing_equal_after_variable{
```
strager, line 3011: Blocking: This change will reject the following legal code:
```js
let x
new Array()
```

We should only report an error if a semicolon cannot be inserted (i.e. if there is no newline before `new`).

src/quick-lint-js/parse.h
```diff
@@ -3006,6 +3006,19 @@ class parser {
           break;
         }
 
+        case token_type::kw_new: {
```
strager, line 3009: We should also report an error for other tokens, such as `class` and `typeof`.

src/quick-lint-js/parse.h
```diff
@@ -3006,6 +3006,19 @@ class parser {
...
+          const char8 *here = this->lexer_.end_of_previous_token();
+          this->error_reporter_->report(error_missing_equal_after_variable{
+              .expected_equal = source_code_span(here, here),
+          });
+          expression *ast = this->parse_expression_remainder(
+              variable,
+              precedence{.commas = false, .in_operator = allow_in_operator});
+          this->lexer_.insert_semicolon();
```
strager, line 3017: Hmm, inserting a semicolon after the expression unconditionally seems wrong. What about the following code?
```js
let a new Array(), b = a.length;
```
I think your change will cause an error at `,` (in addition to the expected error between `a` and `new`).

strager (changes requested):

src/quick-lint-js/parse.h
```diff
@@ -3026,6 +3026,23 @@ class parser {
...
+        case token_type::kw_new: {
+          bool has_leading_newline = this->peek().has_leading_newline;
+          if (!has_leading_newline) {
+            const char8 *here = this->lexer_.end_of_previous_token();
+            this->error_reporter_->report(error_missing_equal_after_variable{
+                .expected_equal = source_code_span(here, here),
+            });
+          }
+          this->visit_binding_element(variable, v, declaration_kind);
+          this->skip();
```
strager, line 3040: Why are we skipping the `class` keyword? `class { potato }` is not the same as `{ potato }`.

test/test-parse-var.cpp
```diff
@@ -498,6 +511,42 @@ TEST(test_parse, parse_invalid_let) {
                             compound_assignment_operator),  //
             declaring_token, offsets_matcher(&code, 0, u8"let"))));
   }
+
+  {
+    for (string8 kw : {u8"typeof", u8"class", u8"new"}) {
+      spy_visitor v;
+      padded_string code(u8"let x " + kw + u8" Array;");
```
strager, line 518: What does `let x = class Array;` even mean? This test looks dubious.

test/test-parse-var.cpp
```diff
@@ -498,6 +511,42 @@ TEST(test_parse, parse_invalid_let) {
...
+      parser p(&code, &v);
+      p.parse_and_visit_module(v);
+      EXPECT_THAT(v.visits, ElementsAre("visit_variable_declaration",  // x
+                                        "visit_variable_use",          // Array
```
strager, line 522: Blocking: `typeof Array` should emit a `visit_variable_typeof_use`.

src/quick-lint-js/parse.h
```diff
@@ -3026,6 +3026,23 @@ class parser {
...
+          this->parse_and_visit_expression(
+              v, precedence{.commas = false, .in_operator = allow_in_operator});
```
strager, line 3042: Blocking: I think this breaks parsing of the following code:
```js
let x
class C {}
x = new C();
```
Please add a test and fix this. I think the problem is that you're parsing and visiting an expression even if `has_leading_newline` is true. If `has_leading_newline` is true, we should automatically insert a semicolon and end variable declaration statement parsing.

src/quick-lint-js/parse.h
```diff
@@ -3026,6 +3026,23 @@ class parser {
               });
           goto initialize_variable;
 
+        case token_type::kw_typeof:
+        case token_type::kw_class:
+        case token_type::kw_new: {
```
strager, line 3031: Nit: Sort cases alphabetically.
```suggestion
        case token_type::kw_class:
        case token_type::kw_new:
        case token_type::kw_typeof: {
```

strager (changes requested):

src/quick-lint-js/parse.h
singalhimanshu: Yes that was wrong I have fixed it now.

strager (approved):
Looks good to me! Let me know when I should merge.

singalhimanshu:
> Looks good to me! Let me know when I should merge.

I guess we can merge it.

PR #330, opened 2021-05-25
feat: add coc-quick-lint-js

fannheyward:


fannheyward:

fannheyward:
recheck

plugin/coc-quick-lint-js/src/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      return [];
+    }
+  }
+
+  public async lint(document: TextDocument) {
+    if (!workspace.match(documentSelector, document)) return;
+    this.diagnosticCollection.clear();
+
+    const filePath = Uri.parse(document.uri).fsPath;
+    const out = cp.spawnSync(this.bin, ['--output-format=vim-qflist-json', filePath], { encoding: 'utf8' }).stdout;
```
strager, line 87: Blocking: Invoking the CLI every keystroke is very inefficient. We should instead interact with quick-lint-js in-process or with a quick-lint-js server.

plugin/coc-quick-lint-js/src/index.ts
```diff
@@ -0,0 +1,122 @@
+import * as cp from 'child_process';
```
strager, line 1: Almost-blocking: Is there a reason why we're not reusing code from the VS Code plugin? The APIs this plugin is interacting with look similar to the APIs our existing VS Code extension uses.

plugin/coc-quick-lint-js/esbuild.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
strager, line 1: Is a build system really necessary? Let's write plain, un-transpiled JavaScript unless transpiling is required or very helpful for some reason.

plugin/coc-quick-lint-js/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+coc.nvim extension for quick-lint-js
+
+## Install
+
+`:CocInstall coc-quick-lint-js`
+
+## License
+
+MIT
```
strager, line 11: Blocking: Our code should be licensed under GPL v3 or newer, unless there's a good reason to license this code under an MIT-style license.

plugin/coc-quick-lint-js/README.md
```diff
@@ -0,0 +1,15 @@
+# coc-quick-lint-js
```
strager, line 1: Could this be part of the Vim plugin (`plugin/vim/`)? Or does coc.nvim require its own infrastructure separate from Vim's and Neovim's for plugin management?

Ideally, this plugin would work out-of-the-box with little or no configuration if you install quick-lint-js with a system package manager, assuming you already have coc.nvim configured. This is the case with the ALE-based Vim plugin already.

strager (changes requested):

fannheyward:
@strager I'm trying `quick-lint-js-wasm` but came error: `ReferenceError: TextEncoder is not defined`.

plugin/coc-quick-lint-js/README.md
fannheyward: No. Like VSCode plugin is published to https://marketplace.visualstudio.com/items?itemName=quick-lint.quick-lint-js, coc-quick-lint-js can be published to npmjs.com, then user can install it by `:CocInstall coc-quick-lint-js`.

plugin/coc-quick-lint-js/src/index.ts
fannheyward: Hi, I'm using `quick-lint-js-wasm` like VSCode, but came error `TextEncoder is not defined`.

plugin/coc-quick-lint-js/README.md
strager: In that case, let's keep it in its own directory (`plugin/coc-quick-lint-js/`).

I think we shouldn't include "quick-lint-js" in the directory name. Could we call it `plugin/coc.nvim/`?

plugin/coc-quick-lint-js/src/index.ts
fannheyward: I've tested with Node.js 14 and 16, both are failed.

strager: > I've tested with Node.js 14 and 16, both are failed.

Hmm, weird. Perhaps we can find a workaround. How can I test it on my machine?

fannheyward: First, you need to install coc.nvim by something like `Plug 'neoclide/coc.nvim', {'branch': 'release'}`, then add `set runtimepath^=~/src/quick-lint-js/plugin/coc-quick-lint-js` to your vimrc.

1. `cd wasm` and build https://github.com/quick-lint/quick-lint-js/tree/master/wasm
1. `cd plugin/coc-quick-lint-js`
2. `yarn install`
3. `vim xxx.js`, `:CocList extensions` will list `coc-quick-lint-js`.

plugin/coc-quick-lint-js/README.md
fannheyward: After I fix the TextEncoder error, I'll improve this.

strager:
> For coc extensions, there are several ways:
> 
>     [...]
>
>     3. coc runs in-process, with wasm, similar the 2nd but more speed.

Let's implement this option. One important reason is that quick-lint-js won't need to be installed separately.

plugin/coc-quick-lint-js/src/index.ts
fannheyward: I can make it working by add `const { TextDecoder, TextEncoder } = require("util");` to `wasm/quick-lint-js.js`.

strager:
I tried the extension. Looks nice so far!

User experience feedback:

* Can we have linting occur while in insert mode? When I'm typing text, it doesn't update diagnostics until after I exit insert mode.
* The error pop-up says "[quick-lint-js E057] [W] use of undeclared variable: asdsadsadasdsadasd". Can we customize this message? We should put the metadata ("[quick-lint-js E057] [W]") at the end, not the beginning.
* It looks like there's a delay in linting of about half a second. There's also some debouncing: if I hold down 'p' to paste a line over and over again, linting doesn't happen until I stop holding 'p'. Can we turn off these, and get snappy linting?
* It looks like there's also a delay for the error message pop-up. If you move the cursor, there's a delay until the pop-up updates. If you hold down a motion, the pop-up only updates after you stop doing the motion. Could we improve this experience?
* It looks like there's a bug involving non-ASCII Unicode. It might be a bug in Coc.nvim. (I'm using version v0.0.80.) This code should have the `a` highlighted, but our extension is highlighting the space after the `a`:

      /*🎸*/ a  ;

Code feedback:

I didn't look at the code closely, but it looks like a copy-paste of the VS Code plugin. Can we avoid duplicating the code?

fannheyward:
> Can we have linting occur while in insert mode? When I'm typing text, it doesn't update diagnostics until after I exit insert mode.

Enable `diagnostic.refreshOnInsertMode` in your coc-settings.json

> Can we customize this message?

You can custom it by `diagnostic.format`, by default is `[%source%code] [%severity] %message`

> It looks like there's a delay in linting of about half a second

It's limit in coc, coc fires textChanged with some delay, same as VSCode

> bug involving non-ASCII Unicode

Reproduced, will look into this.

strager:
> > It looks like there's a delay in linting of about half a second
> 
> It's limit in coc, coc fires textChanged with some delay, same as VSCode

I don't see this delay in VS Code.

> > bug involving non-ASCII Unicode
> 
> Reproduced, will look into this.

If this is a coc.nvim bug, we can work around it by violating the LSP spec and using a locator that counts Unicode code points (instead of counting UTF-16 code units). I'm fine with fixing this in separate PR, after the major bits of the plugin are ready.

fannheyward:
Sorry for the delay.

> a delay in linting of about half a second
> a delay for the error message pop-up

The delay is limited in coc in 300ms.

fannheyward:
@strager do you have any plan to publish `quick-lint-js-wasm` to npmjs?

For VSCode, you can [build the quick-lint-js-wasm package](https://github.com/quick-lint/quick-lint-js/blob/master/wasm/README.md#Building), then use `vsce` to bundle it.

For coc, I can also bundle the wasm, but there're something inconvenient:

1. in `vsce`, you can use `.vscodeignore` to special what files to bundle, but `.npmignore` doesn't work in this way, I can use [bundleddependencies](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bundleddependencies), but this will bundle all files in `node_modules/quick-lint-js-wasm/`, this makes the bundle size large.
2. another way is copy the wasm dist to this plugin manually
3. another problem is we need to build wasm first to bundle it. We can use https://www.npmjs.com/package/quick-lint-js without to build anything. Can we make `quick-lint-js-wasm` to a separate package, and plugins, both VSCode and coc can just download it from npmjs to use.

strager:
> I'd like to finish this extension first, then we can move duplicate code into a node lib and used in both VSCode and coc, something like https://github.com/clangd/node-clangd.

Okay, this sounds good.

I noticed that changes to quick-lint-js-wasm require running `yarn install --force`. If we split off a library, I don't want force contributors to run `yarn install --force` after every .js change. Would Yarn workspaces fix this? Would upgrading to Yarn 2.0 fix this? Would switching to npm fix this?

> @strager do you have any plan to publish `quick-lint-js-wasm` to npmjs?

No. I oppose doing this.

> this makes the bundle size large because `bundleddependencies` will include all modules that `quick-lint-js-wasm` uses, here is prettier. Maybe we can remove prettier from quick-lint-js-wasm to make it small.

To build the package which we ship, we'd run `yarn install --production`, which would not install Prettier.

> Can we make `quick-lint-js-wasm` to a separate package, and plugins, both VSCode and coc can just download it from npmjs to use.

Uploading to npm then downloading from npm then uploading to npm sounds silly and bug-prone.

> build wasm is expensive because we need to install emscripten SDK, Ninja etc.

I agree, but in order to develop this plugin, one also needs to install Vim or Neovim, Coc.nvim, Node.js, and Yarn (and maybe other stuff). (Emscripten does stand out as being bloated and needing careful installation, though.)

---

We can worry about deployment later, after landing this PR. Let's focus on getting the extension working (without copy-pasting) first.

fannheyward:
Have no idea about `yarn install --force` issue, the `coc-quick-lint-js` extension is done. The follow steps:

1. build wasm, remove `build-emscripten` in wasm, other wise this will be bundled in.
2. `npm publish` to npmjs, then users can `:CocInstall coc-quick-lint-js`.

strager:
> > Uploading to npm then downloading from npm then uploading to npm sounds silly and bug-prone.
> 
> You don't need to upload - download - upload, all you need is publish `quick-lint-js-wasm` as node package, then in VSCode and coc, also any other editors maybe, install quick-lint-js extension will download the package to use.

I was talking about contributors, not end users. Sorry for being unclear.

> The `coc-quick-lint-js` extension is done if the code/features are working as your expected.

plugin/coc.nvim/extension.js still looks like a copy-paste from the VS Code plugin. This PR isn't ready until the code is no longer duplicated.

LeeWannacott:
@fannheyward Hey, How would I test this plugin without using `:CocInstall coc-quick-lint-js`? . Perhaps you could update the README.md for potential contributors?
Thanks.

fannheyward:
@LeeWannacott I've updated the README.

1. build the [quick-lint-js-wasm](https://github.com/quick-lint/quick-lint-js/blob/master/wasm/README.md#Building) package
2. `set runtimepath^=/path/to/quick-lint-js/plugin/coc.nvim` in your vimrc/init.vim

LeeWannacott:
Thanks, I love CoC!; keep up the good work!

strager:
Do you plan on finishing this task? If not, that's fine.

I rewrote the VS Code extension from scratch. It now uses a Node.js module and has a different build process than before. (No more emscripten.)

fannheyward:
@strager let me check the new VS Code extension first.

fannheyward:
@strager I've tested the new Node.js module for VSCode, unfortunately, it can't work with coc.nvim. Although coc.nvim provided lots of similar API as VSCode, it's not VSCode.

截屏2021-11-04 下午9 04 08

Because `quick-lint-js` can run in Language Server Mode, now you can use it in coc.nvim with following settings.json to your coc-settings.json:

```json
  "languageserver": {
    "quick-lint-js": {
      "command": "quick-lint-js",
      "args": ["--lsp"],
      "filetypes": [
        "javascript"
      ]
    }
  }
```

I'm going to close this PR, thank you @strager for adding LSP mode in quick-lint-js, now there's no need to make an extension, just copy&paste this short JSON configs.

Issue #172, opened 2021-03-21
12$: Prevent stack overflows from excessive nesting


strager:
```javascript
((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((
```

strager:
((((((((((((((((((((
    {{{{{{{{{{{{{{{{{{{{
    ({{{{{{{{{{{{{{{{{{{
    [[[[[[[[[[[[[[[[[[[[
    ({({({({({({({({({({

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

singalhimanshu:
> ({{{{{{{{{{{{{{{{{{{
> ({({({({({({({({({({

These both report the following error: `fatal: token not implemented in parse_object_literal: left_paren`

strager:
> These both report the following error: `fatal: token not implemented in parse_object_literal: left_paren`.

Oh.

> Do I have to handle these cases?

Nah.

PR #332, opened 2021-05-28
Prevent stack overflows from excessive nesting

singalhimanshu:
Fixes: #172


src/parse.cpp
```diff
@@ -264,10 +267,17 @@ expression* parser::parse_primary_expression(precedence prec) {
...
   case token_type::left_paren: {
+    this->depth_++;
+    if (this->depth_ > this->limit_) {
+      QLJS_PARSER_DEPTH_LIMIT_EXCEEDED();
+    }
     source_code_span left_paren_span = this->peek().span();
     this->skip();
 
     if (this->peek().type == token_type::right_paren) {
+      if (this->depth_ > 0) {
```
strager, line 278: When would depth be 0? Sounds like a bug. If anything, we should *assert* that the depth is >0:
```suggestion
      QLJS_ASSERT(this->depth_ > 0);
```

src/parse.cpp
```diff
@@ -264,10 +267,17 @@ expression* parser::parse_primary_expression(precedence prec) {
   // (x) => {}    // Arrow function.
   // (x + y * z)  // Parenthesized expression.
   case token_type::left_paren: {
+    this->depth_++;
```
strager, line 270: Blocking: Where is the matching `depth--` for inputs like `(x)`? It seems like you only handle `()`.

If I put `(x);` in a file 301 times, your patch causes quick-lint-js to fail. This is wrong.

I think we should use a scope guard to prevent mistakes like this. See `parser::function_guard` for an example.

src/quick-lint-js/parse.h
```diff
@@ -137,6 +142,24 @@ class parser {
     this->have_unimplemented_token_jmp_buf_ = false;
     return ok;
   }
+  // Returns true if parsing succeeded without QLJS_PARSER_DEPTH_LIMIT_EXCEEDED
+  // being called.
+  //
+  // Returns false if QLJS_PARSER_DEPTH_LIMIT_EXCEEDED was called.
+  template <QLJS_PARSE_VISITOR Visitor>
+  bool parse_and_visit_module_catching_parser_depth_limit_exceeded(Visitor &v) {
```
strager, line 150: Let's not make two functions (`parse_and_visit_module_catching_unimplemented` and `parse_and_visit_module_catching_parser_depth_limit_exceeded`). Let's join the two. They do basically the same thing anyway.

Call the function `parse_and_visit_module_catching_fatal_parse_errors` and the variable `fatal_parse_error_jmp_buf_` or something.

src/quick-lint-js/parse.h
```diff
@@ -633,6 +663,9 @@ class parser {
 
     case token_type::end_of_file:
     case token_type::right_curly:
+      if (this->depth_ > 0) {
+        this->depth_--;
+      }
```
strager, line 668: What is this for? What is this supposed to match? This code looks misplaced, but maybe I'm not seeing something.

src/quick-lint-js/parse.h
```diff
@@ -3412,8 +3449,13 @@ class parser {
 #if QLJS_HAVE_SETJMP
   bool have_unimplemented_token_jmp_buf_ = false;
   std::jmp_buf unimplemented_token_jmp_buf_;
+  bool have_parser_depth_limit_exceeded_jmp_buf = false;
+  std::jmp_buf parser_depth_limit_exceeded_jmp_buf_;
 #endif
 
+  std::uint16_t depth_ = 0;
+  const std::uint16_t limit_ = 300;
```
strager, line 3457: Don't allocate `limit_` within the object:
```suggestion
  static constexpr std::uint16_t limit_ = 300;
```

strager, line 3457: Nit: Pick a better name than 'limit'. Perhaps 'stack limit'.

src/quick-lint-js/parse.h
```diff
@@ -3412,8 +3449,13 @@ class parser {
 #if QLJS_HAVE_SETJMP
   bool have_unimplemented_token_jmp_buf_ = false;
   std::jmp_buf unimplemented_token_jmp_buf_;
+  bool have_parser_depth_limit_exceeded_jmp_buf = false;
+  std::jmp_buf parser_depth_limit_exceeded_jmp_buf_;
 #endif
 
+  std::uint16_t depth_ = 0;
```
strager, line 3456: Nit: `std::uint16_t` is usually a bad idea for members of heavy-weight classes like `parser`. `std::uint_fast16_t` would be more appropriate, or just `int`.

test/test-parse.cpp
```diff
@@ -539,6 +539,80 @@ TEST(test_escape_first_character_in_keyword,
   EXPECT_EQ(escape_first_character_in_keyword(u8"b1n z"_sv), u8"\\u{62}1n z");
   EXPECT_EQ(escape_first_character_in_keyword(u8"ZYXW"_sv), u8"\\u{5a}YXW");
 }
+
+#if QLJS_HAVE_SETJMP
+TEST(test_no_overflow, parser_depth_limit_not_exceeded) {
```
strager, line 544: This test should be enabled even if `QLJS_HAVE_SETJMP` is not supported.

test/test-parse.cpp
```diff
@@ -539,6 +539,80 @@ TEST(test_escape_first_character_in_keyword,
...
+  {
+    string8 opening_parens(300, '(');
```
strager, line 546: Instead of repeating `300` everywhere, we should make a constant. Perhaps make `parser::limit_` public and reuse it?

strager (changes requested):
Not caught by this PR, at least on my machine:

* `function f(){` 4001 times.
* `() => {` 4001 times.
* `(` followed by `{x:` 4001 times.
* `if (x)` 8001 times.
* `class C { m() {` 4001 times.
* `await` 4001 times.

test/test-parse.cpp
singalhimanshu: Done

test/test-parse.cpp
singalhimanshu: Done

src/quick-lint-js/parse.h
singalhimanshu: Done

src/quick-lint-js/parse.h
singalhimanshu: Yup it was misplaced.

singalhimanshu:
This code is being repeated a few times:
```cpp
  this->depth_++;
  if (this->depth_ > this->stack_limit) {
    QLJS_PARSER_UNIMPLEMENTED();
  }
```
Should I put this in the constructor of `depth_guard` or is that a bad idea?

strager:
> This code is being repeated a few times:
> 
> ```c++
>   this->depth_++;
>   if (this->depth_ > this->stack_limit) {
>     QLJS_PARSER_UNIMPLEMENTED();
>   }
> ```
> 
> Should I put this in the constructor of `depth_guard`

Yes.

src/quick-lint-js/parse.h
```diff
@@ -3392,6 +3409,26 @@ class parser {
     bool old_value_;
   };
 
+  class depth_guard {
+   public:
+    explicit depth_guard(parser *p, int old_depth_value) noexcept
+        : parser_(p), old_depth_value_(old_depth_value) {
```
strager, line 3415: Instead of accepting `old_depth_value` as a parameter, this constructor could take the value from `p`:
```suggestion
    explicit depth_guard(parser *p) noexcept
        : parser_(p), old_depth_value_(p->depth_) {
```

src/quick-lint-js/parse.h
```diff
@@ -3392,6 +3409,26 @@ class parser {
...
+      p->depth_++;
+      if (p->depth_ > p->stack_limit) {
+        p->crash_on_unimplemented_token(__FILE__, __LINE__, __func__);
+      }
+    }
+
+    depth_guard(const depth_guard &) = delete;
+    depth_guard &operator=(const depth_guard &) = delete;
+
+    ~depth_guard() noexcept { this->parser_->depth_ = this->old_depth_value_; }
```
strager, line 3425: Nit: Also assert that `this->parser_->depth_ == this->old_depth_value_+1;`, ensuring that depth changes are balanced?

src/quick-lint-js/parse.h
```diff
@@ -3392,6 +3409,26 @@ class parser {
     bool old_value_;
   };
 
+  class depth_guard {
+   public:
+    explicit depth_guard(parser *p, int old_depth_value) noexcept
+        : parser_(p), old_depth_value_(old_depth_value) {
+      p->depth_++;
+      if (p->depth_ > p->stack_limit) {
+        p->crash_on_unimplemented_token(__FILE__, __LINE__, __func__);
```
strager, line 3418: Almost-blocking: `crash_on_unimplemented_token` seems like the wrong function to call here. I think the error message will be confusing to users. Did you mean to call `crash_on_depth_limit_exceeded` here?

Almost-blocking: Passing this `__FILE__`, `__line__`, and `__func__` is kinda pointless IMO. Did you mean for the caller to give us this information? If we make a separate function, I'm not sure it's worth keeping line information.

src/parse.cpp
```diff
@@ -73,11 +73,16 @@ parser::function_guard parser::enter_function(function_attributes attributes) {
...
+parser::depth_guard parser::enter_depth() {
+  return depth_guard(this, this->depth_);
+}
+
 parser::loop_guard parser::enter_loop() {
   return loop_guard(this, std::exchange(this->in_loop_statement_, true));
 }
 
 expression* parser::parse_expression(precedence prec) {
+  depth_guard guard = this->enter_depth();
```
strager, line 85: Nit: If we simplify `depth_guard`'s constructor, then we can avoid the `enter_depth` function:
```suggestion
  depth_guard guard(this);
```

src/quick-lint-js/parse.h
```diff
@@ -2463,6 +2471,11 @@ class parser {
   template <QLJS_PARSE_VISITOR Visitor>
   void parse_and_visit_if(Visitor &v) {
     QLJS_ASSERT(this->peek().type == token_type::kw_if);
+    depth_guard guard = this->enter_depth();
+    this->depth_++;
+    if (this->depth_ > this->stack_limit) {
+      QLJS_PARSER_UNIMPLEMENTED();
+    }
```
strager, line 2478: Blocking: You're incrementing `this->depth_` twice. This is wrong. Delete this code and rely on `depth_guard`:
```suggestion
    depth_guard guard = this->enter_depth();
```

My assertion suggestion on line 3425 would detect this (I think).

src/quick-lint-js/parse.h
```diff
@@ -3410,20 +3447,24 @@ class parser {
...
 
  public:
+  static constexpr const int stack_limit = 300;
   // For testing and internal use only.
   [[nodiscard]] loop_guard enter_loop();
 };
 }
 
 #undef QLJS_PARSER_UNIMPLEMENTED
+#undef QLJS_PARSER_DEPTH_LIMIT_EXCEEDED
```
strager, line 3467: You never defined this macro. Undo this change:
```suggestion
```

src/parse.cpp
```diff
@@ -1344,6 +1349,7 @@ expression* parser::parse_arrow_function_body_impl(
 expression* parser::parse_function_expression(function_attributes attributes,
                                               const char8* span_begin) {
   QLJS_ASSERT(this->peek().type == token_type::kw_function);
+  depth_guard guard = this->enter_depth();
```
strager, line 1352: What tests cover this code?

src/quick-lint-js/parse.h
```diff
@@ -2463,6 +2471,11 @@ class parser {
   template <QLJS_PARSE_VISITOR Visitor>
   void parse_and_visit_if(Visitor &v) {
     QLJS_ASSERT(this->peek().type == token_type::kw_if);
+    depth_guard guard = this->enter_depth();
```
strager, line 2474: What about `while`, `with`, etc.?

Add tests for this code.

test/test-parse.cpp
```diff
@@ -539,6 +539,80 @@ TEST(test_escape_first_character_in_keyword,
   EXPECT_EQ(escape_first_character_in_keyword(u8"b1n z"_sv), u8"\\u{62}1n z");
   EXPECT_EQ(escape_first_character_in_keyword(u8"ZYXW"_sv), u8"\\u{5a}YXW");
 }
+
+TEST(test_no_overflow, parser_depth_limit_not_exceeded) {
+  {
+    string8 opening_parens(parser::stack_limit - 1, '(');
```
strager, line 545: Could we use a loop to avoid repeating these 6 lines of code many times?

strager (changes requested):
I think this patch needs some more tests for me to have confidence.

strager:
~depth_guard's assertion is failing on Windows because Windows' longjmp invokes C++ destructors (because longjmp is implemented using exceptions, apparently): https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/longjmp?view=msvc-160 The innermost depth_guard's destructor isn't called (because the constructor didn't finish, I guess?) but the destructor for each other depth_guard does get called. This imbalance leads to the assertion failure.

To make the assertion pass, have the constructor not change `depth_` before calling `crash_on_depth_limit_exceeded()`.

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+      spy_visitor v;
+      parser p(&code, &v);
+      bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+      EXPECT_TRUE(ok);
+    }
+  }
+
+  {
+    string8 left_curly(parser::stack_limit, '{');
+    padded_string code(left_curly);
```
strager, line 557: Nit: Why are the limits for two tests different (other than `[` vs `{`)?

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+    bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+    EXPECT_TRUE(ok);
+  }
+
+  {
+    for (const string8 &exp :
+         {u8"function f(){", u8"() => {", u8"if(){", u8"await", u8"do{",
+          u8"for(){", u8"try{", u8"while(){", u8"with({}){"}) {
+      string8 exps;
+      exps.reserve(exp.size() * (parser::stack_limit));
```
strager, line 569: Nit:
```suggestion
      exps.reserve(exp.size() * parser::stack_limit);
```

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+      EXPECT_FALSE(ok);
+      ElementsAre(
+          ::testing::VariantWith<error_depth_limit_exceeded>(::testing::_));
+    }
+  }
+
+  {
+    string8 object_decl(u8"{x:");
+    string8 paren_object_decl(u8"(");
+    paren_object_decl.reserve(object_decl.size() * (parser::stack_limit + 1));
```
strager, line 633: Nit: Reserved capacity is off by 1:
```suggestion
    paren_object_decl.reserve(paren_object_decl.size() + object_decl.size() * (parser::stack_limit + 1));
```

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+      parser p(&code, &v);
+      bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+      EXPECT_TRUE(ok);
+    }
+  }
+
+  {
+    string8 object_decl(u8"{x:");
+    string8 paren_object_decls(u8"(");
+    paren_object_decls.reserve(object_decl.size() * (parser::stack_limit - 2));
```
strager, line 584: Nit: Reserved capacity is off by 1:
```suggestion
    paren_object_decls.reserve(paren_object_decls.size() + object_decl.size() * (parser::stack_limit - 2));
```

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+      EXPECT_FALSE(ok);
+      ElementsAre(
+          ::testing::VariantWith<error_depth_limit_exceeded>(::testing::_));
+    }
+  }
+
+  {
+    for (const string8 &exp :
+         {u8"function f(){", u8"() => {", u8"if(){", u8"await ", u8"do{",
+          u8"for(){", u8"try{", u8"while(){", u8"with({}){"}) {
```
strager, line 614: Same comment on lines 566-567, except here it'll actually matter when we fix the false negatives (i.e. when quick-lint-js reports an error for `if(){}`).

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+    spy_visitor v;
+    parser p(&code, &v);
+    bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+    EXPECT_TRUE(ok);
+  }
+
+  {
+    for (const string8 &exp :
+         {u8"function f(){", u8"() => {", u8"if(){", u8"await", u8"do{",
+          u8"for(){", u8"try{", u8"while(){", u8"with({}){"}) {
```
strager, line 567: `if(){` looks invalid. Write `if(true){` instead?

`for` and `while` too.

test/test-parse.cpp
```diff
@@ -539,6 +539,111 @@ TEST(test_escape_first_character_in_keyword,
...
+    padded_string code(paren_object_decl);
+    spy_visitor v;
+    parser p(&code, &v);
+    bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+    EXPECT_FALSE(ok);
+    ElementsAre(
+        ::testing::VariantWith<error_depth_limit_exceeded>(::testing::_));
+  }
+}
+#endif
```
strager, line 646: More cases I think you missed:
```js
if (true)
  if (true)
    if (true)
      ...
```
```js
class C { m() {
  class C { m() {
    class C { m() {
      ...
```

I think the solution is to add a guard to `parse_and_visit_statement`, not to selective statements.

test/test-parse.cpp
singalhimanshu: This is because of the way `[` and `{` are parsed, the number of function calls (to `parse_expression`) made for `((` or `[[` is 3 times and for `{{` it is 2 times.

strager: That specific different sounds uninteresting to test. I'd merge these test cases into one loop, and test with `parser::stack_limit - 1` repetitions.

test/test-parse.cpp
```diff
@@ -539,6 +540,208 @@ TEST(test_escape_first_character_in_keyword,
...
+      const int rep_count = parser::stack_limit / 2 - 1;
+      string8 exps;
+      exps.reserve((exp.size() + 1) * rep_count);
+      for (int i = 0; i < rep_count; i++) {
+        exps.append(exp);
+      }
+      for (int i = 0; i < rep_count; i++) {
+        exps.push_back(right_square_bracket);
+      }
+      padded_string code(exps);
```
strager, line 600: Maybe we could make a function to help build these strings:

```c++
padded_string repeated_code(string8_view before,
                            string8_view inner,
                            string8_view after,
                            int depth);
```

```suggestion
      padded_string code = repeated_code(exp, u8"", u8"}", parser::stack_limit / 2 - 1);
```

test/test-parse.cpp
```diff
@@ -539,6 +540,208 @@ TEST(test_escape_first_character_in_keyword,
...
+    spy_visitor v;
+    parser p(&code, &v);
+    bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
+    EXPECT_TRUE(ok);
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+
+  {
+    const string8 try_exp = u8"try{ ";
+    const string8 catch_exp = u8"} catch(e) {}";
```
strager, line 631: Could we do the loop-over-pairs thing you do on line 546 to avoid repeating a bunch of this code just to distinguish try/do/for/class? (Or maybe my `repeated_code` helper would fix it.)

In fact, instead of using a pair, you can make a struct to hold before, inner, after, and depth. Then *all* of these tests would be data-driven. See `TEST_F(test_file_canonical, parent_of_non_root_removes_component_posix)` or `TEST_F(test_parse_expression, await_in_normal_function_vs_async_function_vs_top_level)` for examples.

strager (approved):
Test code looks a bit crummy, but the implementation looks great!

test/test-parse.cpp
```diff
@@ -539,6 +540,121 @@ TEST(test_escape_first_character_in_keyword,
...
+TEST(test_no_overflow, parser_depth_limit_not_exceeded) {
+  {
+    for (const string8 &exps : {
+             repeated_str(u8"(", u8"10", u8")", parser::stack_limit - 2),
+             repeated_str(u8"[", u8"10", u8"]", parser::stack_limit - 2),
+             repeated_str(u8"{", u8"10", u8"}", parser::stack_limit - 2),
+             repeated_str(u8"while(true) ", u8"10", u8"",
+                          parser::stack_limit - 2),
+             repeated_str(u8"for(;;) ", u8"10", u8"", parser::stack_limit - 2),
+             repeated_str(u8"await ", u8"", u8"", parser::stack_limit - 2),
```
strager, line 568: Nit: Hmm...?
```suggestion
             repeated_str(u8"await ", u8"10", u8"", parser::stack_limit - 2),
```

test/test-parse.cpp
```diff
@@ -539,6 +540,121 @@ TEST(test_escape_first_character_in_keyword,
...
+#if QLJS_HAVE_SETJMP
+TEST(test_overflow, parser_depth_limit_exceeded) {
+  for (const string8 &exps : {
+           repeated_str(u8"(", u8"10", u8")", parser::stack_limit + 1),
+           repeated_str(u8"[", u8"10", u8"]", parser::stack_limit + 1),
+           repeated_str(u8"{", u8"10", u8"}", parser::stack_limit + 1),
+           repeated_str(u8"while(true) ", u8"10", u8"",
+                        parser::stack_limit + 1),
+           repeated_str(u8"for(;;) ", u8"10", u8"", parser::stack_limit + 1),
+           repeated_str(u8"await ", u8"", u8"", parser::stack_limit + 1),
```
strager, line 619: Nit:
```suggestion
           repeated_str(u8"await ", u8"10", u8"", parser::stack_limit + 1),
```

test/test-parse.cpp
```diff
@@ -539,6 +540,121 @@ TEST(test_escape_first_character_in_keyword,
...
+           repeated_str(u8"if(true) ", u8"10", u8"", parser::stack_limit + 1),
+           repeated_str(u8"function f() { ", u8"", u8"}",
+                        parser::stack_limit + 1),
+           repeated_str(u8"() => { ", u8"", u8"}", parser::stack_limit / 2),
```
strager, line 623: Nit: I'd rather not be so precise.
```suggestion
           repeated_str(u8"() => { ", u8"", u8"}", parser::stack_limit + 1),
```

strager (approved):

PR #334, opened 2021-05-30
CI: check for broken links on website

strager:
Ensure all internal links on our website are working.
    
External links are not checked to avoid flakiness.

### DIFF
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -11,7 +11,7 @@ jobs:
   build:
     name: website
     runs-on: ubuntu-latest
-    container: ghcr.io/quick-lint/quick-lint-js-github-builder:v1
+    container: ghcr.io/quick-lint/quick-lint-js-github-builder:v3
     env:
       QLJS_COLLECT_COPYRIGHT_NO_WARNINGS: 1
     steps:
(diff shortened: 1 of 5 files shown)


strager:
@emadflash Can you take a look at this PR and see if anything looks wrong?

.github/workflows/build-website.yml
```diff
@@ -40,6 +40,29 @@ jobs:
...
         run: cd website && zip -r ../website.zip www
 
+      - name: set up check-links
+        run: |
+          python3.7 -m venv ENV
+          ENV/bin/pip install bs4 requests
+      - name: check links (dev)
+        run: |
+          cd website/
+          yarn start localhost 9001 &
```
emadflash, line 50: why start the server inside website/ ? it should be website/public
something is wrong, its not reporting errors that way

emadflash (changes requested):

.github/workflows/build-website.yml
strager: When I develop the website, I always run `yarn start` from within website/.

The website/public/ directory contains the page sources, but doesn't contain the server entry point. The server entry point is website/run-server.mjs.

emadflash: ok got. so yarn sets the entry point. i was using `python3 -m http.server 8080` inside /website/

strager: Ah, I see why you were confused. The website's guts have been restructured since you made check-links.

PR #336, opened 2021-05-30
Add --config-file to --help and shell completions

strager:


strager:
@maniac-en Can you sanity-check this PR?

maniac-en:
> @maniac-en Can you sanity-check this PR?

All shell completions are working as expected, so it should be ok to merge!

PR #323, opened 2021-05-21
Show error message when mouse over mark

matttheus:
To get this done I needed change the textarea to a div and add the attribute contenteditable="true" to the editor.

PR #339, opened 2021-06-01
GitHub pr/258

matttheus:
Creating this pull request because #323 was closed because staging branch get deleted.


strager:
Regarding the failing test:

* We should probably not generate attributes with "undefined" as the value.
* If we do want to generate attributes, we need to change the <mark> detection logic in the `code_block` function in website/src/error-documentation.mjs.

matttheus:
> Regarding the failing test:
> 
>     * We should probably not generate attributes with "undefined" as the value.
> 
>     * If we do want to generate attributes, we need to change the  detection logic in the `code_block` function in website/src/error-documentation.mjs.

Please, look at #364.

strager:
This is being developed on PR #364.

PR #343, opened 2021-06-03
Pin to working version of eshost for test-all-js-engines

strager:
test-all-js-engines uses the latest published eshost version. It seems
that version 8.0.0 is broken. test-all-js-engines used to report an
error for the following JavaScript code, but doesn't any more:

    if (true)

Version 7.6.0 seems to work better. Use that working version.


strager:
@singalhimanshu confirmed that this fixes test-all-js-engines.

Issue #335, opened 2021-05-30
10$: Website: fix validation failures


strager:
https://validator.nu/?doc=https%3A%2F%2Fquick-lint-js.com%2F

jgil-r:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
@jgil-r Did you look at all other pages for validation errors?

jgil-r:
I didn't, I'll take a look right now.

jgil-r:
@strager I found a bunch of other validation errors. Ill fix them and submit a new pull request.

strager:
These pages still have validator errors or warnings:

* https://quick-lint-js.com/license/
* https://quick-lint-js.com/hiring/
* https://quick-lint-js.com/benchmarks/ (but I don't care about the type=module warning)

strager:
Thanks for the fixes.

PR #346, opened 2021-06-04
Full website validation errors

jgil-r:
Fixed validation errors in install/error/cli/benchmark pages. #335

### DIFF
--- a/website/public/benchmarks/criterion-json-to-html.mjs
+++ b/website/public/benchmarks/criterion-json-to-html.mjs
@@ -69,9 +69,15 @@ export function makeBenchmarkHTML(seriess) {
       )}
 
       <div class="x-axis-label">response time (lower is better)</div>
+      <!-- 
+        title attribute must be broken into new line
+        because the html spec treats &#013; as a parse error, which
+        gives us validation errors.
+      -->
       <label
         class="log-scale javascript-only"
-        title="checked: logarithmic X axis (emphasizing ratios).&#013;unchecked: linear X axis (emphasizing time)."
+        title="checked: logarithmic X axis (emphasizing ratios).
+        unchecked: linear X axis (emphasizing time)."
         ><input type="checkbox" class="log-scale-select" /> log scale</label
       >
     </div>
--- a/website/public/benchmarks/criterion-json-to-html.mjs
+++ b/website/public/benchmarks/criterion-json-to-html.mjs
@@ -69,15 +69,10 @@ export function makeBenchmarkHTML(seriess) {
       )}
 
       <div class="x-axis-label">response time (lower is better)</div>
-      <!-- 
-        title attribute must be broken into new line
-        because the html spec treats &#013; as a parse error, which
-        gives us validation errors.
-      -->
       <label
         class="log-scale javascript-only"
         title="checked: logarithmic X axis (emphasizing ratios).
-        unchecked: linear X axis (emphasizing time)."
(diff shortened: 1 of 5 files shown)


website/public/benchmarks/criterion-json-to-html.mjs
```diff
@@ -69,9 +69,15 @@ export function makeBenchmarkHTML(seriess) {
       )}
 
       <div class="x-axis-label">response time (lower is better)</div>
+      <!-- 
+        title attribute must be broken into new line
+        because the html spec treats &#013; as a parse error, which
+        gives us validation errors.
+      -->
```
strager, line 76: Nit: No need for this comment. Delete it.

website/public/benchmarks/criterion-json-to-html.mjs
```diff
@@ -69,9 +69,15 @@ export function makeBenchmarkHTML(seriess) {
...
       <label
         class="log-scale javascript-only"
-        title="checked: logarithmic X axis (emphasizing ratios).&#013;unchecked: linear X axis (emphasizing time)."
+        title="checked: logarithmic X axis (emphasizing ratios).
+        unchecked: linear X axis (emphasizing time)."
```
strager, line 80: Blocking: This adds a bunch of spaces before the word 'unchecked':
Screen Shot 2021-06-04 at 00 14 53

It's supposed to be flush with the left side:
Screen Shot 2021-06-04 at 00 15 09

website/public/install/index.ejs.html
```diff
@@ -1540,7 +1544,6 @@ <h4 id="manual-vim-pathogen">Install with Pathogen</h4>
           This option requires
           <a href="https://github.com/tpope/vim-pathogen">Pathogen</a>.
         </p>
-        <p />
```
strager, line 1543: Heh, how'd that get there? xD

strager (changes requested):

website/public/benchmarks/criterion-json-to-html.mjs
jgil-r: Would you like me to change it back to how it was initially? If so, we will still get the following validation error: 
```
A numeric character reference expanded to carriage return
```

I was reading about the [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#multiline_titles) attribute and removing the extra spacing after the line break seems to give the right behavior.

strager: Your latest version looks good to me.

strager:
It looks like there are merge conflicts. Rebase, then I'm happy to merge.

strager (approved):

website/public/install/index.ejs.html
```diff
@@ -137,7 +137,12 @@
       }
 
       .long-shell-command-line {
+<<<<<<< HEAD
         /* 6 = "$ ".length + 4 spaces of indentation */
+=======
+        /* 4 = "$ ".length + 4 spaces of indentation */
+        display: inline-block;
+>>>>>>> fixed validation errors for /install page
```
strager, line 145: Blocking: Don't commit merge markers.

strager (changes requested):

website/public/install/index.ejs.html
jgil-r: Just removed them, thank you for you patience!

strager (approved):

PR #350, opened 2021-06-06
CI: fix emacs test failure

### LINKED ISSUE #349: CI: emacs test failure
https://github.com/quick-lint/quick-lint-js/runs/2754717510

```
Run ./plugin/emacs/test-emacs.sh
Contacting host: stable.melpa.org:443
Contacting host: elpa.gnu.org:80
Failed to download `gnu' archive.
Package `let-alist-1.0.4' is unavailable
Error: Process completed with exit code 255.
```

singalhimanshu:
The changes are done as per these instructions - https://github.com/melpa/melpa#usage
Fix #349

### DIFF
--- a/plugin/emacs/test/quicklintjs-test.el
+++ b/plugin/emacs/test/quicklintjs-test.el
@@ -13,7 +13,7 @@
   (setq package-user-dir cache-dir-name
         package-check-signature nil)
   (add-to-list 'package-archives
-               '("MELPA" . "https://stable.melpa.org/packages/"))
+               '("melpa" . "https://melpa.org/packages/") t)
   (package-initialize)
 
   (unless package-archive-contents


plugin/emacs/test/quicklintjs-test.el
```diff
@@ -13,7 +13,7 @@
   (setq package-user-dir cache-dir-name
         package-check-signature nil)
   (add-to-list 'package-archives
-               '("MELPA" . "https://stable.melpa.org/packages/"))
+               '("melpa" . "https://melpa.org/packages/") t)
```
strager, line 16: Which was the problem: capitalization of the first string, `stable.`, or ` t`?

strager (approved):
@wggr Is there a reason you opted for MELPA Stable? Did you assume that MELPA Stable would be more reliable for CI?

From https://github.com/melpa/melpa#melpa-stable:

> Note that the MELPA maintainers do not use MELPA Stable themselves, and do not particularly recommend its use.

strager:
It looks like the problem is fixed upstream, so this PR isn't needed to fix CI.

plugin/emacs/test/quicklintjs-test.el
singalhimanshu: I guess this wasn't the issue, the issue is probably because of older emacs version which has some emacs-gpg issues.

singalhimanshu:
> It looks like the problem is fixed upstream, so this PR isn't needed to fix CI.

I dont' think it is solved, the problem seems to be random.

PR #351, opened 2021-06-07
More validation errors

jgil-r:
Fixed validation errors in /license page where paragraph tag is not allowed to
be a children of an ordered list, and added lang attribute to html element (this
attribute was also added to /hiring and /benchmarks pages) #335. In /benchmarks page, we still get the following validation error `Error CSS: color: 70% is not a valid color 3 or 6 hexadecimals numbers`, which seems to be a validator bug. I have opened up an issue with validator, which can be tracked [here](https://github.com/validator/validator/issues/1160)

### DIFF
--- a/website/public/benchmarks/index.ejs.html
+++ b/website/public/benchmarks/index.ejs.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <!-- Copyright (C) 2020  Matthew "strager" Glazar -->
 <!-- See end of file for extended copyright information. -->
-<html>
+<html lang="en">
   <head>
     <%- await include("../common-head.ejs.html") %>
     <script>
@@ -23,11 +23,11 @@
     <link href="benchmark.css" rel="stylesheet" />
     <style>
       .linter-name {
-        color: hsla(var(--hue), 70%, 40%);
+        color: hsla(var(--hue), 70%, 40%, 1);
       }
       @media (prefers-color-scheme: dark) {
         .linter-name {
-          color: hsla(var(--hue), 85%, 75%);
+          color: hsla(var(--hue), 85%, 75%, 1);
         }
       }
     </style>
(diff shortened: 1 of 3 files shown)


website/public/benchmarks/index.ejs.html
```diff
@@ -95,7 +95,9 @@ <h2>Methodology</h2>
         <p>These benchmarks measure the following linters:</p>
         <ul>
           <li>
-            <strong class="linter-name" style="--hue: 0">quick-lint-js</strong>
+            <strong class="linter-name" style="--hue: 0"
```
strager, line 98: Why these whitespaces changes? Are they significant? Or am I missing something?

website/public/license/index.ejs.html
```diff
@@ -2181,162 +2181,165 @@ <h4>How to Apply These Terms to Your New Programs</h4>
...
+              party alleging that the Package constitutes direct or contributory
+              patent infringement, then this Artistic License to you shall
+              terminate on the date that such litigation is filed.
+            </p>
           </li>
-
           <li>
-            <p>Disclaimer of Warranty:</p>
+            <h4>Disclaimer of Warranty:</h4>
             <p>
```
strager, line 2343: I'm confused. Why is the first `<p>` disallowed but the second `<p>` allowed?

website/public/benchmarks/index.ejs.html
jgil-r: hmm not sure, ive been running prettier. Ill fix that.

website/public/license/index.ejs.html
jgil-r: For the elements I changed to `<h4>`. I assumed they were some level of subheading compared to the text that is was describing. A paragraph tag would work there if you would like me to change it to that. The main issue here was that the text I changed to `<h4>` tags were paragraph tags acting as headings outside of `<li>`, which is what was triggering the validation error.

strager: Here's what it should look like (approximately): https://github.com/quick-lint/quick-lint-js/blob/1a92c01a12d490427d297984ac64347af7cbe46f/website/public/npm.svg#L77-L102

The `<h4>`s should be *outside* the lists.

jgil-r: Yeah I see what you are saying. The first `<p>` was not allowed because it was outside of a list item, and `<ol>` or `<ul>` elements can only have a `<li>` as a child. That`s why I had to wrap every section in a `<li>` tag. I did notice the numbering disappearing, so I figured I could add a nested list, the problem with that though was that the numbering wouldn't be continuous past other list items.

For example, this was giving validation error:
```html
<ol>
    <p>Permission for Use and Modification Without Distribution</p>
    <li>
         You are permitted to use the Standard Version and create and use
         Modified Versions for any purpose without restriction, provided that
         you do not Distribute the Modified Version.
    </li>
</ol>
```

This will not give validation error:
```html
<ol>
    <li>
        <h4>Permission for Use and Modification Without Distribution</h4>
        <p>
              You are permitted to use the Standard Version and create and use
              Modified Versions for any purpose without restriction, provided
              that you do not Distribute the Modified Version.
        </p>
    </li>
</ol>
```

Also, the reason the continuous numbering disappeared (photo below being the original) was because the `<li>` tags were being used anywhere it was needed along with other invalid children. I can probably just use css to target those items that need numbers with the ::before selector. I hope this makes sense.

list_error

strager: > For example, this was giving validation error:

Ah, okay. Thanks for clarifying. I misunderstood the diff.

> I can probably just use css to target those items that need numbers with the ::before selector. I hope this makes sense.

Using CSS sounds awful! xD

> the problem with that though was that the numbering wouldn't be continuous past other list items.

Would something like this work:
```html
<h4>Permission for Use and Modification Without Distribution</h4>
<ol start="1">
<li>You are permitted to use ... the Modified Version.</li>
</ol>

<h4>Permissions for Redistribution of the Standard Version</h4>
<ol start="2">
<li>You may Distribute verbatim copies ... of the Package.</li>
<li>You may apply any bug fixes, ... subject to the Original License.</li>
</ol>

<h4>...</h4>
<ol start="4">
<li>...</li>
</ol>
```

jgil-r: Yeah that works. Would you like me to remove the list default left padding? it makes it indent and not line up with its heading.

list_new

strager: > Would you like me to remove the list default left padding? it makes it indent and not line up with its heading.

With CSS? Sounds unnecessary to me. I'm fine with how your screenshot looks.

strager (approved):

strager:
Is this ready to merge?

jgil-r:
> Is this ready to merge?

Yes!

PR #354, opened 2021-06-11
Cleaned up your SVGs in /website/public

javabryan:
Tested on local server 9001. A couple of these had copyright info so I left that in. Hopefully this saves you some bytes!


website/public/gnome-terminal.svg
```diff
@@ -1,241 +1,16 @@
...
-        <dc:type
-           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
-        <dc:creator>
-          <cc:Agent>
-            <dc:title>GNOME Design Team</dc:title>
-          </cc:Agent>
-        </dc:creator>
-        <dc:source />
-        <cc:license
-           rdf:resource="http://creativecommons.org/licenses/by-sa/4.0/" />
```
strager, line 129: Please don't drop all this metadata.

PR #362, opened 2021-06-16
Disallow parameter named 'await' in async arrow function

### LINKED ISSUE #215: 7$: Disallow parameter named 'await' in async arrow function
This code is illegal, but quick-lint-js reports no error for it:
```javascript
(async (await) => null);
```

erlliam:
Resolves #215

What should be done for the broken test in test-parse-var.cpp?
Also I'm not sure what the error message should be so I just wrote random stuff.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -561,6 +561,17 @@ expression* parser::parse_async_expression_only(token async_token) {
                 .arrow = this->peek().span(),
             });
       }
+      for (auto parameter : parameters) {
+        if (parameter->kind() == expression_kind::variable &&
+            parameter->variable_identifier_token_type() ==
+                token_type::kw_await) {
+          // async (await) => {}  // Invalid
+          this->error_reporter_->report(
+              error_cannot_declare_await_in_async_function{
+                  .name = parameter->variable_identifier(),
+              });
+        }
+      }
       // TODO(strager): Should we call maybe_wrap_erroneous_arrow_function?
       return parse_arrow_function_arrow_and_body(std::move(parameters));
     } else {
@@ -597,6 +608,13 @@ expression* parser::parse_async_expression_only(token async_token) {
       goto variable_reference;
     }
 
+    if (this->peek().type == token_type::kw_await) {
+      // async await => {}  // Invalid
+      this->error_reporter_->report(
+          error_cannot_declare_await_in_async_function{
+              .name = this->peek().identifier_name(),
+          });
+    }
     std::array<expression*, 1> parameters = {
         this->make_expression<expression::variable>(
             identifier(this->peek().span()), this->peek().type)};
--- a/test/test-parse-expression-statement.cpp
+++ b/test/test-parse-expression-statement.cpp
@@ -1369,6 +1369,64 @@ TEST(test_parse, let_as_statement_body_allows_asi) {
                 ElementsAre(spy_visitor::visited_variable_assignment{u8"x"}));
   }
 }
+
+TEST(test_parse, disallow_await_parameter_in_async_arrow_function) {
+  {
+    spy_visitor v;
+    padded_string code(u8"(async (await) => null)"_sv);
+    parser p(&code, &v);
+    p.parse_and_visit_expression(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_variable_declaration",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope"));
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_cannot_declare_await_in_async_function, name,
+                    offsets_matcher(&code, strlen(u8"(async ("), u8"await"))));
+  }
+
+  {
+    spy_visitor v;
+    padded_string code(u8"(async await => null)"_sv);
+    parser p(&code, &v);
+    p.parse_and_visit_expression(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_variable_declaration",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope"));
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_cannot_declare_await_in_async_function, name,
+                    offsets_matcher(&code, strlen(u8"(async "), u8"await"))));
+  }
+
+  {
+    spy_visitor v = parse_and_visit_statement(u8"async(await)"_sv);
+    EXPECT_THAT(v.visits,
+                ElementsAre("visit_variable_use", "visit_variable_use"));
+  }
+
+  {
+    spy_visitor v;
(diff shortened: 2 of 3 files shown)


strager:
> What should be done for the broken test in test-parse-var.cpp?

Something like this:
https://github.com/quick-lint/quick-lint-js/blob/8a0a1db8685673b7d8d6b9957bc31b148175962f/test/test-parse-var.cpp#L1842-L1846

src/quick-lint-js/error.h
```diff
@@ -1171,6 +1171,13 @@
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_await_parameter_disallowed_in_async_arrow_function, "E696",        \
+      { source_code_span await_token; },                                       \
+      .error(QLJS_TRANSLATABLE("you can't use a parameter named await in an "  \
+                               "async arrow function"),                        \
```
strager, line 1178:
```suggestion
      .error(QLJS_TRANSLATABLE("unexpected parameter named 'await' in an "  \
                               "async arrow function"),                        \
```

test/test-parse-expression-statement.cpp
```diff
@@ -1369,6 +1369,30 @@ TEST(test_parse, let_as_statement_body_allows_asi) {
...
+                    await_token,
+                    offsets_matcher(&code, strlen(u8"(async ("), u8"await"))));
+  }
+
+  {
+    spy_visitor v = parse_and_visit_expression(u8"((await) => null)"_sv);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope"));
+  }
```
strager, line 1394: This is already tested by test_parse.variables_can_be_named_contextual_keywords, no?

test/test-parse-expression-statement.cpp
```diff
@@ -1369,6 +1369,30 @@ TEST(test_parse, let_as_statement_body_allows_asi) {
                 ElementsAre(spy_visitor::visited_variable_assignment{u8"x"}));
   }
 }
+
+TEST(test_parse, disallow_await_parameter_in_async_arrow_function) {
+  {
+    spy_visitor v;
+    padded_string code(u8"(async (await) => null)"_sv);
```
strager, line 1376: What about without parentheses? `(async await => null)`

test/test-parse-expression-statement.cpp
erlliam: Good point

src/parse.cpp
```diff
@@ -541,6 +541,12 @@ expression* parser::parse_async_expression_only(token async_token) {
         this->skip();
         continue;
       }
```
erlliam, line 543: Should I add the `// async await => {}  // Invalid` comment here?

test/test-parse-var.cpp
```diff
@@ -1802,13 +1804,18 @@ TEST(test_parse, variables_can_be_named_contextual_keywords) {
...
     }
 
-    // TODO(#215): Disallow 'await' as parameter to async arrow function.
     for (string8 code : {
              u8"(async " + name + u8" => null)",
              u8"(async (" + name + u8") => null)",
              u8"(" + name + u8" => null)",
              u8"((" + name + u8") => null)",
          }) {
+      if (name == u8"await" &&
```
erlliam, line 1813: Should this change land as it's own commit before all the other changes?

src/parse.cpp
strager: But shouldn't the comment say `async (await) => {}`?

src/parse.cpp
```diff
@@ -541,6 +541,12 @@ expression* parser::parse_async_expression_only(token async_token) {
...
+      if (this->peek().type == token_type::kw_await) {
+        this->error_reporter_->report(
+            error_await_parameter_disallowed_in_async_arrow_function{
+                .await_token = this->peek().span(),
+            });
```
strager, line 548: Blocking: We should only emit an error if we know this is an arrow function. Here's a legal example of `async(await` which is not an arrow function:
```js
var async = console.log;
var await = 'taco';
async(await);
```

src/quick-lint-js/error.h
```diff
@@ -1171,6 +1171,13 @@
       .error(QLJS_TRANSLATABLE("continue can only be used inside of a loop"),  \
              continue_statement))                                              \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_await_parameter_disallowed_in_async_arrow_function, "E696",        \
```
strager, line 1175: Nit: Sort alphabetically by struct name.

test/test-parse-var.cpp
strager: Nah, this is fine as-is.

strager (changes requested):

src/parse.cpp
```diff
@@ -522,6 +522,7 @@ expression* parser::parse_async_expression_only(token async_token) {
   // async()         // Function call.
   case token_type::left_paren: {
     bool newline_after_async = this->peek().has_leading_newline;
+    std::optional<source_code_span> await_as_parameter_token = std::nullopt;
```
strager, line 525: What if there are multiple parameters named `await`?
```js
(async (await, await, await) => {});
```

src/parse.cpp
```diff
@@ -522,6 +522,7 @@ expression* parser::parse_async_expression_only(token async_token) {
...
 
     vector<expression*> parameters(
```
strager, line 527: Could we reuse `parameters` instead of maintaining a separate variable?

test/test-parse-expression-statement.cpp
```diff
@@ -1369,6 +1369,46 @@ TEST(test_parse, let_as_statement_body_allows_asi) {
...
```
strager, line 1376: What if `(async (await) => null)` is found within an async function? Would it report multiple errors?

The following code produces two errors, but I think your commit will make it produce three errors:
```js
async function f() {
  (async (await) => null);
}
```
```
hello.js:2:11: error: missing operand for operator [E026]
hello.js:2:11: error: cannot declare 'await' inside async function [E069]
```

Maybe we should avoid calling `parse_expression` if we see an `await` token?

Man, this stuff is confusing.

src/parse.cpp
```diff
@@ -566,11 +561,14 @@ expression* parser::parse_async_expression_only(token async_token) {
...
             });
       }
-      if (await_as_parameter_token) {
-        this->error_reporter_->report(
-            error_await_parameter_disallowed_in_async_arrow_function{
-                .await_token = *await_as_parameter_token,
-            });
+      for (auto parameter : parameters) {
+        if (parameter->variable_identifier_token_type() ==
+            token_type::kw_await) {
```
strager, line 566: Blocking: What if the parameter isn't a variable? E.g. `([destructured, array]) => {}`. Your code will crash. So I think we need to check `parameter->kind()`.

strager (changes requested):

strager (approved):

Issue #267, opened 2021-05-05
7$: Better error if generator `*` is before `function` 


strager:
In the following code, the user tried to declare a generator function:
```javascript
*function myGenerator() {
  yield 'hello';
}
```
quick-lint-js reports confusing errors for this code. quick-lint-js should instead report something like the following:
> generator function '*' belongs after the 'function' keyword

Akhil132-dev:
I think it would be right =>  function* myGenerator() {
  yield 'hello';
}

singalhimanshu:
What should be the error recovery in this case? Should we handle it as if it is a generator function?

strager:
> What should be the error recovery in this case? Should we handle it as if it is a generator function?

Yes, handle it as a generator function.

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

singalhimanshu:
What should be the error for this case - 
```js
*
function f(x) { yield x; }
```
Should it be the same error i.e. `error_generator_function_star_belongs_before_name` or something else?

strager:
> What should be the error for this case -
> 
> ```js
> *
> function f(x) { yield x; }
> ```

Good question.

I see three reasonable options:

1. unexpected `*`
2. missing operand for `*` (this is what is reported today)
3. generator function star belongs after `function` keyword (before function name)

I don't really care which we do. I have a slight preference for 1 or 2.
---

Unfortunately, the following code is legal, so we need to be a bit careful with when we issue a diagnostic:
```js
console.log('hi')
*function f() {}
console.log('yup')

f() // runtime error
```

PR #363, opened 2021-06-17
Report different error for *function f() {}

singalhimanshu:
The following code:

    *function f(x) { yield x; }

Should report error_generator_function_star_belongs_before_name

Fix: #267

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -2043,7 +2043,8 @@ function_attributes parser::parse_generator_star(
       QLJS_ASSERT(false);
       return function_attributes::async_generator;
     case function_attributes::generator:
-      QLJS_ASSERT(false);
+      // @@@ remove
+      // QLJS_ASSERT(false);
       return function_attributes::generator;
     case function_attributes::normal:
       return function_attributes::generator;
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -460,7 +460,7 @@ expression* parser::parse_primary_expression(precedence prec) {
             this->skip();
           }
           expression* function =
-              this->parse_function_expression(attrb, this->peek().begin);
+              this->parse_function_expression(attrb, star_token.begin);
           return function;
         }
       }
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -432,39 +432,14 @@ expression* parser::parse_primary_expression(precedence prec) {
   case token_type::kw_in:
   case token_type::question: {
     if (this->peek().type == token_type::star) {
-      lexer_transaction transaction = this->lexer_.begin_transaction();
       token star_token = this->peek();
-      this->skip();
-      if (!this->peek().has_leading_newline) {
-        bool has_leading_async = this->peek().type == token_type::kw_async;
-        function_attributes attrb = function_attributes::generator;
-        if (has_leading_async) {
-          this->skip();
-          attrb = function_attributes::async_generator;
-        }
-        if (this->peek().type == token_type::kw_function) {
-          this->skip();
-          if (this->peek().type == token_type::identifier) {
-            this->error_reporter_->report(
-                error_generator_function_star_belongs_before_name{
-                    .function_name = this->peek().span(),
-                    .star = star_token.span()});
-          } else {
-            this->error_reporter_->report(
-                error_generator_function_star_belongs_after_keyword_function{
-                    .star = star_token.span()});
-          }
-          this->lexer_.roll_back_transaction(std::move(transaction));
-          this->skip();
-          if (has_leading_async) {
-            this->skip();
-          }
-          expression* function =
-              this->parse_function_expression(attrb, star_token.begin);
-          return function;
-        }
+      std::optional<function_attributes> attrb =
+          this->try_parse_function_with_leading_star();
+      if (attrb.has_value()) {
+        expression* function =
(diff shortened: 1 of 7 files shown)


src/parse.cpp
```diff
@@ -431,6 +431,41 @@ expression* parser::parse_primary_expression(precedence prec) {
...
+                error_generator_function_star_belongs_after_keyword_function{
+                    .star = star_token.span()});
+          }
+          this->lexer_.roll_back_transaction(std::move(transaction));
+          this->skip();
+          if (has_leading_async) {
+            this->skip();
+          }
+          expression* function =
+              this->parse_function_expression(attrb, this->peek().begin);
```
strager, line 463: Minor: The span for the function expression is incorrect. It should start at the `*`. This matters when making diagnostics for code like the following:
```js
(*async function f(){}) = 42
```

We could write a test in test_parse_expression.cpp which checks the span.

```suggestion
              this->parse_function_expression(attrb, star_token.begin);
```

src/parse.cpp
```diff
@@ -431,6 +431,41 @@ expression* parser::parse_primary_expression(precedence prec) {
...
+          } else {
+            this->error_reporter_->report(
+                error_generator_function_star_belongs_after_keyword_function{
+                    .star = star_token.span()});
+          }
+          this->lexer_.roll_back_transaction(std::move(transaction));
+          this->skip();
+          if (has_leading_async) {
+            this->skip();
+          }
```
strager, line 461: These are the cases I see:

* `*` not followed by a function expression
* `*` followed by a newline
* `*async function f(){}`
* `*async function(){}`
* `*function f(){}`
* `*function(){}`

If we put lines 435-461 into a function, then the caller only cares about two cases:

* `*` not followed by a function expression, or `*` followed by a newline
* `*async function` or `*function`

For the former case, the caller doesn't need any extra information. For the latter case, the caller needs `attrb` and `star_token.begin` (see my other comment on line 463). `star_token.begin` can easily be determined by the caller (it can `this->peek()` before calling the helper function).

I think we could factor lines 435-461 into a function with the following signature:
```c++
std::optional<function_attributes> try_parse_function_with_leading_star();
```

src/quick-lint-js/parse.h
```diff
@@ -343,10 +343,51 @@ class parser {
...
+                      .star = star_token.span()});
+            }
+            this->lexer_.roll_back_transaction(std::move(transaction));
+            this->skip();
+            if (has_leading_async) {
+              this->skip();
+            }
+            this->parse_and_visit_function_declaration(
+                v, function_attributes_type,
+                /*begin=*/this->peek().begin,
```
strager, line 379: (same comment as other version)

strager (approved):

src/quick-lint-js/parse.h
```diff
@@ -2495,6 +2512,45 @@ class parser {
     }
   }
 
+  std::optional<function_attributes> try_parse_function_with_leading_star() {
```
strager, line 2515: The semantics of this function are a bit weird, so I suggest including a doc comment.

* If the function returns nullopt, no token is consumed
* If the function returns a `function_attributes`, tokens are consumed until the `kw_function` (i.e. the `*` and possibly a following `async` are skipped)

Also, including examples in the doc comment would be nice. (I notice that most functions don't have a doc comment with examples. Maybe we should change that...)

src/quick-lint-js/parse.h
```diff
@@ -2495,6 +2512,45 @@ class parser {
...
+        this->lexer_.roll_back_transaction(std::move(transaction));
+        this->skip();
+        if (has_leading_async) {
+          this->skip();
+        }
+        return attrb;
+      }
+    }
+    this->lexer_.roll_back_transaction(std::move(transaction));
+    return std::nullopt;
```
strager, line 2551: Instead of nesting `if`s so much, let's use early returns. Example (I hope this diff is readable...):
```suggestion
    if (this->peek().has_leading_newline) {
      this->lexer_.roll_back_transaction(std::move(transaction));
      return std::nullopt;
    }

      function_attributes attrb = function_attributes::generator;
      bool has_leading_async = this->peek().type == token_type::kw_async;
      // *async function f() {}
      if (has_leading_async) {
        attrb = function_attributes::async_generator;
        this->skip();
      }
      // *function f() {}
      if (this->peek().type != token_type::kw_function) {
        this->lexer_.roll_back_transaction(std::move(transaction));
        return std::nullopt;
      }

        this->skip();
        if (this->peek().type == token_type::identifier) {
          this->error_reporter_->report(
              error_generator_function_star_belongs_before_name{
                  .function_name = this->peek().span(),
                  .star = star_token.span(),
              });
        } else {
          this->error_reporter_->report(
              error_generator_function_star_belongs_after_keyword_function{
                  .star = star_token.span()});
        }
        this->lexer_.roll_back_transaction(std::move(transaction));
        this->skip();
        if (has_leading_async) {
          this->skip();
        }
        return attrb;
```

strager (approved):
sexy

Lemme know when I should merge.

src/quick-lint-js/parse.h
singalhimanshu: > Maybe we should change that

Yes

singalhimanshu:
> Lemme know when I should merge.

It is ready to merge from my side

PR #364, opened 2021-06-20
Create error message pop out box in website demo 

matttheus:
Related to #323 which was closed. 

Basically I fixed the tests error that was occurring in demo and error-documentation tests. Most of the fix was just add the message, code and severity to editor diagnostics attributes.


website/test/test-demo.mjs
```diff
@@ -51,76 +96,192 @@ describe("markEditorText", () => {
...
+        end: 5,
+        message: "unknow error",
+        code: unKnownErrorCode,
+        severity: 5,
+      },
+    ]);
+    assert.strictEqual(
+      editor.innerHTML,
+      'hello<mark data-message="unknow error" data-code="E0XXX" data-severity="5"></mark>world'
+    );
```
strager, line 236: Hmm, these tests seem fragile. What if the order of attributes changes? What if the attribute order is randomized?

Most of these tests are concerned with the *position* of the marks. We shouldn't be testing unimportant *attributes* of the marks.

An exception is tests which have multiple marks. However, your commit doesn't distinguish the marks, so having the tests check the attributes is just noise which makes it harder to understand the tests.

I see two reasonable solutions:

* Don't give message/code/severity to `markEditorText`, and have `markEditorText` omit the `data-` attributes.
* In each assertion, filter the attributes out of the HTML. We could do this with ugly string hacks, or we could do this by removing attributes with DOM APIs.

website/src/error-documentation.mjs
```diff
@@ -66,7 +66,7 @@ markdownParser.renderer.rules = {
     // Wrap BOM in a <span>.
     if (
       ["\u{feff}", "<mark>\u{feff}</mark>"].some((bom) =>
-        codeHTML.startsWith(bom)
+        codeHTML.includes(bom)
```
strager, line 69: Blocking: This is incorrect. U+FEFF *must* be at the the *beginning* to be considered a BOM. Otherwise, it is a joiner between characters. We don't want to show joiners as if they were a BOM.

strager (changes requested):

website/src/error-documentation.mjs
matttheus: Okay, I fixed, I miss understand the concept of bom

website/test/test-demo.mjs
matttheus: I create a condition to check if there is message/code/severity before create  data-attributes

strager:
Blocking: This change breaks the presentation of the BOM in the documentation for E095 (https://quick-lint-js.com/errors/#E095). I think we should run the "html wraps <mark>-d byte order mark" test with the data- attributes, since that's how the code runs in production.

website/test/test-demo.mjs
```diff
@@ -16,13 +16,23 @@ function preElementWithHTML(html) {
...
+        end: 5,
+      },
+    ]);
     assert.strictEqual(editor.innerHTML, "<mark>hello</mark> world");
   });
 
   it("mark last word on line", () => {
     let editor = preElementWithHTML("hello world");
-    markEditorText(editor, dom.window, [{ begin: 6, end: 11 }]);
+    markEditorText(editor, dom.window, [
```
strager, line 30: Nit: Undo these whitespace changes. (Maybe running `yarn fmt` will fix them all?)

strager (changes requested):

website/test/test-demo.mjs
matttheus: Fixed

website/src/error-documentation.mjs
```diff
@@ -84,6 +81,16 @@ markdownParser.renderer.rules = {
   },
 };
 
+function marksHasBOM(codeHTML) {
+  const dom = new jsdom.JSDOM(codeHTML);
+  for (const mark of dom.window.document.querySelectorAll("mark")) {
+    if (mark.textContent?.includes("\u{feff}")) {
+      return true;
+    }
+  }
```
strager, line 90: Blocking: This checks if *any* `<mark>` starts with U+FEFF. We only want to check the *first* `<mark>` iff there is no text before the first `<mark>`.

Perhaps we should inspect `container.firstChild`. If it's a text node, check if it starts with U+FEFF. If it's a `<mark>`, check if its first child text node starts with U+FEFF.

strager (changes requested):

website/src/error-documentation.mjs
matttheus: What you mean by `container.firstChild`?

website/src/error-documentation.mjs
```diff
@@ -84,6 +81,14 @@ markdownParser.renderer.rules = {
   },
 };
 
+function codeHasBOM(codeHTML) {
+  const dom = new jsdom.JSDOM(codeHTML);
+  const mark = dom.window.document.querySelector("mark");
```
matttheus, line 86: According to https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector, this selected mark will be the first one.

website/src/error-documentation.mjs
matttheus: Block fixed

website/src/error-documentation.mjs
```diff
@@ -84,6 +81,14 @@ markdownParser.renderer.rules = {
...
+  return (
+    codeHTML.startsWith("\u{feff}") || mark?.textContent.startsWith("\u{feff}")
```
strager, line 88: Blocking: This code is incorrect for the following HTML (assuming `ZWNBS` is U+FEFF):
```html
hello<mark>ZWNBSworld</mark>
```
This function should return `false` given the above HTML, but it will return `true`.

strager: Some test cases:
```html
<!-- has BOM -->
<mark>ZWNBShello</mark>
<mark data-code="E123">ZWNBShello</mark>
ZWNBShello

<!-- has no BOM -->
hello<mark>ZWNBSworld</mark>
helloZWNBSworld
```

strager (changes requested):

website/test/test-error-documentation.mjs
```diff
@@ -157,6 +157,17 @@ wasn't that neat?
       "<code><mark><span class='unicode-bom'>\u{feff}</span></mark>--BOM"
     );
   });
+
+  it("many possibilities of html code has bom", () => {
```
matttheus, line 161: Look my fancy test

website/test/test-error-documentation.mjs
```diff
@@ -157,6 +157,17 @@ wasn't that neat?
...
+    const possibilities = [
+      "<mark>\u{feff}hello</mark>",
+      '<mark data-code="E123">\u{feff}hello</mark>',
+      "hello<mark>\u{feff}world</mark>",
+    ];
+    possibilities.forEach((possibility) => {
+      expect(codeHasBOM(possibility)).toBe(true);
+    });
+  });
```
matttheus, line 170: Look my fancy test

website/src/error-documentation.mjs
matttheus: > Some test cases:
> 
> ```
> 
> <mark>ZWNBShello</mark>
> <mark data-code="E123">ZWNBShello</mark>
> ZWNBShello
> 
> 
> hello<mark>ZWNBSworld</mark>
> helloZWNBSworld
> ```

I write some tests with this cases and if I did right, this possibilities are reached.

strager (approved):
Looks good!

Please write a new PR description. Then I'll merge this PR.

matttheus:
> Looks good!
> 
> Please write a new PR description. Then I'll merge this PR.

You want I describe all functionality of the show error message, the tests and all so on modifications?

strager:
> You want I describe all functionality of the show error message, the tests and all so on modifications?

Write why you made the change and what the change is *supposed* to do.

The patch/diff already describes the modifications.

matttheus:
This pull request have intent to solve some issues in the tests caused by the implementation of the `data-` attributes inside the `<mark>` tags. Some of the issues were:

- The tests were not assigning `data-` attributes when instantiating the `<mark>` tags, so this tags were coming with undefined value assigned to them, so the testes were getting failed because the expect values in the tests have changed.
- The `code_block` function, which besides other things, checks if exists `BOM` in some HTML code passed to it, broke because the structure of the `<mark>` tag has changed (now having `data-` attributes inside) and this function was making string comparison, so was necessary change this string checking way to a `DOM` manner. With `DOM` manipulation is possible to check the content of the hole instead hard code string checking.

strager:
> This pull request have intent to solve some issues in the tests caused by the implementation of the `data-` attributes inside the `<mark>` tags.

Your PR includes much more than fixes for the tests. It adds an error box on hover. Why not mention the main feature which this PR adds?

matttheus:
> > This pull request have intent to solve some issues in the tests caused by the implementation of the `data-` attributes inside the `<mark>` tags.
> 
> Your PR includes much more than fixes for the tests. It adds an error box on hover. Why not mention the main feature which this PR adds?

Yeah sure, so I will change the PR title too, okay?

Issue #252, opened 2021-04-30
7$: Error if const variable has no initializer


strager:
quick-lint-js should report an error for the following code:
```js
const x;
```

colin1227:
Can I try this one?

strager:
@colin1227 Sure.

colin1227:
@strager 
so if I understand, if a constant variable is only declared and not assigned a value, throw an error. right?

should there error message say this?:

 "{variableName}" is declared without an assigned initial value.

also, I am new to this repo. where do you want this logic?

strager:
> @strager
> so if I understand, if a constant variable is only declared and not assigned a value, throw an error. right?

Correct.

> should there error message say this?:
> 
> "{variableName}" is declared without an assigned initial value.

Yes, that sounds good.

> also, I am new to this repo. where do you want this logic?

Probably in parser::parse_and_visit_let_bindings.

Akhil132-dev:
its a rule in js for const that JavaScript const variables must be assigned a value when they are declared.

Akhil132-dev:
can you help me to setup this project to my local machine so i can work on it

strager:
> can you help me to setup this project to my local machine so i can work on it

Did you see the BUILDING documentation? https://github.com/quick-lint/quick-lint-js/blob/9bdb827a8e7e4728f53234fef1231aa0a747f14c/docs/BUILDING.md

FYI, @colin1227 claimed this task. (Are you still hacking on it, colin?)

colin1227:
@strager yes, sorry, been busier than expected, I'll try to get something committed today

colin1227:
WIP: 
- I still have loose comments
- 3 errors about scoping, string comparison, and something else
- need to write and run tests
[but some progress](https://github.com/colin1227/quick-lint-js/tree/const-no-initalizer-252)

colin1227:
@strager will you be streaming this evening? I want to get a PR up but I need a better understanding of the function I am working in and would like to ask you some questions.

strager:
> @strager will you be streaming this evening?

No. I will next stream Monday (about 49 hours from now).

colin1227:
[new branch](https://github.com/colin1227/quick-lint-js/tree/const-252-new) based on what you mentioned about working within the `parse_and_visit_let_bindings` function instead of the linter.

1.) can't compile rn, current error: `error: array initializer must be an initializer list` line 3018 where I define a variable.
[...]

PR #365, opened 2021-06-21
Report error on uninitialized declaration of const variable.

singalhimanshu:
The following code:

    const x;

Should report error_missing_initializer_in_const_declaration

Fix: #252

### DIFF
--- a/docs/errors/E205.md
+++ b/docs/errors/E205.md
@@ -1,0 +1,11 @@
+# E205: error missing initializer in const declaration
+
+The following code is missing is missing initialization for
+const variable declaration.
+
+    const x;
+
+To fix this error, initialize the variable x with some
+value.
+
+    const x = 10;
--- a/docs/errors/E205.md
+++ b/docs/errors/E205.md
@@ -9,3 +9,7 @@ To fix this error, initialize the variable x with some
 value.
 
     const x = 10;
+
+Another way to fix this error, change const to let.
+
+    let x;
--- a/docs/errors/E205.md
+++ b/docs/errors/E205.md
@@ -1,7 +1,7 @@
 # E205: error missing initializer in const declaration
 
-The following code is missing is missing initialization for
-const variable declaration.
+The following code is missing initialization for const
+variable declaration.
 
     const x;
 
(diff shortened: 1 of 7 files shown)


docs/errors/E205.md
```diff
@@ -0,0 +1,11 @@
...
+
+The following code is missing initialization for const
+variable declaration.
+
+    const x;
+
+To fix this error, initialize the variable x with some
+value.
+
+    const x = 10;
```
strager, line 11: Alternatively, declare the variable with `let` instead.

test/test-lint-parse.cpp
```diff
@@ -85,7 +85,10 @@ TEST(test_lint,
   p.parse_and_visit_module(l);
   l.visit_end_of_module();
 
-  EXPECT_THAT(v.errors, IsEmpty());
```
strager, line 88: I think we should change `const` to `let` on line 80. The test's name tells me that `let` was intended anyway.

test/test-parse-module.cpp
```diff
@@ -42,11 +42,19 @@ TEST(test_parse, export_variable) {
   }
 
   {
-    spy_visitor v = parse_and_visit_statement(u8"export const x;"_sv);
+    padded_string code(u8"export const x;"_sv);
```
strager, line 45: Almost-blocking: Fix the input, not the assertions.
```suggestion
    padded_string code(u8"export const x = null;"_sv);
```

test/test-parse-var.cpp
```diff
@@ -92,7 +92,10 @@ TEST(test_parse, parse_simple_const) {
   ASSERT_EQ(v.variable_declarations.size(), 1);
```
strager, line 92: Almost-blocking: Fix the code, not the assertions.
```js
const x = null
```

Make a separate tests specifically for error_missing_initializer_in_const_declaration.

test/test-parse-var.cpp
```diff
@@ -861,10 +864,15 @@ TEST(test_parse, report_missing_semicolon_for_declarations) {
                     u8"x", variable_kind::_const}));
```
strager, line 864: I think we should change `const` to `let` on line 858 instead.

strager (changes requested):
Implementation looks good, but the tests are meh.

strager (approved):

Issue #214, opened 2021-04-08
7$: Disallow labels named 'await'/'yield' in async/generator functions


strager:

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

singalhimanshu:
The current error for this code:
```js
function* f() {
  yield:
}
```
Is - 
```shell
main.js:2:8: error: missing semicolon after statement [E027]
main.js:2:8: error: unexpected token [E054]
```
What should be the new errorr?

strager:
> The current error for this code:
> 
> ```js
> function* f() {
>   yield:
> }
> ```
> 
> Is -
> 
> ```shell
> main.js:2:8: error: missing semicolon after statement [E027]
> main.js:2:8: error: unexpected token [E054]
> ```
> 
> What should be the new errorr?

That looks *okay*. It could be better: perhaps "unexpected colon; expected expression or semicolon".

Note: the following code is legal:
```js
function* f() {
  true?yield:null;
}
```

PR #368, opened 2021-06-24
Disallow labels named 'await' in async functions

singalhimanshu:
The following code:

    async function f() {
      await:
    }

Should report error_label_named_await_not_allowed_in_async_function.

Fix: #214

### DIFF
--- a/test/test-parse-statement.cpp
+++ b/test/test-parse-statement.cpp
@@ -987,9 +987,52 @@ TEST(test_parse, statement_label_can_be_a_contextual_keyword) {
       EXPECT_THAT(v.visits, ElementsAre("visit_variable_use"));  // x
     }
   }
+}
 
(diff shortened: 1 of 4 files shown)


test/test-parse-statement.cpp
```diff
@@ -987,9 +987,25 @@ TEST(test_parse, statement_label_can_be_a_contextual_keyword) {
       EXPECT_THAT(v.visits, ElementsAre("visit_variable_use"));  // x
     }
   }
+}
 
-  // TODO(#214): Disallow labels named 'await' in async functions.
-  // TODO(#214): Disallow labels named 'yield' in generator functions.
```
strager, line 992: Nit: Preserve this TODO, or write a test?

strager (approved):

PR #370, opened 2021-06-25
revert .vscodeignore changes

joaomoreno:
More info: https://github.com/microsoft/vscode-vsce/issues/588

I would also highly recommend not locking vsce to a specific version. We'll do our best to maintain compatibility, as we've done so far. I didn't want to force you do revert that decision in this PR, since it's up to you, but I highly recommend it.

### DIFF
--- a/plugin/vscode-lsp/package.json
+++ b/plugin/vscode-lsp/package.json
@@ -52,6 +52,6 @@
   },
   "devDependencies": {
     "prettier": "^2.3.0",
-    "vsce": "^1.93.0"
+    "vsce": "^1.95.0"
   }
 }
--- a/plugin/vscode/.vscodeignore
+++ b/plugin/vscode/.vscodeignore
@@ -3,9 +3,6 @@
 !LICENSE
 !README.md
 !extension.js
-!node_modules/
-!node_modules/quick-lint-js-wasm/
-!node_modules/quick-lint-js-wasm/dist/
 !node_modules/quick-lint-js-wasm/dist/copyright.txt
 !node_modules/quick-lint-js-wasm/dist/quick-lint-js-vscode.wasm
 !node_modules/quick-lint-js-wasm/quick-lint-js.js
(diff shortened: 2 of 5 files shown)


joaomoreno:

plugin/vscode/.vscodeignore
```diff
@@ -3,9 +3,6 @@
 !LICENSE
 !README.md
 !extension.js
-!node_modules/
-!node_modules/quick-lint-js-wasm/
-!node_modules/quick-lint-js-wasm/dist/
```
strager, line 8: Minor: I prefer what we have now, which works with as many versions of vsce as possible.

plugin/vscode-lsp/package.json
```diff
@@ -52,6 +52,6 @@
   },
   "devDependencies": {
     "prettier": "^2.3.0",
-    "vsce": "^1.93.0"
+    "vsce": "^1.95.0"
```
strager, line 55: :+1:

strager (approved):
> I would also highly recommend not locking vsce to a specific version.

We use vsce for two workflows:

1. Packaging into a .vsix
2. Uploading to VS Code's marketplace

Do you recommend using the latest version for both workflows? My intuition is that using the latest version of vsce makes sense for the upload workflow (2), but version pinning makes sense for the packaging workflow (1).

joaomoreno:
> Do you recommend using the latest version for both workflows? My intuition is that using the latest version of vsce makes sense for the upload workflow (2), but version pinning makes sense for the packaging workflow (1).

Definitely latest for everything. We make sure vsce is up-to-date with security vulnerabilities in its dependencies, and just for that it's worth it. Plus, new features, better packaging, etc.

strager:
Thank you for reaching out to help clean up damage done by vsce's accidental breaking change. Good dev advocacy 👍

I think this PR doesn't fix anything for quick-lint-js, so I'm going to close it.

Issue #219, opened 2021-04-09
10$: Error on use of `this.#something` outside of class


strager:
Private identifiers in dot expressions (e.g. `this.#something`) are only legal inside of a class. Report an error if such an expression is found outside a class.

strager:
I think we'd need to track whether or not we're in a class declaration. We don't do this currently. (We do track whether or not we're in a function or a loop, though; perhaps we could have similar logic for classes.)

saurabhraj042:
@strager Can I work on this issue ?

strager:
@saurabhraj042 Sure!

singalhimanshu:
@saurabhraj042 Are you still working on it? If not I can work on this issue.

saurabhraj042:
@strager @singalhimanshu please give me few more days to work on this , actually I got covid last month and wasn't able to work on this but I have resumed the work for now.

singalhimanshu:
@strager I would like to work on this task.

PR #383, opened 2021-07-07
Report error on use of this.#x outside of class

singalhimanshu:
The following code:

    this.#x;

Should report `error_cannot_access_private_identifier_outside_class`.

Fix: #219

### DIFF
--- a/docs/errors/E208.md
+++ b/docs/errors/E208.md
@@ -7,16 +7,17 @@ the class.
         #x = 10;
     }
 
-    function f() {
-      this.#x = 20;
+    function f(c) {
+      c.#x = 20;
     }
 
-To fix this error, move function `f` as a member of class `C`.
+To fix this error, move function `f` as a static member of
+class `C` .
 
     class C {
         #x = 10;
-        f() {
-          this.#x = 20;
+        static f(c) {
+          c.#x = 20;
         }
     }
 
@@ -24,9 +25,9 @@ Another way to fix this error, remove `#` before variable
 `x`.
 
     class C {
-        #x = 10;
+        x = 10;
     }
 
-    function f() {
-      this.x = 20;
+    function f(c) {
+      c.x = 20;
     }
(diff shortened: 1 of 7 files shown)


docs/errors/E208.md
```diff
@@ -0,0 +1,32 @@
...
+The following code is accessing a private identifier outside
+the class.
+
+    class C {
+        #x = 10;
+    }
+
+    function f() {
+      this.#x = 20;
+    }
```
strager, line 12: Minor: A better example would be accepting an object as a parameter, not using `this`.

docs/errors/E208.md
```diff
@@ -0,0 +1,32 @@
...
+
+To fix this error, move function `f` as a member of class `C`.
+
+    class C {
+        #x = 10;
+        f() {
```
strager, line 18: A direct analog would be making `f` `static`, right?

docs/errors/E208.md
```diff
@@ -0,0 +1,32 @@
...
+          this.#x = 20;
+        }
+    }
+
+Another way to fix this error, remove `#` before variable
+`x`.
+
+    class C {
+        #x = 10;
```
strager, line 27: Wouldn't they want to remove `#` here too?

strager (changes requested):
Blocking: This PR seems to not work with class expressions:
```js
let C = class {
  #field = 42;
  banana() {
    console.log(this.#field);  // false positive E208
  }
}
new C().banana();
```

src/parse.cpp
```diff
@@ -374,6 +374,7 @@ expression* parser::parse_primary_expression(precedence prec) {
 
   // class {}
   case token_type::kw_class: {
+    class_guard g(this, std::exchange(this->in_class_, true));
     expression* class_expression = this->parse_class_expression();
```
strager, line 378: Could we put the `class_guard` inside `parse_and_visit_class_body` instead to avoid repeating the `class_guard` code in the callers?

strager (approved):

Issue #367, opened 2021-06-24
15$: Prune old builds on c.quick-lint-js.com


strager:
c.quick-lint-js.com stores build artifacts created by CI scripts (in the .github/ directory in the repo). It maintains builds for closed PRs and deleted branches.

Idea: Delete builds from c.quick-lint-js.com which:

* are older than 14 days,
* are referenced by no branch, and
* are referenced by no pull request

This will save us some server storage costs.

matttheus:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
Builds: https://c.quick-lint-js.com/builds/

If we're going to run the script automatically via GitHub Actions, here are some options:

* SFTP (this is what GitHub Actions uses to upload the builds)
* SSH, run some Bash scripts on c.quick-lint-js (this is what `dist/debian/sync-releases-to-apt` does)

If we're going to run the script via a cron job on the server, we need to access Git commit metadata.

matttheus:
Working on #385. You can follow my work there.

strager:
I deployed @matttheus' script. It's running hourly via cron for the github-ci user on c.quick-lint-js.com.

PR #385, opened 2021-07-11
Create script to prune old builds (#367)

matttheus:
With this pull request I pretend to have a script which prune older builds that fit the following requirements:

    * are older than 14 days
    * are referenced by no branch
    * are referenced by no pull request

### DIFF
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -30,7 +30,7 @@ def get_commits(repo_name: str, repository_url: str) -> list:
     try:
         return subprocess.check_output(
             f"cd {repo_name} \
-            && git fetch --prune {pipes.quote(repository_url)} '+refs/pull/*/head:refs/remotes/github-pr/*' \
+            && git fetch --prune --all \
             && git rev-list --all --remotes", shell=True
         ).decode('utf-8').split('\n')
     except subprocess.CalledProcessError as err:
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -57,7 +57,7 @@ if __name__ == '__main__':
                         help="Delete folders.")
     args, _ = parser.parse_known_args()
 
-    builds_path = Path(pipes.quote(args.builds_path))
+    builds_path = Path(args.builds_path)
     repository_url = pipes.quote(args.repository_url)
     repo_name = pipes.quote(args.repo_name)
 
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -29,8 +29,8 @@ def error_print(*args, **kwargs):
 def get_commits(repo_name: str, repository_url: str) -> list:
     try:
         return subprocess.check_output(
-            f"cd {repo_name} \
-            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/prs/*' '+refs/heads/*:refs/remotes/origin/*' \
+            f"cd {pipes.quote(repo_name)} \
+            && git fetch --prune {pipes.quote(repository_url)} '+refs/pull/*/head:refs/remotes/prs/*' '+refs/heads/*:refs/remotes/origin/*' \
             && git rev-list --all --remotes", shell=True
         ).decode('utf-8').split('\n')
     except subprocess.CalledProcessError as err:
@@ -58,8 +58,8 @@ if __name__ == '__main__':
     args, _ = parser.parse_known_args()
 
     builds_path = Path(args.builds_path)
-    repository_url = pipes.quote(args.repository_url)
-    repo_name = pipes.quote(args.repo_name)
+    repository_url = args.repository_url
+    repo_name = args.repo_name
 
     if not os.path.exists(builds_path):
         error_print(
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -29,8 +29,8 @@ def error_print(*args, **kwargs):
 def get_commits(repo_name: str, repository_url: str) -> list:
     try:
         return subprocess.check_output(
-            f"cd {pipes.quote(repo_name)} \
-            && git fetch --prune --all \
+            f"cd {repo_name} \
+            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/prs/*' '+refs/heads/*:refs/remotes/origin/*' \
             && git rev-list --all --remotes", shell=True
         ).decode('utf-8').split('\n')
     except subprocess.CalledProcessError as err:
@@ -57,9 +57,9 @@ if __name__ == '__main__':
                         help="Delete folders.")
     args, _ = parser.parse_known_args()
 
-    builds_path = Path(args.builds_path)
-    repository_url = args.repository_url
-    repo_name = args.repo_name
+    builds_path = Path(pipes.quote(args.builds_path))
+    repository_url = pipes.quote(args.repository_url)
+    repo_name = pipes.quote(args.repo_name)
 
     if not os.path.exists(builds_path):
         error_print(
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -30,8 +30,7 @@ def get_commits(repo_name: str, repository_url: str) -> list:
         return subprocess.check_output(
             f"cd {repo_name} \
             && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/github-pr/*' \
-            && git rev-list --all --remotes",
-            stderr=subprocess.STDOUT, shell=True
+            && git rev-list --all --remotes", shell=True
         ).decode('utf-8').split('\n')
     except subprocess.CalledProcessError as err:
         error_print(err.output.decode())
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -23,6 +23,19 @@ def error_print(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
 
 
+def get_commits(repo_name: str, repository_url: str) -> list:
+    try:
+        return subprocess.check_output(
+            f"cd {repo_name} \
+            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/github-pr/*' \
+            && git rev-list --all --remotes",
+            stderr=subprocess.STDOUT, shell=True
+        ).decode('utf-8').split('\n')
+    except subprocess.CalledProcessError as err:
+        error_print(err.output.decode())
+        exit(err.returncode)
+
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser("Prune old builds")
     parser.add_argument(
@@ -50,16 +63,12 @@ if __name__ == '__main__':
         exit(1)
 
     if not os.path.exists(repo_name):
-        print("Cloning repository")
-        subprocess.run(['git', 'clone', '-q', repository_url])
+        error_print("Cloning repository")
+        subprocess.check_call(
+            ['git', 'clone', '-q', '--bare', repository_url, repo_name])
 
-    print('Getting all commits of the repository\n')
-    commits = subprocess.getoutput(
-        f"cd {repo_name} \
-        && git pull origin master \
-        && git fetch origin 'refs/pull/*/head:refs/remotes/github-pr/*' \
-        && git rev-list --all --remotes"
-    ).split('\n')
+    error_print('Getting all commits of the repository\n')
+    commits = get_commits(repo_name, repository_url)
 
     to_be_deleted = []
     for folder in os.listdir(builds_path):
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2020  Matthew "strager" Glazar
 # See end of file for extended copyright information.
 
+import pipes
 import re
 import os
 import sys
@@ -29,7 +30,7 @@ def get_commits(repo_name: str, repository_url: str) -> list:
     try:
         return subprocess.check_output(
             f"cd {repo_name} \
-            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/github-pr/*' \
+            && git fetch --prune {pipes.quote(repository_url)} '+refs/pull/*/head:refs/remotes/github-pr/*' \
             && git rev-list --all --remotes", shell=True
         ).decode('utf-8').split('\n')
     except subprocess.CalledProcessError as err:
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -1,3 +1,8 @@
+#!/usr/bin/python3
+
+# Copyright (C) 2020  Matthew "strager" Glazar
+# See end of file for extended copyright information.
+
 import os
 import shutil
 import argparse
@@ -61,3 +66,21 @@ if __name__ == '__main__':
                 shutil.rmtree(folder_path)
             except OSError as e:
                 print("Error: %s : %s" % (folder_path, e.strerror))
+
+# quick-lint-js finds bugs in JavaScript programs.
+# Copyright (C) 2020  Matthew "strager" Glazar
+#
+# This file is part of quick-lint-js.
+#
+# quick-lint-js is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# quick-lint-js is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with quick-lint-js.  If not, see <https://www.gnu.org/licenses/>.
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -19,6 +19,10 @@ def how_old(date: datetime) -> int:
     return (datetime.now() - date).days
 
 
+def error_print(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser("Prune old builds")
     parser.add_argument(
--- a/tools/prune-old-builds.py
+++ b/tools/prune-old-builds.py
@@ -1,50 +1,63 @@
 import os
 import shutil
+import argparse
 import subprocess
 from pathlib import Path
 from datetime import datetime
 
+
 FOURTEEN_DAYS = 14
 
 
-def how_old(date: str) -> int:
+def how_old(date: datetime) -> int:
     """Return a integer that represents how many days has passed given a date"""
-    return (datetime.now() - datetime.strptime(date, '%Y-%m-%d')).days
+    return (datetime.now() - date).days
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


matttheus:
The base idea of the script gonna work is already done, the code need performance improvements and I need figure out how to check the third requirement which check if the build is already related to some pull request, to do so, I guess I need implementation with github api.

matttheus:
Maybe I can try this approach here https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally, to create branchs from PRs locally the check the commits. 

To get the PRs IDs I can search using the API of the GitHub to list the opened ones. What you think of this approach?

matttheus:
Give me a overview of the code I did, I guess this can solve the problem, but I don't know if is exactly what you want. Let me know you want a more configurable script.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+        folder_path = builds_path / folder
+        created_at = datetime.fromtimestamp(
+            os.path.getctime(folder_path)).strftime('%Y-%m-%d')
+
+        if not folder in commits:
+            print(
+                f'The build of hash {folder} is not related to a branch anymore and will be deleted')
+            shutil.rmtree(folder_path)
+
+        if how_old(created_at) > FOURTEEN_DAYS:
```
strager, line 41: These criteria should be **and**ed together, not **or**ed.

If a build is old **and** a build's commit isn't part of the repo/PR.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+
+if __name__ == '__main__':
+    builds_path = Path('builds/')
+    repo_url = 'https://github.com/quick-lint/quick-lint-js.git'
+    repo_name = 'quick-lint-js'
+
+    if os.path.exists(repo_name):
+        shutil.rmtree(repo_name)
+
+    subprocess.run(['git', 'clone', '-q', repo_url])
```
strager, line 23: It seems wasteful that we download and delete the repo every time we run prune-old-builds. I think we should cache the download.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for folder in os.listdir(builds_path):
+        folder_path = builds_path / folder
+        created_at = datetime.fromtimestamp(
+            os.path.getctime(folder_path)).strftime('%Y-%m-%d')
+
+        if not folder in commits:
+            print(
+                f'The build of hash {folder} is not related to a branch anymore and will be deleted')
+            shutil.rmtree(folder_path)
```
strager, line 39: I think we should have the script print out the directories to delete, leaving deletion as a separate step. At the very least, this should be an option (like `git clean`'s `-n` vs `-f`).

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+
+    subprocess.run(['git', 'clone', '-q', repo_url])
+
+    print('Getting all commits of the repository')
+    commits = subprocess.getoutput(
+        f"cd {repo_name} && git rev-list --all --remotes").split('\n')
+
+    builds_path = Path('builds/')
+
+    for folder in os.listdir(builds_path):
```
strager, line 31: It looks like this code runs on the server. How do you think we should deploy and run this script?

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+FOURTEEN_DAYS = 14
+
+
+def how_old(date: str) -> int:
+    """Return a integer that represents how many days has passed given a date"""
+    return (datetime.now() - datetime.strptime(date, '%Y-%m-%d')).days
+
+
+if __name__ == '__main__':
+    builds_path = Path('builds/')
```
strager, line 16: The script should accept this path as an argument. Same with the Git repo checkout path.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,50 @@
...
+    print('Getting all commits of the repository')
+    commits = subprocess.getoutput(
+        f"cd {repo_name} && git rev-list --all --remotes").split('\n')
+
+    builds_path = Path('builds/')
+
+    for folder in os.listdir(builds_path):
+        folder_path = builds_path / folder
+        created_at = datetime.fromtimestamp(
+            os.path.getctime(folder_path)).strftime('%Y-%m-%d')
```
strager, line 34: Why do we convert the `datetime` into a `str`, only to parse it back into a `datetime` (line 12)?

strager (commented):
> Maybe I can try this approach here https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally, to create branchs from PRs locally the check the commits.
> 
> To get the PRs IDs I can search using the API of the GitHub to list the opened ones. What you think of this approach?

Even simpler: fetch all the PRs using Git:
```
$ git fetch origin 'refs/pull/*/head:refs/remotes/github-pr/*'
```

No need to interact with GitHub's API for this. (This is what I do on my machines.)

matttheus:
> > Maybe I can try this approach here https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally, to create branchs from PRs locally the check the commits.
> > To get the PRs IDs I can search using the API of the GitHub to list the opened ones. What you think of this approach?
> 
> Even simpler: fetch all the PRs using Git:
> 
> ```
> $ git fetch origin 'refs/pull/*/head:refs/remotes/github-pr/*'
> ```
> 
> No need to interact with GitHub's API for this. (This is what I do on my machines.)

Should I consider even the closed ones?

tools/prune-old-builds.py
matttheus: We could run this script using cron or keep this script in the server and run it when the pipeline get triggered through ssh.

tools/prune-old-builds.py
matttheus: I don't understand what you mean. Are you suggesting I change this line 39 and put at end of the script?

tools/prune-old-builds.py
matttheus: With cache you mean keep the repository in the server, right?

matttheus:
Sample script output:

image

strager:
> Should I consider even the closed ones?

That's a good question.

Yes, let's keep the builds for the closed PRs.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
+import os
```
strager, line 1: Nit: Put a short copyright notice at the top of the file too. (See other files for examples.)

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+        exit(1)
+
+    if not os.path.exists(repo_name):
+        print("Cloning repository")
+        subprocess.run(['git', 'clone', '-q', repository_url])
+
+    print('Getting all commits of the repository\n')
+    commits = subprocess.getoutput(
+        f"cd {repo_name} \
+        && git pull origin master \
```
strager, line 48: There is no need to `pull`. We only need the `fetch` command.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+        && git fetch origin 'refs/pull/*/head:refs/remotes/github-pr/*' \
```
strager, line 49: We should use `--prune` to remove old branches from our local set.

We should use `+` to update branches which are force-pushed.

```suggestion
        && git fetch --prune origin '+refs/pull/*/head:refs/remotes/github-pr/*' \
```

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+    if not os.path.exists(builds_path):
+        print(
+            f"error: The directory \033[1m{builds_path}\033[0m does not exists.")
+        exit(1)
+
+    if not os.path.exists(repo_name):
+        print("Cloning repository")
+        subprocess.run(['git', 'clone', '-q', repository_url])
+
+    print('Getting all commits of the repository\n')
```
strager, line 45: Nit: Write diagnostic messages to standard error, not standard output.

tools/prune-old-builds.py
strager: > Are you suggesting I change this line 39 and put at end of the script?

Yes.

1. Figure out which folders to delete
2. Delete all the folders (if requested with `-f`)

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+        folder_path = builds_path / folder
+        created_at = datetime.fromtimestamp(os.path.getctime(folder_path))
+
+        if not folder in commits and how_old(created_at) > FOURTEEN_DAYS:
+            print(
+                f"""Build \033[93m{folder}\033[0m accomplish the following criteria:\n* Is not part of the repo/PR\n* Has more then fourteen days\n\033[1;31mThis build will be deleted.\033[0m\n""")
+            try:
+                shutil.rmtree(folder_path)
+            except OSError as e:
+                print("Error: %s : %s" % (folder_path, e.strerror))
```
strager, line 63: If deleting any folder fails, the script should exit with a non-zero status code. This will indicate to cron/whatever that there is a problem.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+        && git pull origin master \
+        && git fetch origin 'refs/pull/*/head:refs/remotes/github-pr/*' \
+        && git rev-list --all --remotes"
+    ).split('\n')
+
+    for folder in os.listdir(builds_path):
+        folder_path = builds_path / folder
+        created_at = datetime.fromtimestamp(os.path.getctime(folder_path))
+
+        if not folder in commits and how_old(created_at) > FOURTEEN_DAYS:
```
strager, line 57: We should only consider deleting folders which look like commit hashes (40 characters 0-9a-f). If there's a folder which doesn't look like a commit hash, we should warn about it, but never delete it.

tools/prune-old-builds.py
strager, line 49: We should use the given repository URL instead of assuming that `origin` is valid. E.g. `git fetch https://github.com/quick-lint/quick-lint-js.git refs/...`

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+        print(
+            f"error: The directory \033[1m{builds_path}\033[0m does not exists.")
+        exit(1)
+
+    if not os.path.exists(repo_name):
+        print("Cloning repository")
+        subprocess.run(['git', 'clone', '-q', repository_url])
+
+    print('Getting all commits of the repository\n')
+    commits = subprocess.getoutput(
```
strager, line 46: `subprocess.getoutput` has a few problems:

* It mixes error output with normal output into the returned string. This means that network errors won't be visible to the human at all.
* **Must fix**: If the command fails, the script proceeds as if nothing bad happens. This is very bad. If a network error occurs, then `git rev-list --all --remotes` won't run. If `git rev-list --all --remotes` doesn't run, no commits end up in the output. If no commits end up in the output, we'll attempt to delete our archive of old builds!
* `subprocess.getoutput` always invokes a shell. This makes it harder to safely inject arguments (such as `repo_name` or `repository_url`.

Use `subprocess.check_output` with a list instead, which avoids all these problems. https://docs.python.org/3/library/subprocess.html?highlight=subprocess#subprocess.check_output

tools/prune-old-builds.py
strager: That makes sense.

I guess updates to this script need to be manually deployed (i.e. copied from my machine to the server). I think that's fine.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,81 @@
...
+    repo_name = 'quick-lint-js'
+
+    if not os.path.exists(builds_path):
+        print(
+            f"error: The directory \033[1m{builds_path}\033[0m does not exists.")
+        exit(1)
+
+    if not os.path.exists(repo_name):
+        print("Cloning repository")
+        subprocess.run(['git', 'clone', '-q', repository_url])
```
strager, line 43: We should make a bare clone to avoid unnecessary disk usage.
```suggestion
        subprocess.run(['git', 'clone', '-q', '--bare', repository_url])
```

strager, line 43: Specify the destination directory explicitly to prevent changes to the URL from changing the destination:
```suggestion
        subprocess.run(['git', 'clone', '-q', repository_url, repo_name])
```

strager, line 43: If `git clone` fails, you're ignoring the failure. Is this intentional? If so, add a comment. If not, use `subprocess.check_call` instead to raise an exception if `git clone` fails.

strager (changes requested):

strager:
Hi @matttheus, are you still working on this patch?

matttheus:
> Hi @matttheus, are you still working on this patch?

Hi @strager, yes. I've been overload of degree work. I'll make updates this week, sorry.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,112 @@
...
+    print(*args, file=sys.stderr, **kwargs)
+
+
+def get_commits(repo_name: str, repository_url: str) -> list:
+    try:
+        return subprocess.check_output(
+            f"cd {repo_name} \
+            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/github-pr/*' \
+            && git rev-list --all --remotes",
+            stderr=subprocess.STDOUT, shell=True
```
strager, line 33: Because you mix stderr and stdout, if `git fetch` (or `git rev-list`) prints any warning messages, those messages end up in `get_commits`'s return value. Let's keep stderr and stdout separate.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,112 @@
...
+        help="The url to the repository.",
+        nargs=1,
+        type=str)
+    parser.add_argument("-f", "--delete", action="store_true",
+                        help="Delete folders.")
+    args, _ = parser.parse_known_args()
+
+    builds_path = Path(args.builds_path[0])
+    # https://github.com/quick-lint/quick-lint-js.git
+    repository_url = args.repository_url[0]
```
strager, line 58: Nit: `nargs` causes `argparse` to give you lists. If you omit `nargs`, you get just the `str`. Remove `nargs`:
```suggestion
    parser.add_argument(
        "builds_path",
        help="The path where the builds are stored.",
        type=str)
    parser.add_argument(
        "repository_url",
        help="The url to the repository.",
        type=str)
    parser.add_argument("-f", "--delete", action="store_true",
                        help="Delete folders.")
    args, _ = parser.parse_known_args()

    builds_path = Path(args.builds_path)
    # https://github.com/quick-lint/quick-lint-js.git
    repository_url = args.repository_url
```

tools/prune-old-builds.py
```diff
@@ -0,0 +1,112 @@
...
+    repo_name = 'quick-lint-js'
```
strager, line 59: Minor: I think this should be customizable.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,112 @@
...
+
+def error_print(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+
+def get_commits(repo_name: str, repository_url: str) -> list:
+    try:
+        return subprocess.check_output(
+            f"cd {repo_name} \
+            && git fetch --prune {repository_url} '+refs/pull/*/head:refs/remotes/github-pr/*' \
```
strager, line 31: If `repository_url` contains characters interpreted by the shell, this will do funky things.

I recommend avoiding `shell=True`. If we can't avoid it, shell-escape the string:
```suggestion
            && git fetch --prune {pipes.quote(repository_url)} '+refs/pull/*/head:refs/remotes/github-pr/*' \
```

strager, line 31: Must fix: This only pulls PRs, not branches. We shouldn't delete a build if its commit exists on a branch (e.g. `master`).

strager (changes requested):

tools/prune-old-builds.py
matttheus: What level of customization do you want? I mean, do you want a configuration file? Or do you want the `repo_name` come from args too.

strager: `repo_name` should come from the command-line arguments, similar to `builds_path` and `repository_url`.

matttheus: All right

strager (approved):

tools/prune-old-builds.py
```diff
@@ -0,0 +1,115 @@
...
+
+def error_print(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+
+def get_commits(repo_name: str, repository_url: str) -> list:
+    try:
+        return subprocess.check_output(
+            f"cd {pipes.quote(repo_name)} \
+            && git fetch --prune --all \
```
strager, line 33: Must fix: This doesn't seem to fetch PRs. For example, your PR isn't fetched:
```
$ git show 791e3fa12f4bcf12999a9ac5e8dde09a6436d36d
fatal: bad object 791e3fa12f4bcf12999a9ac5e8dde09a6436d36d
```

strager (changes requested):

tools/prune-old-builds.py
strager, line 33: The repository URL (given by the user) should be used every fetch, not just on the initial clone.

tools/prune-old-builds.py
```diff
@@ -57,9 +57,9 @@ def has_valid_name(folder_name: str) -> bool:
                         help="Delete folders.")
     args, _ = parser.parse_known_args()
 
-    builds_path = Path(args.builds_path)
-    repository_url = args.repository_url
-    repo_name = args.repo_name
+    builds_path = Path(pipes.quote(args.builds_path))
```
strager, line 60: Must fix: `pipes.quote` can change the string. (For example, it adds single quotes around a string if the string contains any spaces.) We only want to use `pipes.quote` when writing a shell script, not when calling functions such as `os.path.exists`.

strager (changes requested):

tools/prune-old-builds.py
```diff
@@ -0,0 +1,115 @@
...
+    builds_path = Path(args.builds_path)
+    repository_url = pipes.quote(args.repository_url)
+    repo_name = pipes.quote(args.repo_name)
+
+    if not os.path.exists(builds_path):
+        error_print(
+            f"error: The directory \033[1m{builds_path}\033[0m does not exists.")
+        exit(1)
+
+    if not os.path.exists(repo_name):
```
strager, line 69: Must fix: `repo_name` is shell-escaped thus can't be given to `os.path.exists`. Use the unescaped string here.

tools/prune-old-builds.py
```diff
@@ -0,0 +1,115 @@
...
+        error_print("Cloning repository")
+        subprocess.check_call(
+            ['git', 'clone', '-q', '--bare', repository_url, repo_name])
```
strager, line 72: Must fix: `repository_url` and `repo_name` are shell-escaped thus shouldn't be given to `subprocess.check_call` in a list. Use the unescaped string here.

strager (changes requested):
You are using `pipes.quote` incorrectly. Use `pipes.quote` only when creating a shell command.

matttheus:
> You are using `pipes.quote` incorrectly. Use `pipes.quote` only when creating a shell command.

I guess I fixed it

strager:
Looks good! I deployed it. It's running hourly (via cron for the github-ci user). Shrank 18 GiB down to 9.7G GiB. Very nice.

PR #391, opened 2021-07-17
update emacs plugins error parsing

wgrr:
notes:

I didn't ran asciidoctor to generate manpages and html because my version (debian sid) is different than previously used to generate, and it makes a diff-mess in both man and html, so left that out of the patches

emacs-location and utf8 code point counting looks awkward, i feel it could handle the case where offset is in middle of a code point by going backward, altought im unsure if i should

the exit_fail_on is ignored in main when output_format is emacs_lisp because it's the easier path i saw, im unsure if there is a better place for that, it needs to return 0 for emacs so flymake (perhaps flycheck too) can disable quick-lint-js backend if some input is crashing quick-lint-js, it avoids inifnite loops spawning quick-lint-js

### DIFF
--- a/plugin/emacs/flycheck-quicklintjs.el
+++ b/plugin/emacs/flycheck-quicklintjs.el
@@ -25,9 +25,10 @@
 ;;   ;; The default is 0.5 (500ms)
 ;;   (setq-local flycheck-idle-change-delay 0)
 ;;
-;;   ;; Run quick-lint-js program when the buffer is changed and when 'js-mode` is
-;;   ;; loaded
-;;   (setq-local flycheck-check-syntax-automatically '(mode-enabled idle-change)))
+;;   ;; Run quick-lint-js program when the buffer is changed and when 'js-mode`
+;;   ;; is loaded
+;;   (setq-local flycheck-check-syntax-automatically
+;;               '(mode-enabled idle-change)))
 ;; (add-hook 'js-mode-hook #'my-flycheck-quicklintjs-setup)
 ;;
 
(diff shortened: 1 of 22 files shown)


src/emacs-lisp-error-reporter.cpp
```diff
@@ -0,0 +1,92 @@
...
+    : output_(output), locator_(locator), code_(code) {}
+
+void emacs_lisp_error_formatter::write_before_message(
+    severity sev, const source_code_span &origin) {
+  std::string_view severity_type{};
+  if (sev == severity::note) {
+    return;
+  }
+  emacs_source_range r = this->locator_.range(origin);
+  // Emacs point starts at 1
```
strager, line 51: If we need to offset by 1, we should do this in `emacs_locator`, not here.

src/emacs-lisp-error-reporter.cpp
```diff
@@ -0,0 +1,92 @@
...
+                << " "
+                << "\"" << this->code_ << "\" \"";
+}
+
+void emacs_lisp_error_formatter::write_message_part(severity sev,
+                                                    string8_view message) {
+  if (sev == severity::note) {
+    return;
+  }
+  this->output_ << out_string8(message);
```
strager, line 64: Almost-blocking: What if the message contains `"`? We should escape `"` and possibly other characters, right?

test/test-emacs-lisp-error-reporter.cpp
```diff
@@ -0,0 +1,164 @@
...
+  source_code_span parenthesis_span(&input[4 - 1], &input[4 - 1]);
+  emacs_lisp_error_reporter reporter = this->make_reporter(&input);
+  reporter.report(error_expected_parenthesis_around_if_condition{
+      .where = parenthesis_span,
+      .token = '(',
+  });
+  reporter.finish();
+  EXPECT_EQ(this->get_output(),
+            "(((4 . 4) 0 \"E018\" \"if statement is missing '(' around "
+            "condition\"))");
```
strager, line 76: Nit: We can avoid backslashes by using a raw string literal:
```suggestion
            R"--((((4 . 4) 0 "E018" "if statement is missing '(' around condition")))--");
```
(But maybe this is more confusing because of the parentheses...)

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -45,23 +46,34 @@
...
+            (msg (nth 3 l)))
+        (push (flycheck-error-new-at-pos
+               (car region)
+               (if (= sev 0) 'error 'warning)
+               msg
+               :id code
+               :buffer buffer
+               :checker checker
+               :end-pos (cdr region)) err-accum)))
+    err-accum))
```
strager, line 65: Is there a `map` function we can use instead of manually `push`ing onto a temporary list? Something like this:
```suggestion
  (map (car (read-from-string output)) (lambda (l)
      (let ((region (nth 0 l))
            (sev (nth 1 l))
            (code (nth 2 l))
            (msg (nth 3 l)))
        (push (flycheck-error-new-at-pos
               (car region)
               (if (= sev 0) 'error 'warning)
               msg
               :id code
               :buffer buffer
               :checker checker
               :end-pos (cdr region)) err-accum))))
```

strager (changes requested):
Looks awesome!

See almost-blocking comment. Otherwise, I'm pretty happy with this.

wgrr:
updated:
fix ci build
rewrite to use `mapcar` instead of `dolist`
attempt to make string literal to fit in < 80 chars less grose
escape " and \
offset by one in locator instead of reporter
use c++ raw string syntax

test/test-emacs-lisp-error-reporter.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include <quick-lint-js/emacs-location.h>
+#include <quick-lint-js/location.h>
+#include <quick-lint-js/padded-string.h>
+#include <sstream>
+
+namespace quick_lint_js {
+namespace {
+class test_emacs_lisp_error_reporter : public ::testing::Test {
+ protected:
+  emacs_lisp_error_reporter make_reporter() {
```
strager, line 17: This function is unused, according to the failing CI build.

test/test-emacs-lisp-error-reporter.cpp
```diff
@@ -0,0 +1,188 @@
...
+  padded_string input(u8"import { x } ;"_sv);
+  source_code_span span(&input[12], &input[12]);
+
+  emacs_lisp_error_reporter reporter = this->make_reporter(&input);
+  reporter.report(error_expected_from_and_module_specifier{span});
+  reporter.finish();
+  EXPECT_EQ(
+      this->get_output(),
+      R"--((((13 . 13) 0 "E129" "expected 'from \"name_of_module.mjs\"'")))--");
+}
```
strager, line 168: :+1:

strager (approved):

wgrr:
@strager ok, now i think my battle with CI is over, there was more linting issues with gcc besides clang, i had to setup an environment to run gcc-8, i cant build with `-Wall -Werror` because gcc-10/11 freaks out about shadowing variables in parser.h, let me know if you spot more issues.

wgrr:
@strager lgtm

PR #394, opened 2021-07-22
Emacs: use correct quick-lint-js.config file

strager:
@wgrr

### DIFF
--- a/plugin/emacs/flycheck-quicklintjs.el
+++ b/plugin/emacs/flycheck-quicklintjs.el
@@ -68,6 +68,10 @@
 https://quick-lint-js.com"
   :command ("quick-lint-js"
             "--output-format=emacs-lisp"
+            (eval (let ((file (buffer-file-name)))
+                    (if file
+                      `("--path-for-config-search" ,file)
+                      ())))
             "--stdin"
             (eval flycheck-quicklintjs-args))
   :standard-input t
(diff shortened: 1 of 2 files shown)


plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -68,6 +68,7 @@
 https://quick-lint-js.com"
   :command ("quick-lint-js"
             "--output-format=emacs-lisp"
+            "--path-for-config-search" (eval buffer-file-name)
```
wgrr, line 71: maybe blocking: in emacs, a buffer is just utf8 encoded data in memory, and it's not necessarily tied to a file, thus buffer-file-name would return nil in such situation, however, this is a rare case, i personally don't know any workflow that would end up in such scenario, a way to reproducing such case:
`C-x b tmpbuf RET RET M-x js-mode`

my concerning with this is flycheck dont handling `nil` in this case, which would crash the plugin, or if it handles as an empty string, quick-lint-js fails would fail to run due `--path-for-config-search` without a value, right?

strager: Good point! (Vim has the same thing. I should see how we handle this case in Vim...)

wgrr:
```suggestion
            (eval (if-let ((bfn (buffer-file-name))) `("--path-for-config-search" ,bfn) ""))
```
i bet this wouldnt work because we are already quoted by flycheck, but something in this lines

wgrr (approved):
lgtm

PR #401, opened 2021-07-22
fix #400: add semicol to generate an error in parse_expression_remainder

### LINKED ISSUE #400: parser crashing with IDENT DOT SEMICOL
this was found with emacs default scrach buffer message:
```
;; This buffer is for text that is not saved, and for Lisp evaluation.
;; To create a file, visit it with C-x C-f and enter text in its buffer.
```

quick-lint-js version commit: 91f23fd86fa67d9e5d4ce859b37f1fad206525ca

simpler input file
```
% cat /dev/shm/test
die.;
% 
```

quick-lint-js output:
```
% quick-lint-js /dev/shm/test
/home/w/g/quick-lint-js/src/parse.cpp:957: fatal: token not implemented in parse_expression_remainder: semicolon on line 1 column 5
% 
```

wgrr:
first time touching the parser code, i hope i didn't screw up too much, i have a bad feeling about my second test case, i think errors should stop in `error_missing_operand_for_operator` and ignore `error_missing_property_name_for_dot_operator`. im spleepy atm so i couldn't find an easy path for doing that, so pardon me it should.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -942,6 +942,7 @@ next:
     case token_type::minus:
     case token_type::plus:
     case token_type::question:
+    case token_type::semicolon:
     case token_type::right_paren: {
       source_code_span empty_property_name(dot_span.end(), dot_span.end());
       children.back() = this->make_expression<expression::dot>(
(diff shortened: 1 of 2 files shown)


strager:
> i have a bad feeling about my second test case, i think errors should stop in `error_missing_operand_for_operator` and ignore `error_missing_property_name_for_dot_operator`.

Ideally we'd make another error for `.;` ("unexpected dot" or something). But two errors is fine for now. Other operators have a similar issue:

```js
*;
```
```
hello.js:1:1: error: missing operand for operator [E026]
hello.js:1:1: error: missing operand for operator [E026]
```

Issue #405, opened 2021-07-26
internal check failure with FUNC OPAREN CPAREN


wgrr:
with current master, the input `funcion()` causes an internal assertion failure

```
% echo 'function()' | quick-lint-js --stdin
/home/w/g/quick-lint-js/src/./quick-lint-js/parse.h:1077: internal check failed in parse_and_visit_statement_block_no_scope: this->peek().type == token_type::left_curly
Illegal instruction
%
```

PR #408, opened 2021-08-01
fix for #405

wgrr:


src/quick-lint-js/parse.h
```diff
@@ -1290,12 +1287,32 @@ class parser {
...
+    if (this->peek().type != token_type::left_curly) {
+      if (!name.has_value()) {
+        QLJS_PARSER_UNIMPLEMENTED();
+      }
+      this->error_reporter_->report(error_missing_function_body{
+          .function_name = *name,
+      });
+      return;
+    }
+    this->parse_and_visit_function_body_no_scope(v);
```
strager, line 1313: Nit: `no_scope` in the name `parse_and_visit_function_body_no_scope` is confusing. (It's also confusing for `parse_and_visit_function_parameters_and_body_no_scope`, but that's not your fault.) How about we inline the function to avoid the confusing function name entirely:
```suggestion
    v.visit_enter_function_scope_body();
    this->parse_and_visit_statement_block_no_scope(v);
```

src/quick-lint-js/parse.h
```diff
@@ -1290,12 +1287,32 @@ class parser {
...
+    function_guard guard = this->enter_function(attributes);
+    this->parse_and_visit_function_parameters_no_scope(v, name);
+    // function f() // Invalid.
+    if (this->peek().type != token_type::left_curly) {
+      if (!name.has_value()) {
+        QLJS_PARSER_UNIMPLEMENTED();
+      }
+      this->error_reporter_->report(error_missing_function_body{
+          .function_name = *name,
+      });
```
strager, line 1310: I think it would be better to show the error after the parameter list's `)`, similar to E064 (*missing body for 'if' statement*):
```suggestion
      this->error_reporter_->report(error_missing_function_body{
          .expected_body = this->lexer_.end_of_previous_token(),
      });
```

src/quick-lint-js/parse.h
```diff
@@ -1290,12 +1287,32 @@ class parser {
       QLJS_PARSER_UNIMPLEMENTED();
```
strager, line 1287: Nit: The comment on line 1283 doesn't make sense anymore. (`return` is no different from `break`.)

See also my comment on test_parse.function_statement_without_parameter_list_or_body.

test/test-parse-function.cpp
```diff
@@ -546,10 +546,15 @@ TEST(test_parse, function_statement_without_parameter_list_or_body) {
...
-                    offsets_matcher(&code, strlen(u8"{ function "), u8"f"))));
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(
+            ERROR_TYPE_FIELD(
+                error_missing_function_parameter_list, function_name,
+                offsets_matcher(&code, strlen(u8"{ function "), u8"f")),
+            ERROR_TYPE_FIELD(
+                error_missing_function_body, function_name,
+                offsets_matcher(&code, strlen(u8"{ function "), u8"f"))));
```
strager, line 557: Minor: I think this test should remain as-is. One error seems better than two errors for this case IMO.

Idea: Make `parse_and_visit_function_parameters_no_scope` return a boolean. If the parameter list is present, it returns `true`; if the parameter list is missing, it returns `false`. In `parse_and_visit_function_parameters_and_body_no_scope`, stop parsing if `parse_and_visit_function_parameters_no_scope` returns `false`.

test/test-parse-function.cpp
```diff
@@ -610,13 +622,121 @@ TEST(test_parse, function_statement_without_parameter_list_or_body) {
...
+                error_missing_function_body, function_name,
+                offsets_matcher(&code, strlen(u8"function "), u8"f")),
             ERROR_TYPE_FIELD(
                 error_missing_operand_for_operator, where,
                 offsets_matcher(&code, strlen(u8"function f"), u8".")),
             ERROR_TYPE_FIELD(
                 error_missing_semicolon_after_statement, where,
                 offsets_matcher(&code, strlen(u8"function f.x()"), u8""))));
   }
+  {
```
strager, line 635: Nit: Put these test cases into their own `TEST`:
```suggestion
}

TEST(test_parse, function_statement_without_body) {
  {
```

test/test-parse-function.cpp
```diff
@@ -610,13 +622,121 @@ TEST(test_parse, function_statement_without_parameter_list_or_body) {
...
+                                      "visit_exit_function_scope"));  //
+    EXPECT_THAT(v.errors,
+                ElementsAre(ERROR_TYPE_FIELD(
+                    error_missing_function_body, function_name,
+                    offsets_matcher(&code, strlen(u8"export default function "),
+                                    u8"f"))));
+  }
+
+  {
+    padded_string code(u8"(function f())"_sv);
```
strager, line 715: Nit: This is a function expression, not a function statement. The `TEST`'s name implies only function statements are tested. Perhaps the `TEST` should be renamed, or this test case moved into its own `TEST`.

strager (approved):
Thanks for the fix! It looks good, but it could be better.

Let me know if you want to merge your changes, or feel free to make improvements.

wgrr:
@strager my initial version would have avoided most of the problems you found, here is its diff for the `parse.h` with .function_name fixed:
```diff
diff --git a/src/quick-lint-js/parse.h b/src/quick-lint-js/parse.h
index cba9919f..cdd4b48f 100644
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -1259,6 +1259,16 @@ class parser {
         QLJS_PARSER_UNIMPLEMENTED();
       }
       this->skip();
+      if (this->peek().type != token_type::left_curly) {
+        const char8 *expected_body = this->lexer_.end_of_previous_token();
+        this->error_reporter_->report(error_missing_function_body{
+            .expected_body = source_code_span(expected_body, expected_body)
+
+        });
+
+        // Don't parse a function body.
+        return;
+      }

       break;


```

Why i splitted this on multiple functions was to keep reporting errors without hard-to-reason control flows like goto's or fallthrough's (i do agree that this causes too much noisy to nearly no benefit so i will change that). IMO the above diff doesn't read top-down as nice as the current one but does the job.

I've done your suggested changes too, but it breaks one other test, which is `test_parse.class_method_without_parameter_list`, it happens because it expects `parse_and_visit_function_parameters_and_body` to enter the body, even without an arg list, if we return true on `left_curly` besides the missing arg list, that passes the test.

```diff
    // function f {}  // Invalid.
    case token_type::left_curly:
      if (!name.has_value()) {
        QLJS_PARSER_UNIMPLEMENTED();
      }
      this->error_reporter_->report(error_missing_function_parameter_list{
          .function_name = *name,
      });
-      break;
+      return true;

```

So, from here i see 3 possible path:
 - stick with the small diff that works.
 - fix failing test by deleting some visit expects and add a more errors to its expect list
 - change the meaning of the returned bool to "should parse body" instead of "arg list is present"

Which one you think we should do? or you might have more ideas?

BR.

strager:
> stick with the small diff that works.

Let's go with this solution. I know the flow control is ugly, but error-handling is done like this throughout the parser. =S

strager (approved):

Issue #395, opened 2021-07-22
Emacs LSP Mode: automatically determine project root


strager:
When I `M-x lsp` to enable the LSP Mode plugin, I'm greeted with a scary message:
![Screenshot from 2021-07-21 18-59-09](https://user-images.githubusercontent.com/48666/126580875-4990c10c-62e6-4341-a9d4-9196dca76775.png)

Our plugin should automatically determine the root. I suggest using the filesystem root (`/` on POSIX; `C:\` or `X:\` or whatever on Windows).

strager:
Workaround: `(setq lsp-auto-guess-root t)` (thanks, Keyeh)

Issue #398, opened 2021-07-22
Emacs Flymake: diagnostics not cleared sometimes


strager:
Repro:
1. C-x b RET (open \*scratch\*)
2. M-x js-mode RET
3. Go to the first line and insert `//`
4. Go to the second line and insert `//`

At this point, Flymake *should* report no issues, but it reports an unclosed regexp literal error.

5. On a new line, type `documents`
6. Hover the mouse cursor over the word 'documents' and observe:
![Screenshot from 2021-07-21 20-18-08](https://user-images.githubusercontent.com/48666/126585558-0363dad4-eb7b-4282-b320-842572f822e7.png)

wgrr:
@strager i think this bug is related with state track of quick-lint-js, reported a new issue (#400) for the crash that this input is doing to quick-lint-js

strager:
> @strager i think this bug is related with state track of quick-lint-js process, reported a new issue (#400) for the crash that this input is doing to quick-lint-js

Do you think the plugin gets confused when quick-lint-js crashes?

wgrr:
> Do you think the plugin gets confused when quick-lint-js crashes?

Yes, I'm still investigating that, apparently if quick-lint-js crashes flymake is never calling us back so we don't have a chance to disable ourselves or update the current errors, but im not 100% sure about this, what im tho is we are never taking the :panic branch, even if we crash.
I'll bet it has some relationship with signals.

Issue #133, opened 2021-01-14
27$: Sign pre-built executables


strager:
Digitally sign the Windows and macOS executables we ship on npm.

- [x] Linux
- [x] macOS
- [x] Windows
- [ ] editor plugins; source repo; etc.
- [x] redo signing keys

strager:
For macOS, I need an Apple developer account. This requires 2fac on my Apple ID, and I'm having trouble setting up 2fac. =\

strager:
For Windows, I don't think Microsoft has an option for them to notarize executables. Therefore, I'd need my own CA-signed certificate. Cheap certificates seem to be around 75$: https://cheapcodesign.com/codesigning/authenti-codesigning

strager:
One use case for code signing on all platforms: editor plugins can verify that `./node_modules/.bin/quick-lint-js` is really quick-lint-js and not some malware.

strager:
# macOS

[macOS code signing](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html) (`codesign`)

# Windows

[Windows code signing](https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool) (`signtool`)

# Linux

Linux has no de-facto executable signing.

Separate PGP signatures are common. We don't *need* the signature to be embedded into the executable.

For embedded signatures, there are several options, including:

* [linux-elf-binary-signer](https://github.com/NUAA-WatchDog/linux-elf-binary-signer)
* [bsign](https://launchpad.net/ubuntu/+source/bsign)
* [SignElf](https://sourceforge.net/projects/signelf/)
* [elf64_sign_range](https://docs.oracle.com/cd/E86824_01/html/E54768/elf32-sign_range-3elf.html) (possibly in a libelf fork?)

(I didn't audit these solutions.)

strager:
Goal: Allow Vim plugin to use `node_modules/.bin/quick-lint-js`.

Problem: Any npm package can create `node_modules/.bin/quick-lint-js`. Also, any directory a user `cd`s into might have `node_modules/.bin/quick-lint-js`. `node_modules/.bin/quick-lint-js` might not refer to a 'real' quick-lint-js.

## Linux

Solution: Sign quick-lint-js with GnuPG. Have the Vim plugin verify the signature. The Vim plugin contains the public key.
Problem: Need tools to verify GnuPG signatures. We could ask users to install GPG if they want this feature.

[...]

PR #409, opened 2021-08-01
wip: fix #396, #395 and #398

### LINKED ISSUE #396: Emacs: automatically require plugins?
I tried out the four Emacs plugins. The installation process was a bit more annoying than it should be IMO.

* Flycheck has a `flycheck-verify-setup` which lists various linters. quick-lint-js does not appear in this list until you add `(require 'flycheck-quicklintjs)` to your ~/.emacs.
* The other plugins also need `(require ...)` in ~/.emacs. Some plugins need other stuff too, so this isn't so bad. But the LSP Mode and Elgot plugins work out of the box with just `(require ...)`.

For system-installed quick-lint-js, is it possible to auto-require to make installation easier?

wgrr:
The intention of this draft is mostly get feedback on how configurations should look like for plugins.
I feel doing all at the same place would be better than spread them across plugins. I made up a quick-lint-js search function from node_modules, however I'm unsure if it's the desire, currently the user must choose one or implement their search function, or by default use emacs built-in 'exec-path search function, i think we might want by default try npm, then exec-path? Lint requests timing is a missing unified configuration too.

Also, currently to find node_modules folder, we search backwards from buffers `default-directory` until we found it or not, Emacs 27 has a yet unstable API for a builtin library for projects named `project`, after version 0.3.0 it has a function `project-root` we could use or try before instead, but it's only available from elpa or emacs >= 28, which still in development.

@strager What you feel? is this on the right track?

- [ ] #395
- [ ] #396
- [x] #398


plugin/emacs/flymake-quicklintjs.el
```diff
@@ -40,47 +40,89 @@ Return a list of Flymake diagnostic objects in source buffer SRC-BUF."
...
+                                               qljs-stderr-buf
+                                               src-buf report-fn)
+  "Similar to `flymake-quicklintjs--report' but tries to recover errors from
+quick-lint-js crashes and logs whatever is in QLJS-STDERR-BUF"
+  (flymake-log :warning
+               "quick-lint-js exited with a deadly signal.\n\
+Please consider filing a bug at\
+ https://github.com/quick-lint/quick-lint-js/issues/new\n\
+qjls-stderr-buf:\n\
+%s\n"
```
strager, line 53: :+1:

plugin/emacs/flymake-quicklintjs.el
```diff
@@ -40,47 +40,89 @@ Return a list of Flymake diagnostic objects in source buffer SRC-BUF."
...
+               (with-current-buffer qljs-stderr-buf
+                 (buffer-substring-no-properties (point-min) (point-max))))
+  (with-current-buffer qljs-stdout-buf
+    (save-excursion
+      ;; a crash without any previous issue, try to fix an empty buffer
+      (when (zerop (buffer-size))
+        (goto-char (point-min))
+        (insert-char ?\())
+      ;; the above left a '(' danlging or crashed and
```
strager, line 62: Typo:
```suggestion
      ;; the above left a '(' dangling or crashed and
```

plugin/emacs/flycheck-quicklintjs.el
```diff
@@ -62,25 +53,22 @@
...
+(flycheck-define-command-checker 'javascript-quicklintjs
   "quick-lint-js finds bugs in JavaScript programs.
 
 https://quick-lint-js.com"
-  :command ("quick-lint-js"
-            "--output-format=emacs-lisp"
-            (eval (let ((file (buffer-file-name)))
-                    (if file
-                      `("--path-for-config-search" ,file)
-                      ())))
```
strager, line 74: Did you intentionally rewrite this code in this commit? This change seems unrelated to unifying settings. (The change looks fine, though.)

strager (changes requested):
See *must fix* comment. Otherwise, this looks good.

> I made up a quick-lint-js search function from node_modules, however I'm unsure if it's the desire, currently the user must choose one or implement their search function, or by default use emacs built-in 'exec-path search function, i think we might want by default try npm, then exec-path?

We **must not** search `node_modules` by default until code signing (#133) and signature checking is implemented. Otherwise, the plugin introduces an arbitrary code execution security risk.

**Must fix**: I don't want to even make `node_modules` searching an option, because people will enable it without understanding the security risks. If we do want the feature, we should discourage its use with a scary message.

strager:
> @strager What you feel? is this on the right track?

Unifying the plugin settings is a good idea!

wgrr:
> I don't want to even make `node_modules` searching an option, because people will enable it without understanding the security risks. If we do want the feature, we should discourage its use with a scary message.

Oh, i made this becasue if im not mistaken, i recall you mentioned on a past code review a "what if quick-lint-js was installed with npm problem", i infered that meant "look for quick-lint-js in node_modules", if that's not it, there's any secure way to do it? or should we gave up on that?

strager:
> i recall you mentioned on a past code review a "what if quick-lint-js was installed with npm problem",

Yup.

> i infered that meant "look for quick-lint-js in node_modules",

I did want that feature. I started implementing it for Vim, then realized the security problem with it. So I changed my opinion for now. I do eventually want the feature though.

> if that's not it, there's any secure way to do it?

Not yet. We need a way to ask "is this a trusted executable?". I haven't built that infrastructure yet. I filed #411 for making this feature.

> or should we gave up on that?

Let's give up on node_modules for now. We can save your patch for later.

wgrr:
@strager Could you review these CMake files? i havent written one in ages. Also, I'm still unsure about their need, i've written them because i'm unsure if emacs generated files are forward and backward compatible, so i left that to the host building the package decide (that's what happens in elpa/melpa, for example). But i have a feeling we could generate and track those files beforehand.

wgrr:
Also, explaining how #395 and #396 are fixed

7646781 fixes #396 by relying on Emacs builtin `package` library which automatically loads *-autoloads.el files, and those files are generated by `quicklintjs-pkg.el`. (the suggestion in the issue doesn't fix the issue because it would just move the problem elsewhere, e.g: it would need the user to add a `(add-to-list ...)` instead of require). Furthermore, this change will make the local installed symetric with the remotes i will add later (melpa and dpkg). Another path i could've taken is making a `site-start.el` which is a magic file automatically loaded by emacs, but using the builtin `package` library looks like the standard.

316e270 fixes #395 by adding a wrapper command which marks the `default-directory` as the root folder, unfortunately i think we can't transparently fix this, my first attempt was to use lsp-mode provided callback `lsp-before-initialize-hook`, but this is called after the prompt. I oppose to set the workaround variable `lsp-auto-guess-root` because it overwrites a global user setting and may leave lsp-mode in a broken state if `projectile` or `project` are previously misconfigured.

cmake/FindQuickLintJsEmacs.cmake
```diff
@@ -0,0 +1,81 @@
+# Copyright (C) 2020  Matthew "strager" Glazar
+# See end of file for extended copyright information.
+
+if (QUICK_LINT_JS_EMACS)
+  set(QUICK_LINT_JS_EMACS "${QUICK_LINT_JS_EMACS}")
+else ()
+  find_program(QUICK_LINT_JS_EMACS "emacs")
+endif ()
```
strager, line 8: `find_program` already checks if the program was found. No need to check yourself:
```suggestion
find_program(QUICK_LINT_JS_EMACS "emacs")
```

cmake/FindQuickLintJsEmacs.cmake
```diff
@@ -0,0 +1,81 @@
...
+
+if (NOT QUICK_LINT_JS_EMACS)
+  message(WARNING "Emacs not found. Skipping... ")
```
strager, line 11: Nit: I think this message is noisy. I'd omit it.
```suggestion
```

cmake/FindQuickLintJsEmacs.cmake
```diff
@@ -0,0 +1,81 @@
...
+  return ()
+endif()
+
+if (NOT EMACS_VERSION GREATER_EQUAL 24.5)
+  message(WARNING "Emacs found (${QUICK_LINT_JS_EMACS}), but version ${EMACS_VERSION} is not supported. Skipping...")
+  return ()
+endif ()
+
+set(QUICK_LINT_JS_EMACS_FOUND TRUE)
+message(STATUS "Found Emacs: (${QUICK_LINT_JS_EMACS}) suitable version ${EMACS_VERSION} minimum required is 24.5")
```
strager, line 34: Nit: Match format of similar messages in CMake:
```suggestion
message(STATUS "Found Emacs: ${QUICK_LINT_JS_EMACS} (found suitable version \"${EMACS_VERSION}\", minimum required is \"24.5\"")
```

cmake/FindQuickLintJsEmacs.cmake
```diff
@@ -0,0 +1,81 @@
...
+
+macro(emacs_pkg_target NAME FILE)
```
strager, line 36: Nit: I'd put this macro in plugin/emacs/CMakeLists.txt instead of here. This macro is tightly coupled to quicklintjs-pkg.el, so I think the two belong next to each other.

cmake/FindQuickLintJsEmacs.cmake
```diff
@@ -0,0 +1,81 @@
...
+    ""
+    ${ARGN})
+
+  add_custom_command(
+    OUTPUT ${_OUTPUT}
+    COMMAND
+      ${QUICK_LINT_JS_EMACS}
+      -Q -batch
+      -l package --eval "(package-initialize)"
+      --eval "(add-to-list 'package-directory-list \"${CMAKE_CACHEFILE_DIR}\")"
```
strager, line 50: Why add CMAKE_CACHEFILE_DIR? What files are in the build dir that we need to expose to Emacs?

plugin/emacs/CMakeLists.txt
```diff
@@ -4,13 +4,54 @@
...
 include(GNUInstallDirs)
 
-install(
-  FILES flycheck-quicklintjs.el
-  FILES lsp-quicklintjs.el
-  FILES eglot-quicklintjs.el
-  FILES flymake-quicklintjs.el
-  DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/emacs/site-lisp"
-)
+find_package(QuickLintJsEmacs)
```
strager, line 7: Nit: Spelling:
```suggestion
find_package(QuickLintJSEmacs)
```

strager (commented):
> @strager Could you review these CMake files?

They look good to me.

strager:
> Also, I'm still unsure about their need, i've written them because i'm unsure if emacs generated files are forward and backward compatible, so i left that to the host building the package decide (that's what happens in elpa/melpa, for example). But i have a feeling we could generate and track those files beforehand.

I think a build-time step (as you wrote) makes sense.

strager:
> [316e270](https://github.com/quick-lint/quick-lint-js/commit/316e2704d06f40d0183e1ffafd30fd0f464d2c4f) fixes #395 by adding a wrapper command which marks the `default-directory` as the root folder, unfortunately i think we can't transparently fix this, my first attempt was to use lsp-mode provided callback `lsp-before-initialize-hook`, but this is called after the prompt.

Aww.

I don't know if this is a good workaround. Would an lsp-mode user know to use this separate function? (I.e. would they blindly copy-paste from our documentation?)

Do other lsp-mode plugins have the same problem? If not, how do they solve it?

wgrr:
> Do other lsp-mode plugins have the same problem? If not, how do they solve it?

I just tried their builtin plugins for lsp-eslint, lsp-jsts and lsp-go, all of them prompt for a root direcotry.

The relavant piece which decides to prompt or not is here:
https://github.com/emacs-lsp/lsp-mode/blob/5b2daf6/lsp-mode.el#L8101-L8106

The only path I see is to make `lsp-find-session-folder` return non-nil by inserting the folder on its state file beforehand, I think instead we should just suggest to set `lsp-auto-guess-root`

cmake/FindQuickLintJsEmacs.cmake
wgrr: that's the output directory for the package, otherwise Emacs would make it somewhere on its defaults, (probably `~/.emacs.d/elpa`), i used the CMAKE_CACHEFILE_DIR because i assume cmake calls `${QUICK_LINT_JS_EMACS}` with current directory being the same from it was invoked from, so the root of build directory would have been filled with emacs package folders instead of `${BUILD_DIR}/plugins/emacs`. I cloud've made this from emacs by appending `plugins/emacs` to cwd, or you have something else in mind?

plugin/emacs/CMakeLists.txt
```diff
@@ -4,13 +4,54 @@
...
+  install(
+    DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/flymake-quicklintjs-0.0.1
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/emacs/site-lisp/elpa)
+  install(
+    DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/eglot-quicklintjs-0.0.1
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/emacs/site-lisp/elpa)
+  install(
+    DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lsp-quicklintjs-0.0.1
+    DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/emacs/site-lisp/elpa)
+endif ()
```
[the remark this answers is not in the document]
wgrr: > Could we make just one package?

Yes, to use what emacs calls "multi-file" packages, it would require `tar` available while building the packages tho, i didn't use it here because i imagine that's unlikely to `tar` be available on a Windows machine, what you think? i'm fine with both ways.

cmake/FindQuickLintJsEmacs.cmake
strager: > that's the output directory for the package

Oh. It looked like an input directory to me.

`${CMAKE_CURRENT_BINARY_DIR}` might be better.

> I cloud've made this from emacs by appending `plugins/emacs` to cwd, or you have something else in mind?

I think specifying cwd explicitly is what we should do:
```cmake
  add_custom_command(
    OUTPUT ${_OUTPUT}
    COMMAND
      ${QUICK_LINT_JS_EMACS}
      -Q -batch
      -l package --eval "(package-initialize)"
      --eval "(add-to-list 'package-directory-list .)"
      ...
    WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
    ...
  )
```

plugin/emacs/CMakeLists.txt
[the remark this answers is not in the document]
strager: > it would require `tar` available while building the packages tho, i didn't use it here because i imagine that's unlikely to tar be available on a Windows machine, what you think?

CMake provides a tool to create `tar` archives: https://cmake.org/cmake/help/v3.21/manual/cmake.1.html#run-a-command-line-tool

```
$ cmake -E tar czf src.tar.gz src
$ file src.tar.gz
src.tar.gz: gzip compressed data, last modified: Fri Aug 13 18:15:35 2021, from Unix, original size 1392640
```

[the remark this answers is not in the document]
wgrr: @strager i made the change to use a tar archive as a single package, besides that it feel clunky, it's discouraged by package archives, do you think the last "problem" in this list applies here? i do think. https://github.com/melpa/melpa/blob/5d49574/CONTRIBUTING.org#fixing-typical-problems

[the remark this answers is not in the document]
strager: > it's discouraged by package archives

Having multiple .el files in a package is discouraged? O_o I'm unsure if that's what the documentation actually says.

> do you think the last "problem" in this list applies here? i do think. https://github.com/melpa/melpa/blob/5d49574/CONTRIBUTING.org#fixing-typical-problems

It says:

> MELPA **only** looks at the main .el file (or -pkg.el file, if provided, though we discourage that).

What does it mean to "look at" an .el file?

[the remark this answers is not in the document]
wgrr: > What does it mean to "look at" an .el file?

In short, an .el file can be package on its own, its format something along the lines:
```
;;; <name>.el ---
;;; <headers>

<code>

;;; <name>.el ends here
```

so when it says "look at" an .el file, is an .el file which is in this package format, when a package has multiple .el files, this format is not used, instead there should be a tar archive that contains all .el files along with a magic `<name>-pkg.el`, which must call `(define-package ...args)`.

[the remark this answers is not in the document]
strager: > [a tar archive is] discouraged by package archives, do you think the last "problem" in this list applies here? i do think. https://github.com/melpa/melpa/blob/5d49574/CONTRIBUTING.org#fixing-typical-problems

I think we should do one of two things:
* merge all of our .el files into one .el file
* create a tar, even though MELPA docs say "we discourage that"

I'm happy with the tar if it's already implemented and if it makes developing the plugin easier for us.

strager:
@wgrr Are you waiting for feedback from me? The PR is in a draft state still.

strager:
ping @wgrr

wgrr:
> ping @wgrr

sorry, I've missed your previous comment in notifications, no, I went to a vacation and now I've been procrastinating on finishing this, I'll pick it up this weekend, sorry about the lack of input here.

Issue #407, opened 2021-07-27
unexpected keywords in expressions cause fatal errors


dav000:
`// missing open comment`
`no-case-declarations */;`

`../src/parse.cpp:506: fatal: token not implemented in parse_primary_expression: kw_case on line 2 column 4`

strager:
We should end an expression when we see `case`.

dav000:
There is the same or similar issue with these keywords
`break
case
catch
const
continue
default
do
else
export
extends
finally
function
try
var
with`

dav000:
> We should end an expression when we see `case`.

Does end an expression mean simply returning an invalid expression and not reporting an unexpected token error? It would be easy to add all of the keywords to the existing switch case, but there would also be new unhandled fatal errors for a lot of these keywords.

PR #412, opened 2021-08-06
Handle unexpected keywords in expressions

dav000:
#407 

The new behavior is to end the expression for these keywords. It will not report an unexpected token error in an expression. 

Because the code continues parsing instead of crashing, there are known new issues with handling `catch`, `export`, `extends`, `finally` and `function` expressions. These issues occur when keywords are mentioned in unopened comment blocks.

For example:

```
*                  
* this function extends a string 
* /
```

### DIFF
--- a/test/test-parse-expression-statement.cpp
+++ b/test/test-parse-expression-statement.cpp
@@ -1159,6 +1159,37 @@ TEST(test_parse, trailing_comma_in_comma_expression_is_disallowed) {
   }
 }
 
+TEST(test_parse, handle_unexpected_keyword_in_expression) {
+  //
+  // keywords temporarily omitted because they will crash downstream
+  // ---
+  // catch    parse_expression_remainder
+  // do       parse_and_visit_do_while
+  // else     parse_and_visit_statement
+  // extends  parse_expression_remainder: kw_extends
+  // finally  parse_expression_remainder: kw_finally
+  // function parse_and_visit_function_parameters_and_body_no_scope
+  // with     internal check failed in narrow_cast
+  //
+  for (string8_view statement : {
+           u8"break"_sv,
+           u8"case"_sv,
+           u8"const"_sv,
+           u8"continue"_sv,
+           u8"default"_sv,
+           u8"export"_sv,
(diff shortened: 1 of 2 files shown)


strager:
How did you test this change?

> Because the code continues parsing instead of crashing, there are known new issues with handling `catch`, `export`, `extends`, `finally` and `function` expressions.

Can you include a test case demonstrating the behavior (and why it might not be desirable)?

strager:
Maybe add some test cases to `incomplete_unary_expression_with_following_statement_keyword`?

dav000:
Attached the test file I used to confirm quick-lint-js doesn't crash. there are a few keywords commented out in the test file because they result in crashes in downstream code specific to the keyword. because those aren't simple fixes, they will need to have independent issues opened.
[kw4.txt](https://github.com/quick-lint/quick-lint-js/files/6948593/kw4.txt)

test/test-parse-expression-statement.cpp
```diff
@@ -1159,6 +1159,37 @@ TEST(test_parse, trailing_comma_in_comma_expression_is_disallowed) {
...
+           u8"break"_sv,
+           u8"case"_sv,
+           u8"const"_sv,
+           u8"continue"_sv,
+           u8"default"_sv,
+           u8"export"_sv,
+           u8"try"_sv,
+           u8"var"_sv,
+       }) {
+    padded_string code(string8(u8"*\n"_sv) + string8(statement));
```
strager, line 1184: Nit:
```suggestion
  for (string8 statement : {
           u8"break"_sv,
           u8"case"_sv,
           u8"const"_sv,
           u8"continue"_sv,
           u8"default"_sv,
           u8"export"_sv,
           u8"try"_sv,
           u8"var"_sv,
       }) {
    padded_string code(u8"*\n"_sv + statement);
```

strager (approved):

test/test-parse-expression-statement.cpp
dav000: the suggested code change doesn't compile

Issue #380, opened 2021-07-04
Don't report error missing initializer in const declaration in for...of and for...in statement


singalhimanshu:
The following code currently reports `error: missing initializer in const declaration [E205]`, it should not report any errors

for-of:
```js
const array1 = ['a', 'b', 'c'];

for (const element of array1) {
  console.log(element);
}
```

for-in:
```js
const array1 = ['a', 'b', 'c'];

for (const element in array1) {
  console.log(element);
}
```

singalhimanshu:
This bug was introduced in commit c7247245a14452d314ac26f0197c89301291333b.

strager:
The following are cases I thought of related and semi-related to this task. The cases marked with `!!` are where quick-lint-js is doing the wrong thing (false positive or false negative):
```js
let xs = [];
function cond() {}
function update() {}

for (const x = 3; cond(); update()) {}        //
for (const x = 3, y = 4; cond(); update()) {} //
for (const x = 3 of xs) {}                    // !! should error, but doesn't
for (const x of xs) {}                        // !! shouldn't error, but does
for (const x, y of xs) {}                     // should error
for (let x, y of xs) {}                       // !! should error, but doesn't
for (const x = 3 in xs) {}                    // !! should error, but doesn't
for (const x in xs) {}                        // !! shouldn't error, but does
for (const x, y in xs) {}                     // should error
for (let x, y in xs) {}                       // !! should error, but doesn't
for (const x; cond(); update()) {}            // should error
for (const x, y = 3; cond(); update()) {}     // should error
for (const x = 3, y; cond(); update()) {}     // should error
```

singalhimanshu:
I think these are the 3 different tasks that I'm able to segregate:
- [ ] `for of` and `for in` should not have assignment
```js
let xs = [];
for (const x = 3 in xs) {}
for (const x = 3 of xs) {}
```
- [ ] const in `for of` and `for in`  should not error
```js
let xs = [];
for (const x of xs) {}
for (const x in xs) {}
```
- [ ] Multiple variables are not allowed in `for of` and `for in`
```js
let xs = [];
for (let x, y of xs) {}
for (let x, y in xs) {}
```

strager:
> I think these are the 3 different tasks that I'm able to segregate:

Your split makes sense.

The second case matters the most right now, since it's a false positive.

singalhimanshu:
Fix for the second case is in PR: #421

PR #421, opened 2021-09-01
Don't report an error on uninitialized const in for of/in loop

singalhimanshu:
The following code:

    for (const x of []) {}
    for (const x in []) {}

Should not report `error_missing_initializer_in_const_declaration`
Fix: #380

### DIFF
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -2263,8 +2263,10 @@ class parser {
         // for (let i = 0; i < length; ++length) {}
         // for (let x of xs) {}
         this->lexer_.commit_transaction(std::move(transaction));
-        this->parse_and_visit_let_bindings(lhs, declaring_token,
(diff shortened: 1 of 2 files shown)


src/quick-lint-js/parse.h
```diff
@@ -3066,8 +3068,9 @@ class parser {
   }
 
   template <QLJS_PARSE_VISITOR Visitor>
-  void parse_and_visit_let_bindings(Visitor &v, token declaring_token,
-                                    bool allow_in_operator) {
+  void parse_and_visit_let_bindings(
+      Visitor &v, token declaring_token, bool allow_in_operator,
+      bool allow_const_without_initializer = false) {
```
strager, line 3073: Nit: I dislike default parameters, but I guess this is fine.

strager (approved):
Looks good to me. Is it ready to merge?

PR #422, opened 2021-09-05
Report an error on initialization in for in/of loop

singalhimanshu:
The following code:

    for (const x = 10 of []) {}
    for (let x = 10 in []) {}

Should report error_cannot_assign_to_loop_variable_in_for_of_or_in_loop.

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -117,6 +117,13 @@
             expected_last_component))                                          \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_cannot_assign_to_loop_variable_in_for_of_or_in_loop, "E173",       \
+      { source_code_span equal_token; },                                       \
+      ERROR(QLJS_TRANSLATABLE(                                                 \
+                "cannot assign to loop variable in for of/in loop"),           \
+            equal_token))                                                      \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_cannot_access_private_identifier_outside_class, "E208",            \
       { identifier private_identifier; },                                      \
       ERROR(                                                                   \
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -5,7 +5,6 @@
 #define QUICK_LINT_JS_PARSE_H
 
 #include <cstdlib>
-#include <iostream>
 #include <optional>
 #include <quick-lint-js/assert.h>
 #include <quick-lint-js/buffering-visitor.h>
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -2255,8 +2255,11 @@ class parser {
         default:
           this->lexer_.roll_back_transaction(std::move(transaction));
           this->skip();  // Re-parse 'let'.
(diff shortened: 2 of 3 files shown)


src/quick-lint-js/error.h
```diff
@@ -116,6 +116,13 @@
                 "C-style for loop is missing its third component"),            \
             expected_last_component))                                          \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_cannot_assign_to_loop_variable_in_for_of_or_in_loop, "E173",       \
+      { source_code_span equal_token; },                                       \
+      ERROR(QLJS_TRANSLATABLE(                                                 \
+                "cannot assign to loop variable in for of/in loop"),           \
```
strager, line 123: Minor: Maybe we should make a `statement_kind` for improve error messages. We can do this in a separate PR.

src/quick-lint-js/parse.h
```diff
@@ -3172,10 +3173,21 @@ class parser {
         // let x = 3;
         initialize_variable:
         case token_type::equal: {
+          token equal_token = this->peek();
           expression *ast = this->parse_expression_remainder(
               variable,
               precedence{.commas = false, .in_operator = allow_in_operator});
           this->visit_binding_element(ast, v, declaration_kind);
+          bool is_assignment_not_allowed =
+              this->in_loop_statement_ &&
```
strager, line 3181: Must fix: `this->in_loop_statement_` is the wrong thing to check. Your code produces a false positive for the following code:
```js
function of() {}
while (true) {
  const x = 3
  of();
}
```
```
hello.js:3:11: error: cannot assign to loop variable in for of/in loop [E173]
```

test/test-parse-loop.cpp
```diff
@@ -569,17 +569,52 @@ TEST(test_parse, for_in_loop) {
...
-  //   for (let x = init in xs) {}
-  //
-  // ('var' is allowed, but not 'let'.)
-
-  // TODO(strager): Report error for the following code:
-  //
-  //   for (var x = init of xs) {}
-  //
-  // (for-in is allowed, but not for-of.)
+  {
```
strager, line 572: Nit: I'd make a separate `TEST` for these invalid for-in loops.

test/test-parse-loop.cpp
```diff
@@ -569,17 +569,52 @@ TEST(test_parse, for_in_loop) {
...
+                    equal_token,
+                    offsets_matcher(&code, strlen(u8"for (let x "), u8"="))));
+  }
+
+  {
+    padded_string code(u8"for (var x = 10 in []) {}"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.visits, ElementsAre("visit_variable_declaration",  //
```
strager, line 613: Nit:
```suggestion
    EXPECT_THAT(v.visits, ElementsAre("visit_variable_declaration",  // x
```

strager (changes requested):
Looks good except the false positive.

strager:
This PR inspired #423.

strager (approved):

Issue #144, opened 2021-01-26
Nix: add Vim plugins


strager:
Inspiration: https://github.com/countoren/vim-nix

PR #440, opened 2021-09-23
Fix broken nix-build on darwin add vim related nix packages

countoren:
realtes to this issue [Nix: add Vim plugins #144](https://github.com/quick-lint/quick-lint-js/issues/144)

### DIFF
--- a/src/quick-lint-js/util/utf-8.cpp
+++ b/src/quick-lint-js/util/utf-8.cpp
@@ -10,7 +10,7 @@ namespace quick_lint_js {
 // See: https://www.unicode.org/versions/Unicode11.0.0/ch03.pdf
 char8* encode_utf_8(char32_t code_point, char8* out) {
   auto append = [&out](char32_t x) -> void {
-    QLJS_ASSERT(x <= 0x100);
+    QLJS_ASSERT(x <= 0xff);
     *out = static_cast<char8>(x);
     ++out;
   };


dist/nix/quick-lint-js.nix
```diff
@@ -15,15 +14,25 @@ mkDerivation {
...
+      filter = name: type: !(lib.hasSuffix ".nix" name);
+      src = ../../.;
+    };
+  };
   unpackPhase = null;
+  buildInputs = [ pkgs.python310 ];
 
-  nativeBuildInputs = with pkgs; [ cmake ninja ];
+  nativeBuildInputs = with pkgs; [
+    clang_11 #does not work with clang_12 or clang_7
```
strager, line 28: Shouldn't this dependency on `clang_11` only exist on Darwin (macOS), not on Linux?

dist/nix/quick-lint-js.nix
```diff
@@ -1,9 +1,8 @@
 # Copyright (C) 2020  Matthew "strager" Glazar
 # See end of file for extended copyright information.
 
-{ pkgs
+{ pkgs ? import <nixpkgs>{}
```
strager, line 4: What's this change for? If it's not related to Python or Clang or caching, put it in its own commit with an explanation.

dist/nix/quick-lint-js.nix
```diff
@@ -15,15 +14,25 @@ mkDerivation {
   pname = "quick-lint-js";
   version = builtins.head (builtins.split "\n" (builtins.readFile ../../version));
 
-  src = ../../.;
+  src = with pkgs; lib.cleanSourceWith {
+    filter = lib.cleanSourceFilter;
+    src = lib.cleanSourceWith {
+      filter = name: type: !(lib.hasSuffix ".nix" name);
```
strager, line 20: Why are we ignoring .nix files?

dist/nix/vim-with-quick-lint-js.nix
```diff
@@ -0,0 +1,14 @@
+{ pkgs ? import <nixpkgs> {} 
+, vimrcWithPlugins ? import ./vimrc-with-plugins.nix { inherit pkgs; } 
+, vim  ? pkgs.vim #Pick your vim
+, name ? "jsvim"
+}:
+pkgs.symlinkJoin {
+  inherit name;
+  paths = [ (pkgs.vim_configurable.override { python = pkgs.python3; }) ];
```
strager, line 8: Why do we need Python 3 support in Vim?

dist/nix/vim-with-quick-lint-js.nix
```diff
@@ -0,0 +1,14 @@
+{ pkgs ? import <nixpkgs> {} 
+, vimrcWithPlugins ? import ./vimrc-with-plugins.nix { inherit pkgs; } 
+, vim  ? pkgs.vim #Pick your vim
+, name ? "jsvim"
```
strager, line 4: Call it quick-lint-js-vim or vim-quick-lint-js or something.

dist/nix/vimrc-with-plugins.nix
```diff
@@ -0,0 +1,28 @@
+{ pkgs ? import <nixpkgs>{}
+, additionalPlugins? []
+, additionalCustPlugins? {}
+, additionalVimrc? ""#(pkgs.builtins.readFile ./vimrc)
```
strager, line 4: Remove debug comment?
```suggestion
, additionalVimrc? ""
```

dist/nix/quick-lint-js.nix
strager, line 28: `clang_11` isn't needed at run-time, so it should be in `buildInputs`, not `nativeBuildInputs`.

dist/nix/vimrc-with-plugins.nix
```diff
@@ -0,0 +1,28 @@
...
+}:
+with pkgs;
+let quick-lint-js = import ./quick-lint-js.nix { inherit pkgs;} ;
+in
+vimUtils.vimrcFile {
+  customRC =
+    ''
+      let g:ale_javascript_quick_lint_js_executable = '${quick-lint-js}/bin/quick-lint-js'
+    ''+additionalVimrc;
```
strager, line 13: Instead of inserting a line into the vimrc, we could generate a `plugin/quick-lint-js-nix.vim` file in the `buildPhase` of the plugin derivation (line 15).

dist/nix/vimrc-with-plugins.nix
```diff
@@ -0,0 +1,28 @@
...
+   vam.knownPlugins = vimPlugins // additionalCustPlugins // {
+     quick-lint-js = vimUtils.buildVimPluginFrom2Nix {
+       name = "quick-lint-js";
+       src = "${quick-lint-js}/share/vim/vimfiles";
+     };
```
strager, line 18: Could we have this part express the dependency on ALE? Maybe something like this (based on `vim-addon-xdebug` in `<nixpkgs>/pkgs/misc/vim-plugins/overrides.nix`):

    dependencies = with self; [ ale ];

strager (commented):
Instead of creating a vimrc file, we should just export the `vimUtils.buildVimPluginFrom2Nix` part. Then users can list `"quick-lint-js"` in their own `vam.pluginDictionaries`. (I think? I don't know. I never used Nix to manage this stuff.) (We can still provide a derivation for Vim which uses the plugin.)

dist/nix/quick-lint-js.nix
countoren: nix will bring clang_11 in both case and will use it only during build

dist/nix/vim-with-quick-lint-js.nix
countoren: the idea of this file is to be an example file, you we dont need it.

dist/nix/vimrc-with-plugins.nix
countoren: it meant to be an example line if you want to add your own vimrc configuration. you could use this function either passing your vimrc this way or change this default value and copy paste it.

dist/nix/quick-lint-js.nix
countoren: in order to work clang_11 had to be in ```nativeBuildInputs``` for me

strager: That seems wasteful on Linux where GCC is the default. If we don't have a good reason to use Clang on Linux, we should avoid Clang on Linux.

In my experience, GCC generates faster executables than Clang does. We should prefer GCC if it works and is the default anyway.

Issue #292, opened 2021-05-13
7$: Improve error when using => on object literal method


strager:
quick-lint-js reports a crappy message for the following code:
```js
let o = {
  myMethod() => {
  },
};
```
quick-lint-js should instead say something like "methods should not have =>".

This might also apply to class methods.

oromate:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

PR #444, opened 2021-09-24
Improve error when using => on object literal method

singalhimanshu:
The following code:
    let o = {
      method() => {
      },
    };

Should report error_methods_should_not_have_arrow_operator.
Fix: #292

### DIFF
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -1265,7 +1265,7 @@ class parser {
 
       if (this->peek().type == token_type::equal_greater) {
         this->error_reporter_->report(
-            error_methods_should_not_have_arrow_operator{
+            error_functions_or_methods_should_not_have_arrow_operator{
                 .arrow_operator = this->peek().span(),
             });
         this->skip();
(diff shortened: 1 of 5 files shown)


src/quick-lint-js/parse.h
```diff
@@ -1263,6 +1263,14 @@ class parser {
       }
       this->skip();
 
+      if (this->peek().type == token_type::equal_greater) {
```
strager, line 1266: This applies to `function f() => {}` too, right? If so, we should add a test case for that syntax. And we should tweak the message. (Perhaps have two different messages? One saying "functions" and one saying "methods".)

Same with `class C { f() => {} }`, except those are called methods, so the current wording works.

strager (approved):
Looks good!

src/quick-lint-js/parse.h
singalhimanshu: > This applies to `function f() => {}` too, right?

Right



> Perhaps have two different messages? One saying "functions" and one saying "methods

Do you mean two different errors? Also how do we know if we are in a function or method?

strager: > Do you mean two different errors?

Yes, or a single error type with a parameterized message. I'm leaining toward two different error types.

> Also how do we know if we are in a function or method?

Dunno.

strager (approved):

PR #284, opened 2021-05-11
Neovim: add native lsp support using lspconfig #105

tiagovla:
This is exactly how lspconfig expects this file to be including the description to generate their docs.

Issue #105, opened 2020-12-25
12$: Neovim: native LSP plugin


strager:

tiagovla:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

tiagovla:
Neovim has its own official LSP configuration [plugin](https://github.com/neovim/nvim-lspconfig/), so every server does not need an individual plugin.
I submitted a configuration for quick-lint-js [there](https://github.com/neovim/nvim-lspconfig/pull/887). Until it get accepted or if it doesn't, it can be used with the following configuration:
```lua
-- init.lua
local lspconfig = require("lspconfig")
local configs = require("lspconfig/configs")

if not lspconfig.quick_lint_js then
    configs.quick_lint_js = {
        default_config = {
            cmd = {"quick-lint-js", "--lsp-server"},
            filetypes = {"javascript"},
            root_dir = function(fname)
                return lspconfig.util.find_node_modules_ancestor(fname) or
                           lspconfig.util.find_git_ancestor(fname) or
                           vim.loop.os_homedir()
            end,
            settings = {},
        },
    }
end
lspconfig.quick_lint_js.setup {}
```
If it does get accepted, it can be configured like this:
```lua
require("lspconfig").quick_lint_js.setup {}
```

####  Additional configuration to reproduce the problem I was having with autocomplete, which is a third party plugin:

[Packer](https://github.com/wbthomason/packer.nvim) configuration:
```lua
vim.lsp.set_log_level("debug")
require("packer").startup(function()
    use "wbthomason/packer.nvim"
    use "neovim/nvim-lspconfig" -- install lspconfig
    use "hrsh7th/nvim-compe" -- install autocomplete plugin
end)
-- :PackerInstall :PackerCompile
```

[Compe](https://github.com/hrsh7th/nvim-compe/) configuration:
```lua
require("compe").setup {
    enabled = true,
    autocomplete = true,
    debug = false,
    min_length = 1,
    preselect = "enable",
    throttle_time = 80,
    source_timeout = 200,
    incomplete_delay = 400,
    max_abbr_width = 100,
    max_kind_width = 100,
    max_menu_width = 100,
    documentation = true,

    source = {
        nvim_lsp = true,
    },
}
```

Logs:
```sh
[...]

PR #458, opened 2021-10-01
Neovim: add native lsp support using lspconfig #105

strager:
Based on https://github.com/quick-lint/quick-lint-js/pull/284


strager:
Comment from @tiagovla:

> And #458 looks great. We might also need to test the auto completion feature with nvim-cmp.

I did some testing with nvim-cmp with an unreleased version of quick-lint-js. The quick-lint-js LSP server doesn't seem to crash.

PR #488, opened 2021-11-04
Suggest fix for mispelled await inside async function

keyehzy:
Fix for issue #279.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -818,8 +818,6 @@ expression* parser::parse_await_expression(token await_token, precedence prec) {
       });
     }
 
-    bool is_followed_by_async = this->peek().type == token_type::kw_async;
-
     expression* child = this->parse_expression(prec);
 
     if (child->kind() == expression_kind::_invalid) {
@@ -827,7 +825,7 @@ expression* parser::parse_await_expression(token await_token, precedence prec) {
           .where = operator_span,
       });
     } else if (this->in_async_function_ && this->is_arrow_kind(child) &&
-               !is_followed_by_async) {
+               child->attributes() != function_attributes::async) {
       this->error_reporter_->report(error_await_followed_by_arrow_function{
           .await_operator = operator_span,
       });
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -818,10 +818,18 @@ expression* parser::parse_await_expression(token await_token, precedence prec) {
       });
     }
 
+    bool is_followed_by_async = this->peek().type == token_type::kw_async;
+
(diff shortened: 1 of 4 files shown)


strager:
It looks like there are test failures. (See the CI job failures.) Do you need help fixing the tests?

keyehzy:
I was some commits behind and error code 176 was already taken, oops.

src/parse.cpp
```diff
@@ -818,11 +818,19 @@ expression* parser::parse_await_expression(token await_token, precedence prec) {
...
+    bool is_followed_by_async = this->peek().type == token_type::kw_async;
+
     expression* child = this->parse_expression(prec);
+
     if (child->kind() == expression_kind::_invalid) {
       this->error_reporter_->report(error_missing_operand_for_operator{
           .where = operator_span,
       });
+    } else if (this->in_async_function_ && this->is_arrow_kind(child) &&
+               !is_followed_by_async) {
```
strager, line 830: Nit: Instead of checking for the `async` token, you can ask if `child` is `async`:
```suggestion
               child->attributes() == function_attributes::async) {
```

src/quick-lint-js/error.h
```diff
@@ -88,6 +88,13 @@
       ERROR(QLJS_TRANSLATABLE("'await' is only allowed in async functions"),   \
             await_operator))                                                   \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_await_followed_by_arrow_function, "E177",                          \
```
strager, line 92: Minor: Now E177 is taken. xD Try E178. (I can fix this when I merge if you want.)

src/quick-lint-js/error.h
```diff
@@ -88,6 +88,13 @@
...
+      { source_code_span await_operator; },                                    \
+      ERROR(QLJS_TRANSLATABLE("'await' cannot be followed by an arrow "        \
+                              "function; use 'async' instead"),                \
```
strager, line 95: 👍

strager (approved):
Looks good! Let me know when I should merge.

Because we diagnose the `await` and suggest `async` instead, we should parse the arrow function's body as if `async` was written instead. For example, this should report only one error, not two:

```js
let f = await () => { // should error on 'await'
  await g(); // shouldn't error because of 'await'; we're supposed to be in an async function
};
```

But we can leave that for a separate PR.

keyehzy:
You can merge if you wish. I can make a followup PR, one step at a time.

strager (approved):

PR #498, opened 2021-11-08
Parse child of mispelled await expression as if it was inside async

### LINKED ISSUE #278: 7$: Improve error message for 'await () => {}'

keyehzy:
Right now quick-lint-js for the following code
```javascript
let f = await () => { 
  await g(); 
};
```
correctly detects the intention of user to use `async` instead of `await` but throw another error for `await` expression inside the arrow function. This PR fix this by parsing the subsequent expression as a `async` expression if we detect such intention. 

This PR should also fix #278.

### DIFF
--- a/src/parse.cpp
+++ b/src/parse.cpp
@@ -701,6 +701,27 @@ expression* parser::parse_async_expression_only(token async_token,
 }
 
 expression* parser::parse_await_expression(token await_token, precedence prec) {
+  bool is_followed_by_arrow = [&]() -> bool {
+    switch (this->peek().type) {
+    case token_type::left_paren:
+    case token_type::identifier: {
+      buffering_error_reporter temp_error_reporter(&this->temporary_memory_);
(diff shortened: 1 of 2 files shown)


keyehzy:
Not sure why dnf doesn't like, welp.

PR #504, opened 2021-11-11
💄 Make `:target` elements stand out

vegerot:
On a tall monitor (where you can fit the content of the whole screen
without scrolling), it can be difficult seeing where a particular in-page
link takes you.  This commit makes these elements stand out more

Video demonstration:

https://user-images.githubusercontent.com/5297725/141218315-d84b9b76-7f39-4a7a-a943-c5f7804d0f43.mp4

Notes for reviewer:
- There is a small layout shift with the transition.  However the shift is contained within the single section you select; so impact to CLS should be small if any (note: unmeasured)
- using the `h4` selector may be too broad.  A quick grep shows right now there is no leakage, but it may be prudent to add a class to these elements to prevent this behavior showing up in unexpected places in the future

```bash
rg "h4 id"
public/install/neovim/common-manual.ejs.html
18:<h4 id="manual-neovim-package">Install as a Neovim package</h4>
49:<h4 id="manual-neovim-dein-vim">Install with dein.vim</h4>
69:<h4 id="manual-neovim-packer-nvim">Install with packer.nvim</h4>
94:<h4 id="manual-neovim-vim-plug">Install with Vim-Plug</h4>

public/install/vim/common-manual.ejs.html
17:<h4 id="manual-vim-package">Install as a Vim package</h4>
49:<h4 id="manual-vim-pathogen">Install with Pathogen</h4>
72:<h4 id="manual-vim-vim-plug">Install with Vim-Plug</h4>
90:<h4 id="manual-vim-vundle">Install with Vundle</h4>
```

### DIFF
--- a/dist/arch/lint.sh
+++ b/dist/arch/lint.sh
@@ -26,6 +26,15 @@ trap 'rm -f "${errors}"' EXIT
 # check fails on libm. Even with -Wl,--as-needed, the linker keeps the NEEDED
 # entry, so I don't know how to work around the libm dependency.
 namcap --exclude=emptydir,symlink,unusedsodepends PKGBUILD-dev PKGBUILD-git PKGBUILD-release ./quick-lint-js-*.pkg.tar.zst |& tee "${errors}"
+# HACK(strager): --exclude doesn't work properly in namcap, probably as of
+# commit
+# <https://gitlab.archlinux.org/pacman/namcap/-/commit/6b19729433dd5c3d2d6be0d6f60732fefb006749>
+# (unconfirmed). Manually filter out errors that we wish to exclude.
+sed -i \
+    -e '\,Directory (usr/src/debug/quick-lint-js-dev/quick-lint-js/build) is empty,d' \
+    -e '\,Symlink (usr/lib/debug/\.build-id/5f/.*) points to non-existing \.\./\.\./\.\./\.\./bin/quick-lint-js,d' \
+    "${errors}"
+
 if [ -s "${errors}" ]; then
   printf 'error: namcap reported an error\n' >&2
   exit 1


website/public/main.css
```diff
@@ -166,6 +166,18 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
   content: "➤ ";
 }
 
+/* make "#" selected sections stand out */
+h4:target::before {
+  content: "👉"; /*TODO(😎): localize for different layout directions (e.g. rtl)*/
+}
```
strager, line 172: I think we should add a background to the `h4` instead. Maybe `rgba(0,0,0,0.1)` for light mode and `rgba(255,255,255,0.1)` for dark mode.

website/public/main.css
```diff
@@ -166,6 +166,18 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
...
+h4:target {
+  transition-property: box-shadow, transform;
+  transition-duration: 333ms;
+  transition-timing-function: ease;
+  box-shadow: 0 0.8rem 1.0rem 0.1rem rgba(0, 0, 0, 0.2);
+  transform: translateY(-0.4rem);
```
strager, line 178: I'm very confused about the styling here. It feels very out of place.

* Why a box shadow? It looks uncanny having a shadow surrounding text and extending to the right (in the middle of nowhere).
* Why does the text move location? (I could understand the pointer moving (sliding into view).)

website/public/main.css
```diff
@@ -166,6 +166,18 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
...
 }
 
+/* make "#" selected sections stand out */
+h4:target::before {
+  content: "👉"; /*TODO(😎): localize for different layout directions (e.g. rtl)*/
+}
+h4:target {
+  transition-property: box-shadow, transform;
+  transition-duration: 333ms;
+  transition-timing-function: ease;
```
strager, line 176: Minor: If we have an animation, we should respect [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion).

strager (changes requested):
I like the idea, but not the implementation.

Issue #475, opened 2021-10-26
12$: CI: add static analysis jobs


strager:
e.g. [MSVC](https://devblogs.microsoft.com/cppblog/microsoft-cpp-code-analysis-with-github-actions/) or clang-tidy

maljaroudi:
This can be done using CMake CMAKE_CXX_CLANG_TIDY option.
Here is how I implemented it: 87850c009ac5c8021e399ca1909067a42fa6c006

Of course, this downloads clang-tidy manually. The docker file can be adjusted to install clang-tidy-9 to make things easier.

The result would be the following (it may be possible to make it fail on warnings if preferred):
https://github.com/maljaroudi/quick-lint-js/runs/4147376794?check_suite_focus=true

strager:
@maljaroudi Does that upload a report so the analysis results can be viewed?

maljaroudi:
@strager Currently, no. The output is only printed on the build step. But it's possible to export it into a YAML file.
I believe it's possible to export the clang-tidy output via --export-fixes=<filename>.yml flag
so it would be something like:
`-DCMAKE_CXX_CLANG_TIDY="clang-tidy-9;-checks=clang-analyzer-*;--export-fixes=clang-tidy-output.yml"`
I can try cleaning up the workflow and test grabbing that file, but I am unsure where you wish to upload the report.

PR #510, opened 2021-11-12
#475 Added export flag for CMake and cleaned up its output

maljaroudi:
Note: The exported  CMake file seems to ignore clang CWE-119 for some reason

This should print out the output of the clang-tidy's clang-analysis-* from both: stdout and the file it exports


maljaroudi:

.github/workflows/check-static.yml
```diff
@@ -0,0 +1,49 @@
...
+jobs:
+  analysis:
+    name: clang-tidy
+    runs-on: ubuntu-latest
+    container: ghcr.io/quick-lint/quick-lint-js-github-builder:v1
+    env:
+      C: clang-9
+      CFLAGS: "-stdlib=libc++"
+      CXX: clang++-9
+      CXXFLAGS: "-std=gnu++17 -stdlib=libstdc++ -O2"
```
strager, line 19: Is `-O2` necessary to make clang-tidy do its thing? If not, I recommend `-O0` (default) to speed up the build.

.github/workflows/check-static.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: clang-tidy install
+        run: |
+          apt-get update
+          apt-get install -y clang-tidy-9
+      - name: CMake Build with clang-tidy enabled
+        run: cmake -DCMAKE_C_COMPILER=clang-9 -DCMAKE_CXX_COMPILER=clang++-9 -DCMAKE_C_FLAGS="${CFLAGS}" -DCMAKE_CXX_FLAGS="${CXXFLAGS}" -DBUILD_TESTING=OFF -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_CLANG_TIDY="clang-tidy-9;-checks=clang-analyzer-*;--export-fixes=clang-tidy-output.yml" -S . -B build
```
strager, line 28: Minor: We already specified `CC` in the environment, so why do we also need to specify the `CMAKE_C_COMPILER` CMake variable? Let's choose one or the other. (I prefer `CMAKE_C_COMPILER`.)

strager, line 28: Nit: `-DCMAKE_BUILD_TYPE=None` might build slightly faster than `-DCMAKE_BUILD_TYPE=Debug`, because it skips debug symbol generation.

strager, line 28: Does `--export-fixes=clang-tidy-output.yml` cause multiple files to be created? If so, should we report all of them?

.github/workflows/check-static.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: build with clang-tidy static analysis
+        run: cmake --build build -j9 1>/dev/null
```
strager, line 30: Our build servers only have two CPUs IIRC. `-j9` is overkill. Let's use the default setting.
```suggestion
        run: cmake --build build 1>/dev/null
```

strager, line 30: Why are we hiding stdout (`1>/dev/null`)?

.github/workflows/check-static.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: clang-tidy output
+        run: cat ./build/src/clang-tidy-output.yml
```
strager, line 32: In order for someone to see the report, they need to look for it, right? I doubt anyone is going to look for the report by hand. Do you plan on making this job post a comment on the PR or send an email? Or fail the build if there are any issues found? If not, then unless we develop a habit of looking at the reports, this job is a waste of CPU time.

strager (commented):
Looks good so far, but I don't know if this job will actually be helpful. Is this what other projects do for static analysis?

.github/workflows/check-static.yml
maljaroudi: I believe the export flag exports to a single yaml file which is specified.

.github/workflows/check-static.yml
maljaroudi: mostly to make the output cleaner to only show clang-tidy output which has line numbers as opposed to the byte offset found in the yaml file.

.github/workflows/check-static.yml
maljaroudi: per suggestion. I have added a fail condition on the latest commit. That should help making the action more noticeable.

maljaroudi:
Here is my attempt to make the workflow more relevant
218d590

[sample action after making the changes ](https://github.com/maljaroudi/quick-lint-js/runs/4259772317?check_suite_focus=true)

strager:
> Here is my attempt to make the workflow more relevant [218d590](https://github.com/quick-lint/quick-lint-js/commit/218d59090f023db20dbecfec8e2fa8003b6e866f)
> 
> [sample action after making the changes ](https://github.com/maljaroudi/quick-lint-js/runs/4259772317?check_suite_focus=true)

Thanks for showing the example!

Here's an excerpt:
```yaml
  - DiagnosticName:  clang-analyzer-deadcode.DeadStores
    DiagnosticMessage:
      Message:         'Value stored to ''input'' is never read'
      FilePath:        '/__w/quick-lint-js/quick-lint-js/src/lex.cpp'
      FileOffset:      6102
      Replacements:    []
```
This is missing a line number. How am I (as a contributor) supposed to use this information? There are many variables called `input` in that file.

maljaroudi:
> > Here is my attempt to make the workflow more relevant [218d590](https://github.com/quick-lint/quick-lint-js/commit/218d59090f023db20dbecfec8e2fa8003b6e866f)
> > [sample action after making the changes ](https://github.com/maljaroudi/quick-lint-js/runs/4259772317?check_suite_focus=true)
> 
> Thanks for showing the example!
> 
> Here's an excerpt:
> 
> ```yaml
>   - DiagnosticName:  clang-analyzer-deadcode.DeadStores
>     DiagnosticMessage:
>       Message:         'Value stored to ''input'' is never read'
>       FilePath:        '/__w/quick-lint-js/quick-lint-js/src/lex.cpp'
>       FileOffset:      6102
>       Replacements:    []
> ```
> 
> This is missing a line number. How am I (as a contributor) supposed to use this information? There are many variables called `input` in that file.

Unfortunately, there are no options for showing line numbers instead of byte offset in the`clang-tidy`output. 
For Vim users, `:goto xxxx` can help in navigating to the byte offset. 

Another option is using the output of "build with clang-tidy static analysis" which shows the line numbers by using `1>/dev/null` to remove the stdout and keep stderr. That way, clang-tidy output is displayed with the line number as opposed to the byte offset. 

I can try exploring other options such as using cppcheck as opposed to clang-tidy if preferred.

strager:
> Another option is using the output of "build with clang-tidy static analysis" step, shown in the example, which shows the line numbers by using `1>/dev/null` to remove the stdout and keep stderr. That way, clang-tidy output is displayed with the line number as opposed to the byte offset.

Would this show errors similar to Clang's normal compile errors? e.g. https://github.com/quick-lint/quick-lint-js/runs/4260383272?check_suite_focus=true

```
[snip]
[ 20%] Building CXX object src/CMakeFiles/quick-lint-js-lib.dir/buffering-error-reporter.cpp.o
[ 21%] Building CXX object src/CMakeFiles/quick-lint-js-lib.dir/byte-buffer.cpp.o
[ 21%] Building CXX object src/CMakeFiles/quick-lint-js-lib.dir/c-api-error-reporter.cpp.o
[ 22%] Building CXX object src/CMakeFiles/quick-lint-js-lib.dir/c-api.cpp.o
In file included from /__w/quick-lint-js/quick-lint-js/src/c-api.cpp:17:
In file included from /__w/quick-lint-js/quick-lint-js/src/./quick-lint-js/parse.h:10:
/__w/quick-lint-js/quick-lint-js/src/./quick-lint-js/buffering-visitor.h:171:51: error: unused parameter 'assignment_operator' [-Werror,-Wunused-parameter]
                                 source_code_span assignment_operator) {
                                                  ^
1 error generated.
make[2]: *** [src/CMakeFiles/quick-lint-js-lib.dir/c-api.cpp.o] Error 1
src/CMakeFiles/quick-lint-js-lib.dir/build.make:204: recipe for target 'src/CMakeFiles/quick-lint-js-lib.dir/c-api.cpp.o' failed
make[1]: *** [src/CMakeFiles/quick-lint-js-lib.dir/all] Error 2
make: *** [all] Error 2
CMakeFiles/Makefile2:1923: recipe for target 'src/CMakeFiles/quick-lint-js-lib.dir/all' failed
Makefile:145: recipe for target 'all' failed
Error: Process completed with exit code 2.
```

> I can try exploring other options such as using cppcheck as opposed to clang-tidy if preferred.

We can use clang-tidy, as long as we have a way to present issues clearly.

---

If simple solutions don't work, could we use something like this? https://github.com/ZedThree/clang-tidy-review

strager:
@maljaroudi It looks like there are some issues with your latest commit. A snippet from https://github.com/quick-lint/quick-lint-js/runs/4330962788?check_suite_focus=true:

```

/github/workspace/benchmark/benchmark-lsp/main.cpp:4:10: error: 'boost/json/serialize.hpp' file not found [clang-diagnostic-error]
#include <boost/json/serialize.hpp>
         ^
/github/workspace/benchmark/benchmark-lsp/quick-lint-js/benchmark-config.h:7:10: error: 'cstdint' file not found [clang-diagnostic-error]
#include <cstdint>
         ^
/github/workspace/benchmark/benchmark-lsp/quick-lint-js/lsp-benchmarks.h:7:10: error: 'memory' file not found [clang-diagnostic-error]
#include <memory>
         ^
/github/workspace/benchmark/benchmark-lsp/quick-lint-js/lsp-logging.h:7:1: error: unknown type name 'namespace' [clang-diagnostic-error]
namespace quick_lint_js {
^
```

Issue #478, opened 2021-10-30
Flaky test: VS Code 'file on disk changes'


strager:
macOS x86_64
https://github.com/quick-lint/quick-lint-js/runs/4052952235?check_suite_focus=true

```
Running file on disk changes ...

AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ []
- [
-   {
-     code: 'E034',
-     endCharacter: 5,
-     endLine: 1,
-     message: 'redeclaration of variable: x',
-     severity: 0,
-     source: 'quick-lint-js',
-     startCharacter: 4,
-     startLine: 1
-   }
- ]
	at /Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/vscode-tests.js:170:14
	at pollAsync (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/vscode-tests.js:1252:9)
	at runNextTicks (internal/process/task_queues.js:59:5)
	at processTimers (internal/timers.js:472:9)
	at async file on disk changes (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/vscode-tests.js:168:5)
	at async runTestAsync (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/test-support.js:51:5)
	at async runTestsAsync (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/test-support.js:35:7)
	at async testMainAsync (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/test-support.js:22:5)
	at async Object.runAsync [as run] (/Users/runner/work/quick-lint-js/quick-lint-js/plugin/vscode/test/vscode-tests.js:1277:3)

FAIL    file on disk changes

Error: test failed

Exit code:   1
Done
```

strager:
It failed on Windows x86 too: https://github.com/quick-lint/quick-lint-js/runs/4088084137?check_suite_focus=true
```
Running file on disk changes ...

AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ []
- [
-   {
-     code: 'E034',
-     endCharacter: 5,
-     endLine: 1,
-     message: 'redeclaration of variable: x',
-     severity: 0,
-     source: 'quick-lint-js',
-     startCharacter: 4,
-     startLine: 1
-   }
- ]
[...]

PR #516, opened 2021-11-17
German Translation of error messages

herrhotzenplotz:
This is still a bit in progress, though all messages are already translated. I also regenerated the messages.pot template, as some of the error messages would otherwise still be in English; I hope that this is okay. As a result, someone needs to check the Swedish translation again.
If anyone has issues with some translations, please discuss as some things obviously do not translate very well. 
I am going over everything once again and shall give you the go for review/merge.


herrhotzenplotz:
No idea, why updating a translation update in this way would fail this one and only check. Looks like a bug to me.

strager:
> No idea, why updating a translation in this way would fail this one and only check. Looks like a bug to me.

It's a known flaky test: #478

You can safely ignore that failure.

po/messages.pot
```diff
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: strager.nds@gmail.com\n"
-"POT-Creation-Date: 2020-10-17 14:17+0100\n"
-"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"POT-Creation-Date: 2021-11-16 19:19+0100\n"
+"PO-Revision-Date: 2021-11-17 17:13+0100\n"
```
strager, line 10: Nit: I think `PO-Revision-Date` should not contain a real date, because this is a .pot file not a .po file. But I don't really know much about this stuff. :woman_shrugging:

po/de.po
```diff
@@ -0,0 +1,867 @@
...
+msgid "unexpected 'catch' without 'try'"
+msgstr "Unerwartetes 'catch' ohne zugehöriges 'try'"
+
+#: src/quick-lint-js/error.h
+msgid "missing body for {1:headlinese}"
+msgstr "Körper für {1:headlinese} fehlt"
+
+#: src/quick-lint-js/error.h
+msgid "a class statement is not allowed as the body of {1:singular}"
+msgstr "Ein Klassen-Statement darf nicht der Körper von {1:plural} sein"
```
strager, line 144: Is singular -> plural intentional?

po/de.po
```diff
@@ -0,0 +1,867 @@
...
+msgid "legacy octal literals may not contain underscores"
+msgstr "Veraltete Oktalliterale dürfen keine Unterstriche enthalten"
+
+#: src/quick-lint-js/error.h
+msgid "{0} with no bindings"
+msgstr "{0} ohne Binding"
+
+#: src/quick-lint-js/error.h
+msgid "a lexical declaration is not allowed as the body of {1:singular}"
+msgstr "Eine lexikalische Deklaration ist nicht als Körper von {0:plural} erlaubt"
```
strager, line 394: Must fix: You used the wrong placeholder. It should be `{1:singular}`, not `{0:plural}`.

po/de.po
```diff
@@ -0,0 +1,867 @@
...
+msgstr "Semikolon fehlt zwischen Initialisierung und Bedingung der for-Schleife"
+
+#: src/quick-lint-js/error.h
+msgid ""
+"incomplete export; expected 'export default ...' or 'export {{name}' or "
+"'export * from ...' or 'export class' or 'export function' or 'export let'"
+msgstr ""
+"Unvollständiger Export. Möglichkeiten sind:"
+" 'export default ...' oder"
+" 'export {name}' oder"
```
strager, line 591: Must fix: `{` needs to be escaped with another `{`:
```suggestion
" 'export {{name}' oder"
```

po/de.po
```diff
@@ -0,0 +1,867 @@
...
+
+#: src/quick-lint-js/error.h
+msgid ""
+"unexpected token in export; expected 'export default ...' or 'export "
+"{{name}' or 'export * from ...' or 'export class' or 'export function' or "
+"'export let'"
+msgstr ""
+"Unerwartetes Token in Export. Möglichkeiten:"
+" 'export default ...' oder"
+" 'export {name}' oder"
```
strager, line 805: Must fix: escape `{` with `{`.

po/de.po
herrhotzenplotz: Nope, definitely not.

po/messages.pot
herrhotzenplotz: Reverted.

strager (approved):

strager:
Please write a nice commit message for me, then I'll merge it. Thanks!

Issue #501, opened 2021-11-10
spans for E094, E104, E106, E111, and E119 are silly


strager:
They should point to where the correction should be made, not a bunch of unrelated characters which aren't the problem.

strager:
Fixed by @ahmafi. Thanks!

PR #524, opened 2021-11-20
Fix spans for E094, E104, E106, E111, and E119

ahmafi:
Fix for issue #501

### DIFF
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -1431,12 +1431,13 @@ class parser {
       this->skip();
       break;
 
-    default:
+    default: {
+      const char8 *here = this->lexer_.end_of_previous_token();
       this->error_reporter_->report(error_missing_body_for_class{
-          .class_keyword_and_name_and_heritage = source_code_span(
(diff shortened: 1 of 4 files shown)


ahmafi:

ahmafi:
rechek

src/quick-lint-js/parse.h
```diff
@@ -2481,9 +2484,9 @@ class parser {
     bool parsed_body =
         this->parse_and_visit_statement_disallowing_declaration(v);
     if (!parsed_body) {
+      const char8 *here = this->lexer_.end_of_previous_token();
       this->error_reporter_->report(error_missing_body_for_for_statement{
```
strager, line 2488: Perhaps we should include a *NOTE* mentioning where the `for` loop starts. See E141 for example.

strager (approved):
Looks great!

Can you write a summarizing commit message, or improve each of your commit's messages? "fix span E104 " doesn't explain *why* the change is being made.

strager:
Merged as commits bdeadfef (nice) through 06bc76f6.

Issue #506, opened 2021-11-11
Link to error doc URLs


strager:
e.g. https://quick-lint-js.com/errors/E074/

We would redirect `/errors/E074/` to `/errors/#E074`. Link to the non-fragment version just in case we refactor the website.

- [ ] LSP server
- [ ] CLI
- [ ] VS Code extension

ahmafi:
For CLI links we need something like this?

![image](https://user-images.githubusercontent.com/12560109/142930406-8363d091-f4be-448c-b3dc-2a166a8a7132.png)

strager:
> For CLI links we need something like this?

Nah, more like this:

<img width="984" alt="Screen Shot 2021-11-22 at 12 47 41 PM" src="https://user-images.githubusercontent.com/48666/142933099-4d3a07b1-3170-437d-99f4-002cdc5ecf46.png">
<img width="980" alt="Screen Shot 2021-11-22 at 12 47 49 PM" src="https://user-images.githubusercontent.com/48666/142933102-fb0efe71-f9b0-465f-9730-2b8b9001ed2a.png">

```sh
#!/bin/sh
printf 'hello.js:4:8: error: missing operand for operator [\e]8;;https://quick-lint-js.com/errors/#E0026\e\\E0026\e]8;;\e\\]\n'
```
https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

PR #534, opened 2021-11-23
CLI,  VSCode and LSP errors hyperlink to online docs

ahmafi:
Fix for #506 

- [x]  CLI
- [x]  VS Code extension
- [x]  LSP server

### DIFF
--- a/docs/cli.adoc
+++ b/docs/cli.adoc
@@ -49,14 +49,15 @@ Incompatible with *--lsp-server*.
 --
 
 *--diagnostic-hyperlinks*=_when_::
-  Control whether the errors are using escape sequences to show errors as hyperlinks or not.
+  Control whether to hyperlink error codes or not.
+  This option is only used if *--output-format=gnu-like*.
   _when_ is one of the following:
 +
 --
-- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On windows
-it behaves like 'never' because windows default terminals don't support escape sequences.
-- *always*: always shows errors as hyperlinks.
-- *never*: never shows errors as hyperlinks.
+- *auto* (default): shows error codes as hyperlinks only if the error output is a terminal.
+On Windows it behaves like 'never'.
+- *always*: always shows error codes as hyperlinks.
+- *never*: never shows error codes as hyperlinks.
 
 Incompatible with *--lsp-server*.
 --
--- a/docs/cli.adoc
+++ b/docs/cli.adoc
@@ -48,6 +48,19 @@ This command has two modes:
 Incompatible with *--lsp-server*.
 --
 
+*--diagnostic-hyperlinks*=_when_::
+  Control whether the errors are using escape sequences to show errors as hyperlinks or not.
+  _when_ is one of the following:
++
+--
+- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On windows
+it behaves like 'never' because windows default terminals don't support escape sequences.
+- *always*: always shows errors as hyperlinks.
+- *never*: never shows errors as hyperlinks.
+
+Incompatible with *--lsp-server*.
+--
+
 *--vim-file-bufnr*=_number_::
   Set the _bufnr_ property for errors printed with the *--output-format=vim-qflist-json* option.
 +
--- a/plugin/vscode/quick-lint-js/vscode-error-reporter.h
+++ b/plugin/vscode/quick-lint-js/vscode-error-reporter.h
@@ -61,14 +61,13 @@ class vscode_error_formatter
         /*severity=*/severity,
     });
 
-    char error_path[14];  // "/errors/E0014" is 14 characters with \0
-    snprintf(error_path, sizeof(error_path), "/errors/%s", code.data());
-    QLJS_ASSERT(std::strlen(error_path) == 13);
-
     ::Napi::Value uri = this->vscode_->uri_class.New(
         {/*scheme*/ ::Napi::String::New(this->env_, "https"),
          /*authority*/ ::Napi::String::New(this->env_, "quick-lint-js.com"),
-         /*path*/ ::Napi::String::New(this->env_, error_path)});
+         /*path*/ ::Napi::String::New(this->env_, "/errors/"),
+         /*query*/ ::Napi::String::New(this->env_, ""),
+         /*fragment*/
+         ::Napi::String::New(this->env_, code.data(), code.size())});
 
     ::Napi::Object code_obj = ::Napi::Object::New(this->env_);
     code_obj.Set("target", uri);
--- a/plugin/vscode/quick-lint-js/vscode-error-reporter.h
+++ b/plugin/vscode/quick-lint-js/vscode-error-reporter.h
@@ -62,11 +62,11 @@ class vscode_error_formatter
     });
 
     ::Napi::Value uri = this->vscode_->uri_class.New(
-        {/*scheme*/ ::Napi::String::New(this->env_, "https"),
-         /*authority*/ ::Napi::String::New(this->env_, "quick-lint-js.com"),
-         /*path*/ ::Napi::String::New(this->env_, "/errors/"),
-         /*query*/ ::Napi::String::New(this->env_, ""),
-         /*fragment*/
+        {/*scheme=*/::Napi::String::New(this->env_, "https"),
+         /*authority=*/::Napi::String::New(this->env_, "quick-lint-js.com"),
+         /*path=*/::Napi::String::New(this->env_, "/errors/"),
+         /*query=*/::Napi::String::New(this->env_, ""),
+         /*fragment=*/
          ::Napi::String::New(this->env_, code.data(), code.size())});
 
     ::Napi::Object code_obj = ::Napi::Object::New(this->env_);
--- a/plugin/vscode/quick-lint-js/vscode-error-reporter.h
+++ b/plugin/vscode/quick-lint-js/vscode-error-reporter.h
@@ -60,7 +60,22 @@ class vscode_error_formatter
             this->env_, reinterpret_cast<const char*>(this->message_.data())),
         /*severity=*/severity,
     });
-    diag.Set("code", ::Napi::String::New(this->env_, code.data(), code.size()));
+
+    char error_path[14];
+    sprintf(error_path, "/errors/%s", code.data());
+
+    ::Napi::Object uri = ::Napi::Object::New(this->env_);
+    uri.Set("scheme", ::Napi::String::New(this->env_, "https"));
+    uri.Set("authority", ::Napi::String::New(this->env_, "quick-lint-js.com"));
+    uri.Set("path", ::Napi::String::New(this->env_, error_path));
+    uri.Set("fragment", ::Napi::String::New(this->env_, ""));
+    uri.Set("query", ::Napi::String::New(this->env_, ""));
+
+    ::Napi::Object code_obj = ::Napi::Object::New(this->env_);
+    code_obj.Set("target", uri);
+    code_obj.Set("value", code.data());
+
+    diag.Set("code", code_obj);
     diag.Set("source", ::Napi::String::New(this->env_, "quick-lint-js"));
     this->diagnostics_.Set(this->diagnostics_.Length(), diag);
   }
--- a/plugin/vscode/quick-lint-js/vscode-error-reporter.h
+++ b/plugin/vscode/quick-lint-js/vscode-error-reporter.h
@@ -61,15 +61,14 @@ class vscode_error_formatter
         /*severity=*/severity,
     });
 
-    char error_path[14];
-    sprintf(error_path, "/errors/%s", code.data());
+    char error_path[14];  // "/errors/E0014" is 14 characters with \0
+    snprintf(error_path, sizeof(error_path), "/errors/%s", code.data());
+    QLJS_ASSERT(std::strlen(error_path) == 13);
 
-    ::Napi::Object uri = ::Napi::Object::New(this->env_);
-    uri.Set("scheme", ::Napi::String::New(this->env_, "https"));
-    uri.Set("authority", ::Napi::String::New(this->env_, "quick-lint-js.com"));
-    uri.Set("path", ::Napi::String::New(this->env_, error_path));
-    uri.Set("fragment", ::Napi::String::New(this->env_, ""));
-    uri.Set("query", ::Napi::String::New(this->env_, ""));
+    ::Napi::Value uri = this->vscode_->uri_class.New(
+        {/*scheme*/ ::Napi::String::New(this->env_, "https"),
+         /*authority*/ ::Napi::String::New(this->env_, "quick-lint-js.com"),
+         /*path*/ ::Napi::String::New(this->env_, error_path)});
 
     ::Napi::Object code_obj = ::Napi::Object::New(this->env_);
     code_obj.Set("target", uri);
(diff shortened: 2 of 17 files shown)


test/test-cli.py
```diff
@@ -242,7 +242,7 @@ def test_config_error_for_multiple_js_files_is_printed_only_once(self) -> None:
                 encoding="utf-8",
             )
             self.assertEqual(result.returncode, 1)
-            self.assertEqual(result.stderr.count("E0164"), 1)
+            self.assertEqual(result.stderr.count("E0164"), 2)
```
strager, line 245: This is confusing. We should instead do something like strip escape sequences in the test.

Must fix: Actually, the CLI shouldn't use escape sequences if the output is a pipe. We should only print escape sequences if the output is a terminal. (You can check if the output is a terminal using `isatty` on POSIX platforms.)

test/test-lsp-error-reporter.cpp
```diff
@@ -62,6 +62,8 @@ TEST_F(test_lsp_error_reporter, big_int_literal_contains_decimal_point) {
             "BigInt literal contains decimal point");
   EXPECT_EQ(look_up(diagnostics, 0, "code"), "E0005");
   EXPECT_EQ(look_up(diagnostics, 0, "source"), "quick-lint-js");
+  EXPECT_EQ(look_up(diagnostics, 0, "codeDescription", "href"),
+            "https://quick-lint-js.com/errors/E0005");
```
strager, line 66: This link doesn't work yet. We should update the website first.

src/text-error-reporter.cpp
```diff
@@ -63,7 +63,9 @@ void text_error_formatter::write_message_part(
 void text_error_formatter::write_after_message(std::string_view code,
                                                diagnostic_severity,
                                                const source_code_span &) {
-  this->output_ << " [" << code << "]\n";
+  this->output_ << " ["
+                << "\x1B]8;;https://quick-lint-js.com/errors/" << code
+                << "\x1B\\" << code << "\x1B]8;;\x1B\\]\n";
```
strager, line 68: Must fix: We shouldn't do this when writing to a pipe or a file. Use `isatty` to check if we are writing to a terminal.

It might be impossible to call `isatty` here since `this->output_` is an `std::ostream`, which might not be backed by a file descriptor. So perhaps we should pass the result of `isatty` to `text_error_reporter`'s constructor.

We might want to provide the option in the constructor anyway if we provide the user the option of enabling or disabling these escape codes. For example, GCC (a C++ compiler) allows the user to choose whether URLs are included for diagnostics using the `-fdiagnostics-urls` option: https://gcc.gnu.org/onlinedocs/gcc-11.2.0/gcc/Diagnostic-Message-Formatting-Options.html

plugin/vscode/test/vscode-tests.js
```diff
@@ -52,9 +52,19 @@ for (let extension of [".js", ".mjs", ".cjs"]) {
...
           {
-            code: "E0034",
+            code: {
+              target: {
+                scheme: "https",
+                authority: "quick-lint-js.com",
+                path: "/errors/E0034",
+                query: "",
+                fragment: "",
+              },
```
strager, line 65: Let's make `normalizeDiagnostics` convert the URI object into a string. This will let us simplify the assertion:
```suggestion
              target: "https://quick-lint-js.com/errors/E0034",
```

plugin/vscode/quick-lint-js/vscode-error-reporter.h
```diff
@@ -60,7 +60,22 @@ class vscode_error_formatter
...
+
+    char error_path[14];
+    sprintf(error_path, "/errors/%s", code.data());
+
+    ::Napi::Object uri = ::Napi::Object::New(this->env_);
+    uri.Set("scheme", ::Napi::String::New(this->env_, "https"));
+    uri.Set("authority", ::Napi::String::New(this->env_, "quick-lint-js.com"));
+    uri.Set("path", ::Napi::String::New(this->env_, error_path));
+    uri.Set("fragment", ::Napi::String::New(this->env_, ""));
+    uri.Set("query", ::Napi::String::New(this->env_, ""));
```
strager, line 72: Must fix: The TypeScript definitions say that the `target` property must be an instance of `Uri`. A new object with the default prototype is not an instance of `Uri`. We should adhere to the documented specification:
```typescript
    export class Diagnostic {
        code?: string | number | {
            value: string | number;
            target: Uri;
        };

        // ...
      }
```

plugin/vscode/quick-lint-js/vscode-error-reporter.h
```diff
@@ -60,7 +60,22 @@ class vscode_error_formatter
             this->env_, reinterpret_cast<const char*>(this->message_.data())),
         /*severity=*/severity,
     });
-    diag.Set("code", ::Napi::String::New(this->env_, code.data(), code.size()));
+
+    char error_path[14];
+    sprintf(error_path, "/errors/%s", code.data());
```
strager, line 65: Must fix: Use `std::snprintf`, not `sprintf`. Also, `QLJS_ASSERT` that `sprintf`/`snprintf` didn't fail and that the result is not truncated.

Or, be lazy and use `std::string` concatenation.

strager (changes requested):

test/test-lsp-error-reporter.cpp
ahmafi: Should I change the links to the fragment version? Or we update the website later?

src/text-error-reporter.cpp
ahmafi: Should we do both?
1. User option for enabling or disabling escape codes
2. Check `stdout` with `isatty`

strager: > Should I do both?
>
> 1. User option for enabling or disabling escape codes
> 2. Check `stdout` with `isatty`

Yes. The default will be #2 ("auto"), and we allow the user to overwrite the decision with #1.

> Also should we write different tests for escaped characters?

In general, we should test what we want to work.

A "proper" test involves writing a terminal emulator. Not worth it. A "hacky" test involves string comparisons, in addition to manual testing, which is fine IMO.

test/test-lsp-error-reporter.cpp
strager: > Should I change the links to the fragment version?

Yeah.

plugin/vscode/test/vscode-tests.js
```diff
@@ -1192,7 +1230,14 @@ function normalizeDiagnostics(vscodeDiagnosticsOrURI) {
     vscodeDiagnostics = vscodeDiagnosticsOrURI;
   }
   return vscodeDiagnostics.map((diag) => ({
-    code: diag.code,
+    code: {
+      target: `${diag.code.target.scheme}://${diag.code.target.authority}${
+        diag.code.target.path
+      }${diag.code.target.query ? "?" + diag.code.target.query : ""}${
+        diag.code.target.fragment ? "#" + diag.code.target.fragment : ""
+      }`,
```
strager, line 1238: Does `diag.code.target.toString()` not work?

src/main.cpp
```diff
@@ -612,6 +614,11 @@ void print_help_message() {
   print_option("", "gnu-like (default if omitted)");
   print_option("", "vim-qflist-json");
   print_option("", "emacs-lisp");
+  print_option("--escape-errors=[WHEN]",
+               "Use escape sequences to show errors as hyperlinks.");
```
strager, line 618: Must fix: `--escape-errors` is a bad name. Let's call it `--diagnostic-hyperlinks` instead.

The fact that we use escape sequences is an implementation detail.

src/main.cpp
```diff
@@ -612,6 +614,11 @@ void print_help_message() {
...
+  print_option("", "WHEN can be 'auto' (default if omitted), 'always',");
+  print_option("", "'never'. auto uses escape sequences when the standard");
+  print_option("", "error is a terminal");
```
strager, line 621: Nit: Detailed usage documentation belongs in the man page (`docs/cli.adoc`).

src/text-error-reporter.cpp
```diff
@@ -23,16 +35,42 @@ void text_error_reporter::set_source(padded_string_view input,
...
   QLJS_ASSERT(this->locator_.has_value());
+
+  bool format_escape_errors = false;
+
+  switch (this->escape_errors_) {
+  case escape_errors::auto_:
+#if defined(_WIN32)
+    if (this->output_.rdbuf() == std::cerr.rdbuf() && _isatty(_fileno(stderr)))
+#else
+    if (this->output_.rdbuf() == std::cerr.rdbuf() && isatty(fileno(stderr)))
```
strager, line 46: Nit: No need for `fileno`:
```suggestion
    if (this->output_.rdbuf() == std::cerr.rdbuf() && isatty(STDIN_FILENO))
```

src/text-error-reporter.cpp
```diff
@@ -23,16 +35,42 @@ void text_error_reporter::set_source(padded_string_view input,
...
+    else
+      format_escape_errors = false;
+    break;
+  case escape_errors::always:
+    format_escape_errors = true;
+    break;
+  case escape_errors::never:
+    format_escape_errors = false;
+    break;
+  }
```
strager, line 58: Must fix: This is very expensive to do every time `report` is called. Do this work in the constructor.

test/test-lsp-server.cpp
```diff
@@ -271,8 +271,12 @@ TEST_F(test_linting_lsp_server, opening_document_lints) {
                         "end": {"line": 0, "character": 9}
                       },
                       "severity": 1,
-                        "message": "variable used before declaration: x"
+                      "message": "variable used before declaration: x",
+                      "code": "E0058",
+                      "codeDescription": {
+                        "href": "https://quick-lint-js.com/errors/#E0058"
                       }
```
strager, line 278: Nit: This test doesn't really need these. I'd undo the changes to this test.

strager (changes requested):
See 'must fix' comments. Other than that, this PR looks good to me.

ahmafi:
I have also tested CLI manually in windows and it works as expected.

src/text-error-reporter.cpp
```diff
@@ -10,9 +11,40 @@
...
+  case escape_errors::always:
+    this->format_escape_errors_ = true;
+    break;
+  case escape_errors::never:
+    this->format_escape_errors_ = false;
+    break;
+  }
+}
+
+bool text_error_reporter::use_escape_if_auto() {
```
strager, line 41: Nit: Maybe `output_supports_escapes` is a better name?

src/text-error-reporter.cpp
```diff
@@ -23,16 +55,22 @@ void text_error_reporter::set_source(padded_string_view input,
...
   QLJS_ASSERT(this->file_path_);
   QLJS_ASSERT(this->locator_.has_value());
-  text_error_formatter formatter(/*output=*/this->output_,
-                                 /*file_path=*/this->file_path_,
-                                 /*locator=*/*this->locator_);
+  text_error_formatter formatter(
+      /*output=*/this->output_,
+      /*file_path=*/this->file_path_,
+      /*locator=*/*this->locator_,
+      /*format_escape_errors*/ this->format_escape_errors_);
```
strager, line 62: Nit:
```suggestion
      /*format_escape_errors=*/this->format_escape_errors_);
```

docs/cli.adoc
```diff
@@ -48,6 +48,19 @@ This command has two modes:
 Incompatible with *--lsp-server*.
 --
 
+*--diagnostic-hyperlinks*=_when_::
+  Control whether the errors are using escape sequences to show errors as hyperlinks or not.
+  _when_ is one of the following:
++
+--
+- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On windows
+it behaves like 'never' because windows default terminals don't support escape sequences.
```
strager, line 57: Nit: One line per sentence:
```suggestion
- *auto* (default): shows errors as hyperlinks only if the error output is a terminal.
On windows it behaves like 'never' because windows default terminals don't support escape sequences.
```

strager, line 57: Minor: I wouldn't mention why it behaves line `never` on Windows.
```suggestion
- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On windows
it behaves like 'never'.
```

docs/cli.adoc
```diff
@@ -48,6 +48,19 @@ This command has two modes:
 Incompatible with *--lsp-server*.
 --
 
+*--diagnostic-hyperlinks*=_when_::
+  Control whether the errors are using escape sequences to show errors as hyperlinks or not.
+  _when_ is one of the following:
++
+--
+- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On windows
```
strager, line 56: Nit: Windows is capitalized:
```suggestion
- *auto* (default): shows errors as hyperlinks only if the error output is a terminal. On Windows
```

docs/cli.adoc
```diff
@@ -48,6 +48,19 @@ This command has two modes:
 Incompatible with *--lsp-server*.
 --
 
+*--diagnostic-hyperlinks*=_when_::
+  Control whether the errors are using escape sequences to show errors as hyperlinks or not.
```
strager, line 52: "show errors as hyperlinks" is weird phrasing IMO. Hyperlink is a verb (as well as a noun), so we can say "hyperlink errors".

But we're not really hyperlinking the errors; we're hyperlinking the error codes. So maybe write this:
```suggestion
  Control whether to hyperlink error codes or not.
```

plugin/vscode/quick-lint-js/vscode-error-reporter.h
```diff
@@ -60,7 +60,20 @@ class vscode_error_formatter
...
     });
-    diag.Set("code", ::Napi::String::New(this->env_, code.data(), code.size()));
+
+    ::Napi::Value uri = this->vscode_->uri_class.New(
+        {/*scheme*/ ::Napi::String::New(this->env_, "https"),
+         /*authority*/ ::Napi::String::New(this->env_, "quick-lint-js.com"),
+         /*path*/ ::Napi::String::New(this->env_, "/errors/"),
+         /*query*/ ::Napi::String::New(this->env_, ""),
+         /*fragment*/
+         ::Napi::String::New(this->env_, code.data(), code.size())});
```
strager, line 70: Nit:
```suggestion
        {/*scheme=*/::Napi::String::New(this->env_, "https"),
         /*authority=*/::Napi::String::New(this->env_, "quick-lint-js.com"),
         /*path=*/::Napi::String::New(this->env_, "/errors/"),
         /*query=*/::Napi::String::New(this->env_, ""),
         /*fragment=*/
         ::Napi::String::New(this->env_, code.data(), code.size())});
```

docs/cli.adoc
strager, line 52: I'd also mention that this option is only used if `--output-format=gnu-like`.

strager (approved):

src/main.cpp
```diff
@@ -612,6 +614,11 @@ void print_help_message() {
   print_option("", "gnu-like (default if omitted)");
   print_option("", "vim-qflist-json");
   print_option("", "emacs-lisp");
+  print_option("--diagnostic-hyperlinks=[WHEN]",
+               "Control if errors are shown as hyperlink. WHEN is one of:");
+  print_option("", "auto (default if omitted)");
```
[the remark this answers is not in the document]
ahmafi: Should I remove "if omitted" from "--output-format" too? "Format" can't be omitted there too.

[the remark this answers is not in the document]
strager: > Should I remove "if omitted" from "--output-format" too? "Format" can't be omitted there too.

Yeah. It's there in `--output-format` by mistake.

strager:
Looks awesome. Thanks for improving the completions!

Can you write a commit message for this PR? I plan to squash your commits into one. (We can land separate commits if you really want.)

PR #566, opened 2021-12-07
📝 Update qljs subtitle for wider audience reach

vegerot:
Every React slut knows about HMR.  "linters" are a more arcane and advanced topic.  I think marketing qljs as HMR (which it pretty much exactly is, if you think about it) is the best way to extend the reach and audience of qljs (especially for SEO!)


strager:
You should update the website as well.

Issue #485, opened 2021-11-02
Website: robots.txt


strager:

oneamanish:
see my pull req.

Youness-Asserare:
I have made a pull request for this issue.

PR #591, opened 2021-12-18
Website: robots.txt

Youness-Asserare:
issue #485
I added robots.txt.
I tested this locally using the  [Google's open source robots.txt library](https://github.com/google/robotstxt).

### DIFF
--- a/website/public/robots.txt
+++ b/website/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /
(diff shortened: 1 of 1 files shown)


website/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /
```
strager, line 2: Why do we want to disallow robots from indexing out site? This is the *opposite* of what we want! I want quick-lint-js.com to appear on DuckDuckGo and other search engines.

Or am I misunderstanding what `Disallow: /` means in this context?

strager (changes requested):

strager:
Looks good, thanks! 👍

Issue #583, opened 2021-12-12
10$: CI: test website with Apache


strager:
Catch .htaccess errors

tonysathre:
I can look at this if you can upload the Apache configs.

strager:
I think these are the configs which matter:

apache2.conf:
```
[snip]
# Sets the default security model of the Apache2 HTTPD server. It does
# not allow access to the root filesystem outside of /usr/share and /var/www.
# The former is used by web applications packaged in Debian,
# the latter may be used for local directories served by the web server. If
# your system is serving content from a sub-directory in /srv you must allow
# access here, or in any related virtual host.
<Directory />
	Options FollowSymLinks
	AllowOverride None
	Require all denied
</Directory>

<Directory /usr/share>
	AllowOverride None
	Require all granted
</Directory>

<Directory /var/www/>
	Options Indexes FollowSymLinks
	AllowOverride None
	Require all granted
</Directory>

#<Directory /srv/>
#	Options Indexes FollowSymLinks
#	AllowOverride None
#	Require all granted
#</Directory>




# AccessFileName: The name of the file to look for in each directory
# for additional configuration directives.  See also the AllowOverride
# directive.
#
AccessFileName .htaccess

#
# The following lines prevent .htaccess and .htpasswd files from being
# viewed by Web clients.
#
<FilesMatch "^\.ht">
	Require all denied
</FilesMatch>

[snip]

FileETag INode MTime
```

sites-enabled/003-quick-lint-js.com.conf:
```
[snip]
	ExpiresActive On
	ExpiresDefault A600
	<Files "*.html">
		# To get good server-side analytics, don't cache pages.
		ExpiresDefault A1
	</Files>

	<Directory /var/www/quick-lint-js.com/www>
		AllowOverride FileInfo
	</Directory>
[snip]
```

tonysathre:
51ec98085edd6cf38960f1082a87fc15d4760d1ef9bdb586f374a7684507e615

PR #597, opened 2021-12-24
CI: test website with Apache #583

tonysathre:
Added CI workflow step to test the website using Apache2.

### DIFF
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -65,12 +65,12 @@ jobs:
           sudo apt-get install apache2 -y
           sudo mkdir -p /var/www/quick-lint-js.com/www
           sudo cp -R website/* /var/www/quick-lint-js.com/www
-          sudo sed -i "s,/var/www/html,/var/www/quick-lint-js.com,g" /etc/apache2/sites-enabled/000-default.conf
+          sudo sed -i "s,/var/www/html,/var/www/quick-lint-js.com/www,g" /etc/apache2/sites-enabled/000-default.conf
           sudo service apache2 restart
-          cd ../../
+          pwd
           ./ENV/bin/python ./tools/check-links http://localhost/
+          kill %1
         shell: bash
-
       - name: upload build to workflow
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -37,7 +37,10 @@ jobs:
         run: cd website && yarn build
       - name: bundle site
         run: cd website && zip -r ../website.zip www
-
+      
+      - name: Setup upterm session
+        uses: lhotari/action-upterm@v1
+      
       - name: set up check-links
         run: |
           python3.7 -m venv ENV
--- a/.github/workflows/build-website.yml
+++ b/.github/workflows/build-website.yml
@@ -51,15 +51,6 @@ jobs:
           ./ENV/bin/python ./tools/check-links http://localhost:9001/
           kill %1
         shell: bash
-      - name: check links (build)
-        run: |
-          cd website/www/
-          python3.7 -m http.server 9002 &
-          sleep 0.5  # HACK(strager): Wait for server to start.
-          cd ../../
-          ./ENV/bin/python ./tools/check-links http://localhost:9002/
-          kill %1
-        shell: bash
       - name: check links (apache2)
         run: |
           sudo apt-get install apache2 -y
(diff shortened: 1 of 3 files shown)


tonysathre:

tonysathre:
recheck

tonysathre:
Closes #583

.github/workflows/build-website.yml
```diff
@@ -51,16 +51,15 @@ jobs:
...
-      - name: check links (build)
+      - name: check links (apache2)
         run: |
-          cd website/www/
-          python3.7 -m http.server 9002 &
-          sleep 0.5  # HACK(strager): Wait for server to start.
-          cd ../../
-          ./ENV/bin/python ./tools/check-links http://localhost:9002/
-          kill %1
+          sudo apt-get install apache2 -y
```
strager, line 56: Nit: Include `DEBIAN_FRONTEND=noninteractive`, just in case:
```suggestion
          DEBIAN_FRONTEND=noninteractive sudo apt-get install apache2 -y
```

.github/workflows/build-website.yml
```diff
@@ -51,16 +51,15 @@ jobs:
...
+          sudo mkdir -p /var/www/quick-lint-js.com/www
+          sudo cp -R website/www/* /var/www/quick-lint-js.com/www
+          sudo sed -i "s,/var/www/html,/var/www/quick-lint-js.com/www,g" /etc/apache2/sites-enabled/000-default.conf
```
strager, line 59: Why not copy the files into `/var/www/html` so we don't need to `mkdir` or `sed`? If it's because some files already exist in `/var/www/html`, you can use `rsync`'s `--delete` flag:
```suggestion
          sudo rsync --archive --delete website/www/ /var/www/html/
```

.github/workflows/build-website.yml
```diff
@@ -51,16 +51,15 @@ jobs:
...
+          sudo service apache2 restart
```
strager, line 60: Minor: Would `reload` work? Might be faster.

strager (approved):
Looks nice!

Did you verify that introducing a syntax error in `website/public/.htaccess` causes the job to fail? (This was the original motivation for the task.)

.github/workflows/build-website.yml
tonysathre: I'm not sure if it's an issue with your `check-links` script or something I'm doing, but your script is not failing the workflow when there's a syntax error in the `.htaccess` file.

 See https://github.com/tonysathre/quick-lint-js/runs/4645869307?check_suite_focus=true#step:14:236

tonysathre:
> Looks nice!
> 
> Did you verify that introducing a syntax error in `website/public/.htaccess` causes the job to fail? (This was the original motivation for the task.)

I'm not sure if it's an issue with your `check-links` script or something I'm doing, but your script is not failing the workflow when there's a syntax error in the `.htaccess` file.

 See https://github.com/tonysathre/quick-lint-js/runs/4645869307?check_suite_focus=true#step:14:235

strager:
> I'm not sure if it's an issue with your `check-links` script or something I'm doing, but your script is not failing the workflow when there's a syntax error in the `.htaccess` file.

It looks like the check-links script is not failing on 500 errors. We should fix that!

strager:
Aside from the 500 issue, is this PR basically ready to go?

tonysathre:
> Aside from the 500 issue, is this PR basically ready to go?

I just need to revert some test stuff but yes.

.github/workflows/build-website.yml
```diff
@@ -51,16 +51,14 @@ jobs:
...
         run: |
-          cd website/www/
-          python3.7 -m http.server 9002 &
-          sleep 0.5  # HACK(strager): Wait for server to start.
-          cd ../../
-          ./ENV/bin/python ./tools/check-links http://localhost:9002/
-          kill %1
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install apache2 rsync openssh-client openssh-server -y
+          sudo rsync --archive --delete website/www/ /var/www/html/
+          sudo sed -i -e '/None/ s/AllowOverride\ None/AllowOverride\ FileInfo/' /etc/apache2/apache2.conf
```
strager, line 58: Nit: Is filtering to `None` really necessary? The substitution pattern already mentions `None`:
```suggestion
          sudo sed -i -e 's/AllowOverride\ None/AllowOverride\ FileInfo/' /etc/apache2/apache2.conf
```

.github/workflows/build-website.yml
```diff
@@ -51,16 +51,14 @@ jobs:
...
+          sudo service apache2 start
+          ./ENV/bin/python ./tools/check-links http://localhost/
         shell: bash
-
```
strager, line 63: Nit: Keep this blank line. Undo the other whitespace changes too.

tonysathre:
Fixes #602

tonysathre:
This should be ready to go now.

Issue #578, opened 2021-12-10
7$: Improve error for `else (cond) {}`


strager:
https://stackoverflow.com/questions/41942368/uncaught-syntaxerror-unexpected-token
https://stackoverflow.com/questions/38105591/syntaxerror-unexpected-token

I think this should be easy to do. In `parse_and_visit_if`, after parsing the body of the `else`, we can check if the next token is `{` with `has_leading_newline == false`. If so, report an error.

singalhimanshu:
> if the next token is `{`

shouldn't we report an error if the next token is '(' ?

strager:
> > if the next token is `{`
> 
> shouldn't we report an error if the next token is `(`?

Both, actually.

If the token after `else` is `(` *and* if the token after the body is `{` on the same line, then report an error.

singalhimanshu:
> _and_ if the token after the body is `{` on the same line

I didn't understand this one. Can you give an example?

strager:
No error:
```js
if (x) banana();
else (orange())
{
  stuff();
}
```

No error:
```js
if (x) banana();
else (orange())
stuff();
```

Error:
```js
if (x) banana();
else (orange()) {
  stuff();
}
```

singalhimanshu:
> ```js
> if (x) banana();
> else (orange()) {
>   stuff();
> }
> ```

What should the error highlight in this case?

strager:
> What should the error highlight in this case?

Between `else` and `(`. "missing 'if' after 'else'"

strager:
@singalhimanshu Your patch had a few mistakes which I had to fix myself. See commits d9339b90fd85bf996ed0da81530c6326cd1e63e6 and ce2ba53fade53bf00392258995fce12f7c9113f5.

PR #600, opened 2021-12-28
Improve error for else (cond) {}

singalhimanshu:
Fix: #578

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -927,6 +927,10 @@
               delete_expression))                                              \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_missing_if_after_else, "E0184", { source_code_span expected_if; }, \
+      ERROR(QLJS_TRANSLATABLE("missing if after else"), expected_if))          \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_missing_operator_between_expression_and_arrow_function, "E0063",   \
       { source_code_span where; },                                             \
       ERROR(QLJS_TRANSLATABLE(                                                 \
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -928,7 +928,7 @@
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
       error_missing_if_after_else, "E0184", { source_code_span expected_if; }, \
-      ERROR(QLJS_TRANSLATABLE("missing if after else"), expected_if))          \
+      ERROR(QLJS_TRANSLATABLE("missing 'if' after 'else'"), expected_if))      \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
       error_missing_operator_between_expression_and_arrow_function, "E0063",   \
--- a/src/quick-lint-js/parse.h
+++ b/src/quick-lint-js/parse.h
@@ -2810,10 +2810,15 @@ class parser {
       this->skip();
       const char8 *end_of_else = this->lexer_.end_of_previous_token();
       bool has_left_paren = this->peek().type == token_type::left_paren;
-      parse_and_visit_body();
+      if (has_left_paren) {
+        this->parse_and_visit_expression(v);
+      } else {
+        parse_and_visit_body();
+      }
       bool has_left_curly = this->peek().type == token_type::left_curly;
       if (this->peek().has_leading_newline == false && has_left_paren &&
           has_left_curly) {
+        // if (cond) {} else (cond) {} // Invalid
         this->error_reporter_->report(error_missing_if_after_else{
             .expected_if = source_code_span(end_of_else, end_of_else),
         });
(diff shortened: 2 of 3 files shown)


src/quick-lint-js/error.h
```diff
@@ -926,6 +926,10 @@
       WARNING(QLJS_TRANSLATABLE("redundant delete statement on variable"),     \
               delete_expression))                                              \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_missing_if_after_else, "E0184", { source_code_span expected_if; }, \
+      ERROR(QLJS_TRANSLATABLE("missing if after else"), expected_if))          \
```
strager, line 931: Write quotes around code:
```suggestion
      ERROR(QLJS_TRANSLATABLE("missing 'if' after 'else'"), expected_if))          \
```

test/test-parse-statement.cpp
```diff
@@ -715,6 +715,25 @@ TEST(test_parse, else_without_if) {
   }
 }
 
+TEST(test_parse, missing_if_after_else) {
+  {
+    spy_visitor v;
+    padded_string code(u8"if (false) {} else (true) {}"_sv);
```
strager, line 721: Nit: add test cases:

* should *not* report for `if (false) {} else true {}` (it should complain about `{` instead)
* should *not* report anything for `if (false) {} else (true)NEWLINE{}`

test/test-parse-statement.cpp
```diff
@@ -715,6 +715,25 @@ TEST(test_parse, else_without_if) {
...
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_block_scope",  //
+                                      "visit_exit_block_scope"));
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_OFFSETS(
+                        &code, error_missing_semicolon_after_statement,  //
+                        where, strlen(u8"if (false) {} else (true)"), u8""),
```
strager, line 730: Must fix: The point of error_missing_if_after_else is to replace error_missing_semicolon_after_statement. We should report one or the other, not both.

src/quick-lint-js/parse.h
```diff
@@ -2808,7 +2808,16 @@ class parser {
 
     if (this->peek().type == token_type::kw_else) {
       this->skip();
+      const char8 *end_of_else = this->lexer_.end_of_previous_token();
+      bool has_left_paren = this->peek().type == token_type::left_paren;
       parse_and_visit_body();
+      bool has_left_curly = this->peek().type == token_type::left_curly;
+      if (this->peek().has_leading_newline == false && has_left_paren &&
+          has_left_curly) {
+        this->error_reporter_->report(error_missing_if_after_else{
```
strager, line 2817: Nit: Include a comment with an example code snippet.
```suggestion
        // if (cond) {} else (cond) {}  // Invalid.
        this->error_reporter_->report(error_missing_if_after_else{
```

src/quick-lint-js/parse.h
```diff
@@ -2808,7 +2808,16 @@ class parser {
 
     if (this->peek().type == token_type::kw_else) {
       this->skip();
+      const char8 *end_of_else = this->lexer_.end_of_previous_token();
+      bool has_left_paren = this->peek().type == token_type::left_paren;
       parse_and_visit_body();
```
strager, line 2813: Idea to fix error_missing_semicolon_after_statement:
```suggestion
      if (has_left_paren) {
        // The body of the 'else' is an expression.
        // copy-pasta from https://github.com/quick-lint/quick-lint-js/blob/7d32d9806fc70cc29bcaf5eaea921302860f9418/src/quick-lint-js/parse.h#L394-L395
        this->parse_and_visit_expression(v);
        // Don't do this: parse_expression_end();  // calls this->consume_semicolon();, which is what's reporting the extraneous error
      } else {
        parse_and_visit_body();
      }
```

strager (changes requested):

src/quick-lint-js/parse.h
```diff
@@ -2810,10 +2810,15 @@ class parser {
...
+      } else {
+        parse_and_visit_body();
+      }
       bool has_left_curly = this->peek().type == token_type::left_curly;
       if (this->peek().has_leading_newline == false && has_left_paren &&
           has_left_curly) {
+        // if (cond) {} else (cond) {} // Invalid
         this->error_reporter_->report(error_missing_if_after_else{
             .expected_if = source_code_span(end_of_else, end_of_else),
         });
```
strager, line 2824: Nit: I think this would be easier to understand if it was moved up to after line 2814. What do you think?

strager (approved):
Looks good to me! Let me know when I should merge.

src/quick-lint-js/parse.h
singalhimanshu: I think it makes sense here.

singalhimanshu:
> Looks good to me! Let me know when I should merge.

Ship it :rocket:

Issue #373, opened 2021-06-26
Improve error when assigning to imported variable


strager:
```js
import {snakeSpeed, updateSnake as update, drawSnake as draw, gameover as gameover,length as length, body as body} from "./snake.js"

gameover = true;
```
```
hello.js:3:1: error: assignment to const variable [E003]
hello.js:1:75: note: const variable declared here [E003]
```

quick-lint-js could be more helpful and say "imported variable" instead of "const variable".

`variable_kind` should tell us whether it's an imported variable or a `const`-declared variable.

suryasis-hub:
Hey can this issue be assigned to me?

PR #606, opened 2021-12-31
Lint: Show more specific error when assigning to imported variable

keyehzy:
Hello and happy new year 🎉 !

I'm making a PR for issue #373.

### DIFF
--- a/src/lint.cpp
+++ b/src/lint.cpp
@@ -642,8 +642,10 @@ void linter::report_error_if_assignment_is_illegal(
     }
     break;
   case variable_kind::_import:
-    this->error_reporter_->report(error_assignment_to_imported_variable{
-        *declaration, assignment, kind});
+    if (!is_global_variable) {
+      this->error_reporter_->report(error_assignment_to_imported_variable{
+          *declaration, assignment, kind});
+    }
     break;
   case variable_kind::_catch:
   case variable_kind::_class:
--- a/src/lint.cpp
+++ b/src/lint.cpp
@@ -627,7 +627,6 @@ void linter::report_error_if_assignment_is_illegal(
 
   switch (kind) {
   case variable_kind::_const:
-  case variable_kind::_import:
     if (is_global_variable) {
       this->error_reporter_->report(
           error_assignment_to_const_global_variable{assignment});
@@ -636,14 +635,15 @@ void linter::report_error_if_assignment_is_illegal(
         this->error_reporter_->report(
             error_assignment_to_const_variable_before_its_declaration{
                 *declaration, assignment, kind});
-      } else if (kind == variable_kind::_import) {
-        this->error_reporter_->report(error_assignment_to_imported_variable{
-            *declaration, assignment, kind});
       } else {
         this->error_reporter_->report(
             error_assignment_to_const_variable{*declaration, assignment, kind});
       }
     }
+    break;
+  case variable_kind::_import:
+    this->error_reporter_->report(error_assignment_to_imported_variable{
+        *declaration, assignment, kind});
     break;
   case variable_kind::_catch:
   case variable_kind::_class:
--- a/src/lint.cpp
+++ b/src/lint.cpp
@@ -642,10 +642,14 @@ void linter::report_error_if_assignment_is_illegal(
     }
     break;
   case variable_kind::_import:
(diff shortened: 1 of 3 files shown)


src/lint.cpp
```diff
@@ -636,6 +636,9 @@ void linter::report_error_if_assignment_is_illegal(
         this->error_reporter_->report(
             error_assignment_to_const_variable_before_its_declaration{
                 *declaration, assignment, kind});
+      } else if (kind == variable_kind::_import) {
```
suryasis-hub, line 639: Are we not rechecking the kind parameter, we can simply move the call to the previous "import" call block.

test/test-lint.cpp
```diff
@@ -1047,8 +1047,9 @@ TEST(test_lint, assign_to_mutable_variable_shadowing_immutable_variable) {
 TEST(test_lint, assign_to_immutable_variable) {
   const char8 declaration[] = u8"x";
   const char8 assignment[] = u8"x";
+  variable_kind kind = variable_kind::_const;
 
-  for (variable_kind kind : {variable_kind::_const, variable_kind::_import}) {
+  {
```
suryasis-hub, line 1052: This empty brace can be removed I guess.

test/test-lint.cpp
```diff
@@ -1047,8 +1047,9 @@ TEST(test_lint, assign_to_mutable_variable_shadowing_immutable_variable) {
 TEST(test_lint, assign_to_immutable_variable) {
```
suryasis-hub, line 1047: NIT: This test needs to be renamed probably by "assign_to_immutable_const_variable" keeping it consistent with the naming paradigm we have used for assign_to_immutable_imported_variable.

src/quick-lint-js/error.h
```diff
@@ -63,6 +63,17 @@
           NOTE(QLJS_TRANSLATABLE("const variable declared here"),              \
                declaration))                                                   \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_assignment_to_imported_variable, "E0185",                          \
```
strager, line 67: When I wrote the task, I had in mind that we would reuse E0003 error_assignment_to_const_variable, but customize its message based on `var_kind`.

Making a separate error code seems fine too. :+1:

strager (approved):
Looks good to me. Let me know when I should merge.

src/lint.cpp
```diff
@@ -642,6 +641,12 @@ void linter::report_error_if_assignment_is_illegal(
       }
     }
     break;
+  case variable_kind::_import:
+    if (!is_global_variable) {
```
strager, line 645: Nit: I think this check is unnecessary. `is_global_variable` should always be `false` for `_import` variables. Maybe we could assert:
```suggestion
    QLJS_ASSERT(!is_global_variable);
```

src/quick-lint-js/error.h
strager: Thinking about it more, I think separate error codes makes more sense.

strager (approved):

strager:
Let me know when I should merge.

keyehzy:
Should be good to merge.

src/lint.cpp
keyehzy: I didn't know QLJS_ASSERT worked like a if release mode 😬

keyehzy: Actually, this didn't work. I wrote the `if` explicitly because "GCC Release 8" was failing on dereferencing null pointer in line 646 of lint.cpp, even though we check in line 625. Using QLJS_ASSERT, as you suggested, didn't fix this issue. Maybe the explicit `if` is necessary here?

strager: > Actually, this didn't work. I wrote the `if` explicitly because "GCC Release 8" was failing on dereferencing null pointer in line 646 of lint.cpp, even though we check in line 625.

Sounds like a false positive compiler warning. We can disable the warning. See this for example: https://github.com/quick-lint/quick-lint-js/blob/9463bbe21754fb37ecb997ecea7c1375fdd912b8/src/lint.cpp#L652-L659

I'm fine with the `if` style.

keyehzy:
Should be good for merging.

PR #607, opened 2021-12-31
Add import warnings instead of constant.

### LINKED ISSUE #373: Improve error when assigning to imported variable

suryasis-hub:
Fixes #373 : The idea is to reuse the existing kind checks from the switch cases.

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -71,9 +71,7 @@
       },                                                                       \
       ERROR(QLJS_TRANSLATABLE("assignment to import variable"), assignment)    \
           NOTE(QLJS_TRANSLATABLE("import variable declared here"),             \
-               declaration))
-                                                       \
-                                                                               \
+               declaration))                                                   \
   QLJS_ERROR_TYPE(                                                             \
       error_assignment_to_const_variable_before_its_declaration, "E0004",      \
       {                                                                        \
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -61,8 +61,7 @@
       },                                                                       \
       ERROR(QLJS_TRANSLATABLE("assignment to const variable"), assignment)     \
           NOTE(QLJS_TRANSLATABLE("const variable declared here"),              \
-               declaration))
-  
+               declaration))                                                   \
   QLJS_ERROR_TYPE(                                                             \
       error_assignment_to_import_variable, "E0185",                            \
       {                                                                        \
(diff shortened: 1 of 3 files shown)


suryasis-hub:

suryasis-hub:
recheck

suryasis-hub:
recheck

src/quick-lint-js/error.h
```diff
@@ -62,7 +62,16 @@
       ERROR(QLJS_TRANSLATABLE("assignment to const variable"), assignment)     \
           NOTE(QLJS_TRANSLATABLE("const variable declared here"),              \
                declaration))                                                   \
-                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
+      error_assignment_to_import_variable, "E0185",                            \
```
strager, line 66: When I wrote the task, I had in mind that we would reuse E0003 error_assignment_to_const_variable, but customize its message based on `var_kind`.

Making a separate error code seems fine too. :+1:

test/test-lint.cpp
```diff
@@ -1044,11 +1044,11 @@ TEST(test_lint, assign_to_mutable_variable_shadowing_immutable_variable) {
   EXPECT_THAT(v.errors, IsEmpty());
 }
 
-TEST(test_lint, assign_to_immutable_variable) {
+TEST(test_lint, assign_to_const_variable) {
   const char8 declaration[] = u8"x";
   const char8 assignment[] = u8"x";
 
-  for (variable_kind kind : {variable_kind::_const, variable_kind::_import}) {
+  for (variable_kind kind : {variable_kind::_const}) {
```
strager, line 1051: :+1:

strager (approved):
Looks good to me. Let me know when I should merge this.

strager:
Wait, this looks like a copy of @keyehzy's #606. What happened?

suryasis-hub:
@strager , actually I raised the pull request and later realized there is one already doing something similar. I had put in my comments there so that either pull request should work.

strager:
I'll take #606 instead, since it was created first.

PR #618, opened 2022-01-15
Website: dark mode improvements

erikvimz:
Enables dark mode on native browser controls, like buttons, scrollbars,
textareas, inputs, etc.

Before:
Screenshot 2022-01-15 at 22 33 30

After:
Screenshot 2022-01-15 at 22 40 26

### DIFF
--- a/website/public/main.css
+++ b/website/public/main.css
@@ -215,11 +215,15 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
   a:visited {
     color: #979ef8;
   }
+  :root {
+    color-scheme: dark;
+  }
   html,
   body {
     /* HACK(strager): !important on html is needed to work around quirks in the
        Dark Reader browser extension (https://darkreader.org/). */
     background-color: #002b36 !important;
+    color-scheme: dark;
   }
   body {
     color: white;
--- a/website/public/main.css
+++ b/website/public/main.css
@@ -1,6 +1,10 @@
 /* Copyright (C) 2020  Matthew "strager" Glazar */
 /* See end of file for extended copyright information. */
 
+html {
+  color-scheme: light dark;
+}
+
 body {
   font-family: sans-serif;
   margin-left: auto;
@@ -215,15 +219,11 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
   a:visited {
     color: #979ef8;
   }
-  :root {
-    color-scheme: dark;
-  }
   html,
   body {
     /* HACK(strager): !important on html is needed to work around quirks in the
        Dark Reader browser extension (https://darkreader.org/). */
     background-color: #002b36 !important;
-    color-scheme: dark;
   }
   body {
     color: white;


erikvimz:

erikvimz:
recheck

website/public/main.css
```diff
@@ -215,11 +215,15 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
   a:visited {
     color: #979ef8;
   }
+  :root {
+    color-scheme: dark;
```
strager, line 219: Must fix: Our website supports both dark and light modes. We shouldn't use dark components in light mode.

Based on my [reading of MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme), I think we can write `color-scheme: light dark;` to let the browser choose the right component theme. I think this should match our dark/light theme logic automatically.

website/public/main.css
```diff
@@ -215,11 +215,15 @@ nav.breadcrumbs ol > li:not(:first-child)::before {
...
+  }
   html,
   body {
     /* HACK(strager): !important on html is needed to work around quirks in the
        Dark Reader browser extension (https://darkreader.org/). */
     background-color: #002b36 !important;
+    color-scheme: dark;
```
strager, line 226: Why do we need to specify `color-scheme` twice?

strager (changes requested):

website/public/main.css
erikvimz: Thought is was absolutely necessary, but upon further testing, it seems it is not.

website/public/main.css
erikvimz: We did not use dark components in light mode.

`color-scheme: light dark;` was inside `@media (prefers-color-scheme: dark)`.

strager: > `color-scheme: dark;` was inside `@media (prefers-color-scheme: dark)`.

Ah, I missed this context.

strager (approved):

strager:
Just a note: The issue you're fixing seems to be specific to Chromium-based browsers on macOS. I couldn't reproduce the problem with Firefox on Windows, Linux, or macOS, or with Edge on Windows or Linux.

erikvimz:
@strager Also works on Safari. 🙃

But yeah it does seem to be generally available only in Chromium-based browsers which is strange because according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme#browser_compatibility) it is supported everywhere expect on Internet Explorer. 😳

Screenshot 2022-01-18 at 18 28 02

strager:
> But yeah it does seem to be generally available only in Chromium-based browsers

Before your patch, Chromium was broken, but Firefox was fine. After your patch, Chromium and Firefox are fine. (I didn't test Safari.)

My guess is that Firefox defaults `color-scheme` to `dark light` but Chromium defaults `color-scheme` to `light`. I'm too lazy to experiment further.

PR #624, opened 2022-02-02
improve error messages for stubborn people

Zeebrow:


strager (changes requested):
This should be implemented as a new translation. See docs/TRANSLATING.md. I think the language code would be `en@snarky`.

PR #637, opened 2022-02-09
Add French localization

even4void:
Here is a translation file for French. I follow the instructions in `docs/TRANSLATING.md`, which means I only generate the `.po` file. Please, let me know if there's any other thing I can/should do at this point.

P.S. Regarding my comment on IRC (2022-02-07 21:30 UTC+1), I think I was quite wrong on potential inconsistencies regarding the usage of quotation marks (not brackets as originally written) around reserved JS keywords. It appears that you use quotation marks consistently when JS specific words are used, and you just discard those quotation marks in specific cases (e.g., "let statement" vs. "named 'let'"). I assume this is one of those English subtleties that I am completely unaware of, but I've already been told that, so no worries at all ;-) Sorry for the noise!


po/fr.po
```diff
@@ -0,0 +1,876 @@
+# quick-lint-js finds bugs in JavaScript programs.
```
strager, line 1: Is this Canadian French or France French?

Canadian French: Call this `fr_CA.po`
France French: Call this `fr_FR.po`

strager (commented):
Looks great!

Please regenerate src/translation-table-generated.cpp and test:

1. Run `go run tools/compile-translations.go`
2. Build quick-lint-js locally, or upload your change to this PR then download your build here: https://c.quick-lint-js.com/builds/
3. Run the CLI: `LANG=fr_FR path/to/quick-lint-js file.js`

po/fr.po
even4void: France French. I renamed the translation file as suggested above.

even4void:
I have a local build. I regenerated src/translation-table-generated.cpp, and rebuilt quick-lint-js using Ninja. The test suite is ok. 

I tested the French localization using the following toy example:

```js
const bla = 12;
let bla = 6;

export const accumulate = (collection, operation) => {
  let array = []

  for (let i = 0; i < collection.length, i++) {
    array.push(operation(collection[i));
  }
  return array;
};
```

It looks like it works as expected. Here are the results I got:

```
% LANG=fr_FR.UTF-8 build/quick-lint-js file.js
file.js:2:5: error: redéclaration de variable: bla [E0034]
file.js:1:7: note: variable déjà déclarée ici [E0034]
file.js:7:45: error: troisième argument manquant pour la boucle de style C [E0093]
file.js:8:36: error: crochet d'indexation non apparié [E0055]
% LANG=en_US.UTF-8 build/quick-lint-js file.js
file.js:2:5: error: redeclaration of variable: bla [E0034]
file.js:1:7: note: variable already declared here [E0034]
file.js:7:45: error: C-style for loop is missing its third component [E0093]
file.js:8:36: error: unmatched indexing bracket [E0055]
```

It also works in Neovim (v0.7.0-dev) with nvim-lspconfig, with the new build (screenshot: https://imgur.com/a/kqntVky).

Do you want me to squash the pending changes with the current commit?

```
R  po/fr.po -> po/fr_FR.po
 M src/quick-lint-js/translation-table-generated.h
 M src/translation-table-generated.cpp
```

strager:
> Do you want me to squash the pending changes with the current commit?

Yes please.

strager (approved):

Issue #632, opened 2022-02-03
7$: Report error on statement before first switch case


strager:
```js
switch (true) {
  console.log("banana");  // should error, but doesn't
case true:
case false:
default:
}
```

LeeWannacott:
![image](https://user-images.githubusercontent.com/49783296/152294522-dfe755d8-253c-4996-8ce1-8d4c488fe76a.png)
This is the error tsserver gives for this issue. Might be helpful to whom ever takes on this issue.

singalhimanshu:
I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

strager:
@singalhimanshu You still have #341 claimed and incomplete.

singalhimanshu:
> @singalhimanshu You still have #341 claimed and incomplete.

I lost track on that one, will probably redo it from scratch.

PR #640, opened 2022-02-13
Report error on statement before first switch case

singalhimanshu:
The following code:
```js
switch (true) {
  console.log("banana");
case true:
case false:
default:
}
```

Should report error_statement_before_first_switch_case.

Resolve: #632

### DIFF
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -1189,6 +1189,13 @@
           return_keyword))                                                     \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_statement_before_first_switch_case, "E0198",                       \
+      { source_code_span unexpected_statement; },                              \
+      ERROR(QLJS_TRANSLATABLE("unexpected statement before first switch "      \
+                              "case, expected 'case' or 'default'"),           \
+            unexpected_statement))                                             \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_stray_comma_in_let_statement, "E0036",                             \
       { source_code_span where; },                                             \
       ERROR(QLJS_TRANSLATABLE("stray comma in let statement"), where))         \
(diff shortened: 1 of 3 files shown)


src/quick-lint-js/error.h
```diff
@@ -1188,6 +1188,13 @@
           QLJS_TRANSLATABLE("return statement returns nothing (undefined)"),   \
           return_keyword))                                                     \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_statement_before_first_switch_case, "E0198",                       \
+      { source_code_span unexpected_statement; },                              \
+      ERROR(QLJS_TRANSLATABLE("unexpected statement before first switch "      \
+                              "case, expected 'case' or 'default'"),           \
```
strager, line 1195: Nit: write a semicolon instead of a comma:
```suggestion
                              "case; expected 'case' or 'default'"),           \
```

src/quick-lint-js/parse-statement-inl.h
```diff
@@ -1593,12 +1595,18 @@ void parser::parse_and_visit_switch(Visitor &v) {
...
+      is_before_first_switch_case = false;
       this->skip();
       QLJS_PARSER_UNIMPLEMENTED_IF_NOT_TOKEN(token_type::colon);
       this->skip();
       break;
 
     default: {
+      if (is_before_first_switch_case) {
+        this->error_reporter_->report(error_statement_before_first_switch_case{
+            .unexpected_statement = this->peek().span(),
```
strager, line 1607: Shouldn't we highlight the entire statement, not just its first token?

strager:
Looks good to me. Let me know when I should merge.

src/quick-lint-js/parse-statement-inl.h
singalhimanshu: I think the first token makes sense for example if someone writes this:
```javascript
switch(cond) {
  csae 1: // 'case' misspelled 
    break;
}
```

singalhimanshu:
> Looks good to me. Let me know when I should merge.

Ship it :rocket:

PR #668, opened 2022-03-15
Doc: make contributors and error codes valid markdown.

LeeWannacott:
I copied over the error codes from the website (including descriptions) which might  be overkill?, maybe you only wanted the ones already in the changelog.md :thinking: 

Also I used `<br>` to break lines instead of two white-spaces on the end because harder to see that...

GitHub Flavored Markdown Specification: https://github.github.com/gfm/

Looks like the syntax you had was for reference links:
```
[foo][bar]

[bar]: /url "title"
```
I could do it in the reference style if you want and have description as part of the error code link like it is on the website? :thought_balloon: 
It would be twice the amount of lines; although the lines for errors are really long in this PR; so it would solve that...

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -421,75 +421,255 @@ Beta release.
 
 [Downloads](https://c.quick-lint-js.com/releases/0.2.0/)
 
-[AidenThing]: https://github.com/AidenThing
-[Amir]: https://github.com/ahmafi
-[Christian Mund]: https://github.com/kkkrist
-[Daniel La Rocque]: https://github.com/dlarocque
-[Dave Churchill]: https://www.cs.mun.ca/~dchurchill/
-[David Vasileff]: https://github.com/dav000
-[Erlliam Mejia]: https://github.com/erlliam
-[Himanshu]: https://github.com/singalhimanshu
-[Jenny "Jennipuff" Wheat]: https://twitter.com/jennipaff
-[Jimmy Qiu]: https://github.com/lifeinData
-[Kim "Linden"]: https://github.com/Lindenbyte
-[Lee Wannacott]: https://github.com/LeeWannacott
-[Matheus de Sousa]: https://github.com/keyehzy
-[Nico Sonack]: https://github.com/herrhotzenplotz
-[Piotr Dąbrowski]: https://github.com/yhnavein
-[Shivam Mehta]: https://github.com/maniac-en
-[coc.nvim]: https://github.com/neoclide/coc.nvim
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -421,75 +421,255 @@ Beta release.
...
-[E0196]: https://quick-lint-js.com/errors/E0196/
-[E0197]: https://quick-lint-js.com/errors/E0197/
-[E0198]: https://quick-lint-js.com/errors/E0198/
-[E0199]: https://quick-lint-js.com/errors/E0199/
-[E0203]: https://quick-lint-js.com/errors/E0203/
-[E0205]: https://quick-lint-js.com/errors/E0205/
-[E0207]: https://quick-lint-js.com/errors/E0207/
+## Contributors
+
+- [AidenThing](https://github.com/AidenThing)<br>
```
strager, line 426: Must fix: A visible list doesn't belong in CHANGELOG.md. We already have AUTHORS.md for the contributor list. Undo this change.

docs/CHANGELOG.md
```diff
@@ -421,75 +421,255 @@ Beta release.
...
+- [Piotr Dąbrowski](https://github.com/yhnavein)<br>
+- [Shivam Mehta](https://github.com/maniac-en)<br>
+- [coc.nvim](https://github.com/neoclide/coc.nvim)<br>
+- [config-global-groups](https://quick-lint-js.com/config/#global-groups)<br>
+- [tiagovla](https://github.com/tiagovla)<br>
+- [wagner riffel](https://github.com/wgrr)<br>
+
+## Error codes
+
+- [E0001](https://quick-lint-js.com/errors/E0001/): variable assigned before its declaration<br>
```
strager, line 449: Must fix: A list of error codes doesn't belong in CHANGELOG.md. I don't know what the purpose of this change is for. Undo.

strager (changes requested):
Must fix: I don't understand the point of your change. It [break the links](https://github.com/quick-lint/quick-lint-js/blob/e6c5cb71510b24a60719882d2f85cb096b124feb/docs/CHANGELOG.md) in the content above. See screenshot:

Screen Shot 2022-03-15 at 12 32 56 PM

compared to [before your change](https://github.com/quick-lint/quick-lint-js/blob/61372e366c86e5f4c1b2cf405c060afc20da57fd/docs/CHANGELOG.md):

Screen Shot 2022-03-15 at 12 33 28 PM

> Looks like the syntax you had was for reference links:

Exactly. Why did you break this?

LeeWannacott:
Closing PR as the changelog reference links work as intended...

PR #669, opened 2022-03-16
Feature: github releases from changelog.md

LeeWannacott:
This Pull Request creates a file in `dist/update-release-notes.go` that takes releases from a CHANGELOG.md based on the https://keepachangelog.com/en/1.1.0/ specification and pushes them to GitHub under Releases:

image

Help:  `go run update-release-notes.go --help`

Example of use: `go run update-release-notes.go -Repo=quick-lint/quick-lint-js -TagsRepo=quick-lint/quick-lint-js -AuthToken=$(cat token.txt)`

I ran it on my fork; Here is the result: https://github.com/LeeWannacott/quick-lint-js/releases

---

`ChatGPT's description:` This is a Go (Golang) program that automates the release notes update process for a GitHub repository. The program is intended to be run periodically and automates the process of creating releases and updating release notes by extracting data from a changelog file and comparing it with the repository’s tags and releases.

The program is composed of several functions that work together to complete the following tasks:

Parse command line flags.
Get the path to the changelog file.
Extract data from the changelog file.
Get the tags from the GitHub repository.
Validate that all tags in the changelog have corresponding releases in the repository.
Update releases with changed release notes.
Create missing releases.
The program uses the GitHub API to interact with the repository and create or update releases. It also uses Go concurrency features to speed up the process of creating and updating releases.


strager:
Is it possible to make the timestamps match the dates for historical releases? It currently says "3 hours ago" for each release.

Is this script idempotent? If I run the script multiple times, will it create duplicate releases or will it do nothing?

LeeWannacott:
> Is this script idempotent? If I run the script multiple times, will it create duplicate releases or will it do nothing?

Yes it is idempotent; HTTP put requests may be made to the API to update a release: https://docs.github.com/en/rest/reference/releases#update-a-release

> Is it possible to make the timestamps match the dates for historical releases? It currently says "3 hours ago" for each release.

I don't know :shrug:  the API doesn't seem to provide such an option:
https://docs.github.com/en/rest/reference/releases#create-a-release

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		log.Fatalln(err)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	sb := string(body)
+	var tagsForEachRelease []Tag
+	json.Unmarshal([]byte(sb), &tagsForEachRelease)
```
strager, line 42: Nit: Don't convert from `[]byte` to `string` to `[]byte`:
```suggestion
	var tagsForEachRelease []Tag
	json.Unmarshal(body, &tagsForEachRelease)
```

strager, line 42: Must fix: Check for unmarshalling errors.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+
+type Tag struct {
+	Name       string `json:"name"`
+	ZipballURL string `json:"zipball_url"`
+	TarballURL string `json:"tarball_url"`
+	Commit     struct {
+		SHA string `json:"sha"`
+		URL string `json:"url"`
+	} `json:"commit"`
+	NodeID string `json:"node_id"`
```
strager, line 26: Minor: Delete unused fields:
```suggestion
	Name       string `json:"name"`
	Commit     struct {
		SHA string `json:"sha"`
	} `json:"commit"`
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+			for j := 1; j < numberOfLinesForLastRelease; j++ {
+				releaseBodyLines += changeLogText[versionLineNumber+j] + "\n"
+			}
+		}
+		releaseNotesForEachVersion = append(releaseNotesForEachVersion, releaseBodyLines+contributorsAndErrors)
+	}
+	return releaseNotesForEachVersion
+}
+
+func sendToGitHubAPI(tagForRelease Tag, releaseNote string, versionTitle string, owner string, repo string) {
```
strager, line 102: Nit: Rename this function to describe what it accomplishes. Perhaps `addOrUpdateGitHubRelease`?

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	// https://docs.github.com/en/rest/reference/releases
+	postBody, _ := json.Marshal(map[string]string{
```
strager, line 104: What is `_`? If it's an error, check for it!

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+		"name":             versionTitle,
+		"body":             releaseNote,
+		"target_commitish": tagForRelease.Commit.SHA,
+	})
+	responseBody := bytes.NewBuffer(postBody)
+	url := fmt.Sprintf("https://api.github.com/repos/%v/%v/releases", owner, repo)
+	req, err := http.NewRequest("POST", url, responseBody)
+	req.Header.Set("Accept", "application/vnd.github.v3+json")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "token insert_github_access_token_here")
```
strager, line 115: Use the `flag` package to accept the token on the command line. Or accept a file containing the token.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		panic(err)
```
strager, line 118: Nit: You should be consistent with how you report errors. Use `log.Fatal` like other places in your program.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	return releaseNotesForEachVersion
+}
+
+func sendToGitHubAPI(tagForRelease Tag, releaseNote string, versionTitle string, owner string, repo string) {
+	// https://docs.github.com/en/rest/reference/releases
+	postBody, _ := json.Marshal(map[string]string{
+		"tag_name":         tagForRelease.Name,
+		"name":             versionTitle,
+		"body":             releaseNote,
+		"target_commitish": tagForRelease.Commit.SHA,
```
strager, line 108: Nit: If this isn't needed, delete.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	})
+	responseBody := bytes.NewBuffer(postBody)
+	url := fmt.Sprintf("https://api.github.com/repos/%v/%v/releases", owner, repo)
+	req, err := http.NewRequest("POST", url, responseBody)
+	req.Header.Set("Accept", "application/vnd.github.v3+json")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "token insert_github_access_token_here")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
```
strager, line 117: Does this check for 400 and 500 status codes? Or do we need to do that ourselves?

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+		panic(err)
+	}
+	defer resp.Body.Close()
+	fmt.Println("response Headers:", resp.Header)
+	body, _ := ioutil.ReadAll(resp.Body)
+	fmt.Println(req)
```
strager, line 123: Must fix: Does this possibly leak the auth token? If so, don't!

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	fmt.Println("response Body:", string(body))
```
strager, line 124: Does this pretty-print the response?

Is this really necessary?

strager (changes requested):

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	counterForChangeLogLength := 0
+	var versionLineNumbers []int
+	var changeLogText []string
+	var versionTitlesForEachRelease []string
+	for scanner.Scan() {
+		counterForChangeLogLength++
+		changeLogText = append(changeLogText, scanner.Text())
+		if r.MatchString(scanner.Text()) {
+			// ## 2.3.0 (2022-02-24) => 2.3.0 (2022-02-24)
+			versionNumberAndDateWithoutHashes := scanner.Text()[3:]
```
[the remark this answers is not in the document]
LeeWannacott: FYI: I had to do `## (?P<versionNumberAndDate>\d+\.\d+\.\d+.*)` apparently Go doesn't like `-` in capture groups :safety_pin:

docs/quick-release-notes/main.go
LeeWannacott: Yes, printing `req` contains the authorization token; will remove. `resp.Header` doesn't; although it contains the expiration date of the token and the permissions in `X-Oauth-Scopes`.

docs/quick-release-notes/main.go
LeeWannacott: Not necessary.

docs/quick-release-notes/main.go
LeeWannacott: It's not a `http` `Patch` request though it doesn't `Update` them I called it: `makeGitHubRelease`

docs/quick-release-notes/main.go
LeeWannacott: Yeah it is an error slightly confusing that Unmarshal syntax is:   `error = json.Unmarshal()` and Marshal is `variable, error := json.Marshal()`... :mag: 
`

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	numberOfLinesForLastRelease := 5
+	for i := numberOfLinesForLastRelease + versionLineNumbers[len(versionLineNumbers)-1]; i < changeLogLength; i++ {
+		contributorsAndErrors += changeLogText[i] + "\n"
+	}
+	var releaseNotesForEachVersion []string
+	// exclude Last version (## 0.2.0) with - 1
+	versionLineNumbersLengthMinusOne := len(versionLineNumbers) - 1
+	for i, versionLineNumber := range versionLineNumbers[:] {
+		releaseBodyLines := ""
+		if i < (versionLineNumbersLengthMinusOne) {
```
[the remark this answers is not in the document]
LeeWannacott: `i` doesn't exist in that scope; but I did something similar.

docs/quick-release-notes/main.go
LeeWannacott: https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/net/http/client.go;l=423 I think this is the code for that. I need to research it more...

docs/quick-release-notes/main.go
LeeWannacott: I will do that last otherwise its annoying while testing on my fork because my owner is different for the releases API than it is for tags API. Although I think generally using it you would have the same for both. Depends whether you want the script to be generic; I thought it was just for quick-lint-js :man_shrugging:

docs/quick-release-notes/main.go
strager: > it doesn't `Update` them

I thought you said it is idempotent: https://github.com/quick-lint/quick-lint-js/pull/669#issuecomment-1068888765

docs/quick-release-notes/main.go
strager: > I don't want to have four flag parameters

We could have separate repos with only two flags: `-Repo LeeWannacott/quick-lint-js -TagsRepo quick-lint/quick-lint-js`

> I thought it was just for quick-lint-js 🤷‍♂️

It is. I thought command-line options would be helpful for testing.

docs/quick-release-notes/main.go
LeeWannacott: Maybe I have a different understanding of idempotent. Putting the fancy terminology aside. It does not duplicate the releases if they already exist, nor does it overwrite/update them. Therefore it only makes them if they don't already exist. Just like if a user clicks a `pay now` button on an e-commerce site multiple times it only pays for it once; that is my understanding of idempotent..

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	owner, repo := "quick-lint", "quick-lint-js"
+	tagsForEachRelease := getTagsFromAPI(owner, repo)
+	// POST /repos/{owner}/{repo}/releases
+	if len(releaseNotesForEachVersion) == len(tagsForEachRelease) && len(releaseNotesForEachVersion) == len(versionTitles) {
+		for i := range releaseNotesForEachVersion[:] {
+			sendToGitHubAPI(tagsForEachRelease[i], releaseNotesForEachVersion[i], versionTitles[i], owner, repo)
+		}
+		fmt.Println("Quick release notes finished...")
+	} else {
+		fmt.Println("Error: Release Note versions in changelog.md and Tags from api are different lengths")
```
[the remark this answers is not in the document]
LeeWannacott: I think I may have done this in the dumbest way possible; but it seems to work...

docs/quick-release-notes/main.go
strager: My bad. I think I meant something else when I said 'idempotent'.

I want the following behavior:
* If I run the script twice, the second run should have no effect. (The second run should notice that release were already made. It shouldn't create new GitHub releases.) (It sounds like this is implemented.)
* If I create a tag and an entry in the changelog for that tag, then I run the script, the script should create one new GitHub release. (It sounds like this is implemented.)
* If I edit the changelog for an existing tag/version/release, then I run the script, the script should modify the existing GitHub release. (It sounds like this is **not** implemented.)

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,167 @@
...
+	Commit     struct {
+		SHA string `json:"sha"`
+		URL string `json:"url"`
+	} `json:"commit"`
+	NodeID string `json:"node_id"`
+}
+
+func getTagsFromAPI(owner string, repo string) []Tag {
+	// https://docs.github.com/en/rest/reference/repos#list-repository-tags
+	pathToTags := fmt.Sprintf("https://api.github.com/repos/%v/%v/tags", owner, repo)
```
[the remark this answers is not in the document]
strager: I was suggesting URL-escaping `owner` and `repo` separately. `url.queryEscape("LeeWannacott")` and `url.queryEscape("quick-lint-js")`, not `url.queryEscape("LeeWannacott/quick-lint-js")`.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+		lineCount++
+	}
+	if scanner.Err() != nil {
+		fmt.Println(scanner.Err())
+	}
+	return changeLogInfo{versionLineNumbers, changeLogText, changeLogLength, versionTitles, versionNumbers}
+}
+
+func createReleaseNotes(changeLogInfo changeLogInfo) []string {
+	re, err := regexp.Compile(`^\[.+\]: .+`)
```
strager, line 233: Nit: Rename `re` to something more descriptive like `linkReferenceDefinitionRE` (https://spec.commonmark.org/0.30/#link-reference-definitions).

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+		log.Fatal(err)
+	}
+	return tags
+}
+
+func getChangeLogInfo(scanner *bufio.Scanner) changeLogInfo {
+	re, err := regexp.Compile(`## (?P<versionNumberAndDate>(?P<versionNumber>\d+\.\d+\.\d+).*)`)
+	if err != nil {
+		log.Fatal(err)
+	}
```
strager, line 204: Call [`regexp.MustCompile`](https://pkg.go.dev/regexp#MustCompile) instead, removing the need for the `if`:
```suggestion
	re, err := regexp.MustCompile(`## (?P<versionNumberAndDate>(?P<versionNumber>\d+\.\d+\.\d+).*)`)
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+	lastVersionIdx := len(changeLogInfo.versionLineNumbers) - 1
+	contributorsAndErrors := ""
+	for i := changeLogInfo.versionLineNumbers[lastVersionIdx]; i < changeLogInfo.changeLogLength; i++ {
+		if re.MatchString(changeLogInfo.changeLogText[i]) {
+			contributorsAndErrors += changeLogInfo.changeLogText[i] + "\n"
+		}
+	}
```
strager, line 243: I think we should scan the whole file, not just the last version. I think link reference definitions can appear anywhere in the document.
```suggestion
	for _, line := range changeLogInfo.changeLogText {
		if re.MatchString(line) {
			contributorsAndErrors += line + "\n"
		}
	}
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+	var releaseNotes []string
+	// exclude Last version (## 0.2.0) with - 1
```
strager, line 245: I'm confused by this comment. Is it misplaced?

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+			// Handle last version (## 0.2.0)
+			if versionLineNumber == changeLogInfo.versionLineNumbers[lastVersionIdx] {
+				for j := 1; j < changeLogInfo.changeLogLength-changeLogInfo.versionLineNumbers[lastVersionIdx]; j++ {
+					currentLineOfText := changeLogInfo.changeLogText[versionLineNumber+j]
+					if !re.MatchString(currentLineOfText) {
+						releaseBodyLines += currentLineOfText + "\n"
+					}
+				}
+			}
+		}
```
strager, line 262: Do we need to stripping link reference definitions for only the last version and not other versions? I think stripping them from all versions would be simpler:
```suggestion
		var nextVersionLineNumber int
		if !(i == (lastVersionIdx)) {
			nextVersionLineNumber = changeLogInfo.versionLineNumbers[i+1]
		} else {
			nextVersionLineNumber = changeLogInfo.changeLogLength
		}
		for j := changeLogInfo.versionLineNumbers[i] + 1; j < nextVersionLineNumber; j++ {
			if !re.MatchString(currentLineOfText) {
				releaseBodyLines += changeLogInfo.changeLogText[j] + "\n"
			}
		}
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+	for i := changeLogInfo.versionLineNumbers[lastVersionIdx]; i < changeLogInfo.changeLogLength; i++ {
+		if re.MatchString(changeLogInfo.changeLogText[i]) {
+			contributorsAndErrors += changeLogInfo.changeLogText[i] + "\n"
+		}
+	}
+	var releaseNotes []string
+	// exclude Last version (## 0.2.0) with - 1
+	for i, versionLineNumber := range changeLogInfo.versionLineNumbers[:] {
+		releaseBodyLines := ""
+		if !(i == (lastVersionIdx)) {
```
strager, line 248: Nit:
```suggestion
		if i != lastVersionIdx {
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+			for j := changeLogInfo.versionLineNumbers[i] + 1; j < changeLogInfo.versionLineNumbers[i+1]; j++ {
+				releaseBodyLines += changeLogInfo.changeLogText[j] + "\n"
+			}
+		} else {
+			// Handle last version (## 0.2.0)
+			if versionLineNumber == changeLogInfo.versionLineNumbers[lastVersionIdx] {
+				for j := 1; j < changeLogInfo.changeLogLength-changeLogInfo.versionLineNumbers[lastVersionIdx]; j++ {
+					currentLineOfText := changeLogInfo.changeLogText[versionLineNumber+j]
```
strager, line 256: Nit: I think counting from `changeLogInfo.versionLineNumbers[i] + 1` (like line 249) or `versionLineNumber + 1` makes more sense:
```suggestion
				for j := versionLineNumber + 1; j < changeLogInfo.changeLogLength; j++ {
					currentLineOfText := changeLogInfo.changeLogText[j]
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+		log.Fatal(err)
+	}
+	responseBody := bytes.NewBuffer(postBody)
+	releasesURL := ""
+	if dataForAPI.requestType == "POST" {
+		repoOwner, repoName := splitAndEncodeURLPath(dataForAPI.repoPath)
+		releasesURL = fmt.Sprintf("https://api.github.com/repos/%v/%v/releases", repoOwner, repoName)
+	} else if dataForAPI.requestType == "PATCH" {
+		releasesURL = dataForAPI.urlWithID
+	}
```
strager, line 285: Nit: This `if` looks weird. It means that `urLWithID` is only used sometimes (in a non-obvious way). I think it'd be less code and clearer code to have the caller build the URL.

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,314 @@
...
+	// https://docs.github.com/en/rest/reference/releases
+	postBody, err := json.Marshal(map[string]string{
+		"tag_name": dataForAPI.tagForRelease,
+		"name":     dataForAPI.versionTitle,
+		"body":     dataForAPI.releaseNote,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	responseBody := bytes.NewBuffer(postBody)
```
strager, line 278: Shouldn't this be called `requestBody`, not `responseBody`?

strager:
Must fix: Please include documentation on how to create a token (including required scopes).

strager:
When I run the script, I get a bunch of "Response Headers" messages. Let's hide these messages by default, 'cus they're quite noisy.

strager:
Must fix: The script doesn't check for some kinds of errors, such as missing/invalid auth token or insufficient permissions/scopes. When such an error occurs, the script might fail with a cryptic message, or might succeed with no error reported. I assume GitHub's API returns a helpful error message which the script should forward to the user.

strager:
If there's an "Unreleased" section in CHANGELOG.md, I think the script should issue a warning saying that this section won't be synced to GitHub.

strager (approved):

docs/quick-release-notes/INSTRUCTIONS.md
```diff
@@ -0,0 +1,28 @@
+
+### Instuctions for generating a Personal access token:
+
+1. Navigate to your GitHub account settings.
+2. Then Developer Settings (bottom-most option in the panel on the left).
+3. Click the Personal access tokens menu.
```
strager, line 6: Easier: Link to https://github.com/settings/tokens

docs/quick-release-notes/INSTRUCTIONS.md
```diff
@@ -0,0 +1,28 @@
...
+
+    public_repo Access public repositories
+
+Store the access token in a file and variable of your choosing using cat.
+
+    $ myToken=$(cat example.txt)
+
+### Running quick-release-notes.
+
+    $ go run main.go -Repo=quck-lint/quick-lint-js -TagsRepo=quick-lint/quick-lint-js -AuthToken=$myToken isDraft=false
```
strager, line 19: Nit: No need for a variable:
```suggestion
Store the access token in a file.

### Running quick-release-notes.

    $ go run main.go -Repo=quck-lint/quick-lint-js -TagsRepo=quick-lint/quick-lint-js -AuthToken=$(cat example.txt) isDraft=false 
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,344 @@
+// Copyright (C) 2020  Matthew "strager" Glazar
```
strager, line 1: Nit: rename to dist/update-release-notes.go

docs/quick-release-notes/INSTRUCTIONS.md
```diff
@@ -0,0 +1,28 @@
...
+
+Store the access token in a file and variable of your choosing using cat.
+
+    $ myToken=$(cat example.txt)
+
+### Running quick-release-notes.
+
+    $ go run main.go -Repo=quck-lint/quick-lint-js -TagsRepo=quick-lint/quick-lint-js -AuthToken=$myToken isDraft=false 
+
+Three flags:
```
strager, line 21:
```suggestion
Four flags:
```

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,344 @@
...
+			repoOwner, repoName := splitAndEncodeURLPath(repoPath)
+			releasePath := fmt.Sprintf("https://api.github.com/repos/%v/%v/releases/%v", repoOwner, repoName, release.ID)
+			req, err := http.NewRequest("DELETE", releasePath, nil)
+			req.Header.Set("Accept", "application/vnd.github.v3+json")
+			req.Header.Set("Content-Type", "application/json")
+			req.Header.Set("Authorization", "token "+authToken)
+			if err != nil {
+				log.Fatal(err)
+			}
+			http.DefaultClient.Do(req)
```
strager, line 118: We're not checking for the response status code. Should we?

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,344 @@
...
+			if permission == "public_repo" || permission == "repo" {
+			} else {
+				fmt.Println(redColor + "WARNING: token doesn't include X-Oauth-Scope: public_repo or repo access. " + resetColor)
+			}
+		}
+	} else {
+		fmt.Println(redColor + "WARNING: GitHub access Token has no permissions for X-Oauth-Scopes (select public_repo or repo scopes)" + resetColor)
+	}
+	fmt.Print("Token Rate-limit:", resp.Header["X-Ratelimit-Remaining"])
+	fmt.Println(" Token Expiration:", resp.Header["Github-Authentication-Token-Expiration"])
```
strager, line 325: Why are we logging this?

docs/quick-release-notes/main.go
```diff
@@ -0,0 +1,344 @@
...
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	requestBody := bytes.NewBuffer(postBody)
+	req, err := http.NewRequest(newReleaseRequest.requestType, requestURL, requestBody)
+	req.Header.Set("Accept", "application/vnd.github.v3+json")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "token "+newReleaseRequest.authToken)
+	resp, err := http.DefaultClient.Do(req)
```
strager, line 307: Are we not checking for 404 errors?

strager:
* Add date next to version number in release titles.
* Delete draft feature. Doesn't make sense. I don't know why I suggested it.

strager:
* Only send an update request if the release notes have changed.

strager:
👀

LeeWannacott:
> * Add date next to version number in release titles.
image

Doing date next to version seems to break it for some reason. I think its the space in between version and date; I'll try putting a `-` in between them and see what happens.

LeeWannacott:
> * Add date next to version number in release titles.
strager decided he just wants the version number for the release title.

dist/quick-release-notes/quick-release-notes.go
```diff
@@ -0,0 +1,431 @@
...
+			if releaseVersion.number == tagVersion.Name {
+				releaseVersionHasTag = true
+				tagVersionForMap = tagVersion.Name
+			}
+		}
+		if releaseVersionHasTag == false {
+			fmt.Println(redColor+"WARNING: release", releaseVersion, "missing tag"+resetColor)
+		}
+		if releaseVersionHasTag {
+			releaseMetaData.ReleaseVersionTagMap[releaseVersion.number] = tagVersionForMap
```
strager, line 230: Because `tagVersion.Name` is equal to `releaseVersion.number` (line 223), we could write `releaseVersion.number` on this line and get rid of the `tagVersionForMap` variable.

dist/quick-release-notes/quick-release-notes.go
```diff
@@ -0,0 +1,431 @@
...
+			if tagVersion.Name == releaseVersion.number {
+				tagHasVersionNumber = true
+				releaseVersionForMap = releaseVersion.number
+			}
+		}
+		if tagHasVersionNumber == false {
+			fmt.Println(redColor+"WARNING: tag", tagVersion.Name, "missing changelog entry"+resetColor)
+		}
+		if tagHasVersionNumber {
+			releaseMetaData.TagReleaseVersionMap[tagVersion.Name] = releaseVersionForMap
```
strager, line 247: Same comment as with `tagVersionForMap` above, but with `releaseVersionForMap` this time.

dist/quick-release-notes/quick-release-notes.go
```diff
@@ -0,0 +1,431 @@
...
+		for _, tagVersion := range releaseTagValidationInput.tags[:] {
+			if releaseVersion.number == tagVersion.Name {
+				releaseVersionHasTag = true
+				tagVersionForMap = tagVersion.Name
+			}
+		}
+		if releaseVersionHasTag == false {
+			fmt.Println(redColor+"WARNING: release", releaseVersion, "missing tag"+resetColor)
+		}
+		if releaseVersionHasTag {
```
strager, line 229: This could be refactored into an `if`-`else`.

dist/quick-release-notes/quick-release-notes.go
```diff
@@ -0,0 +1,431 @@
...
+	}
+}
+
+func validateTagsHaveReleases(releaseTagValidationInput releaseTagValidationInput) releaseMetaData {
+	releaseMetaData := releaseMetaData{
+		ReleaseVersionNoteMap: make(map[string]string),
+		ReleaseVersionTagMap:  make(map[string]string),
+		TagReleaseVersionMap:  make(map[string]string),
+	}
+	for i, releaseVersion := range releaseTagValidationInput.changeLog.versions[:] {
```
strager, line 217: I don't think we need the slicing here.
```suggestion
	for i, releaseVersion := range releaseTagValidationInput.changeLog.versions {
```

LeeWannacott:
Fixed latest tag bug: https://github.com/LeeWannacott/quick-lint-js/commit/b01f9ff9a12317a50b42349ac94c61bfb21598bd

dist/update-release-notes.go
```diff
@@ -0,0 +1,449 @@
...
+
+func createReleaseNotes(changeLog changeLog) []string {
+	linkReferenceDefinitionRE, err := regexp.Compile(`^\[.+\]: .+`)
+	if err != nil {
+		log.Fatal(err)
+	}
+	contributorsAndErrors := bytes.Buffer{}
+	for i, line := range changeLog.changeLogText {
+		if linkReferenceDefinitionRE.MatchString(line) {
+			contributorsAndErrors.WriteString(changeLog.changeLogText[i] + "\n")
```
strager, line 359: Nit:
```suggestion
			contributorsAndErrors.WriteString(line + "\n")
```

dist/update-release-notes.go
```diff
@@ -0,0 +1,449 @@
...
+
+func getTagsFromGitHub(tagsRepoPath string) []tag {
+	// https://docs.github.com/en/rest/reference/repos#list-repository-tags
+	tagsRepoOwner, tagsRepoName := splitAndEncodeURLPath(tagsRepoPath)
+	pathToTags := fmt.Sprintf("https://api.github.com/repos/%v/%v/tags", tagsRepoOwner, tagsRepoName)
+	resp, err := http.Get(pathToTags)
+	if err != nil {
+		log.Fatal(err)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
```
strager, line 302: `ioutil.ReadAll` is deprecated. Call `io.ReadAll` instead.
```suggestion
	body, err := io.ReadAll(resp.Body)
```

dist/update-release-notes.go
```diff
@@ -0,0 +1,449 @@
...
+
+func createMissingReleases(releaseMetaData releaseMetaData, authToken string, repoPath string) {
+	createReleaseWaitGroup := &sync.WaitGroup{}
+	createReleaseChannel := make(chan bool, len(releaseMetaData.ReleaseVersionTagMap))
+
+	spawnedThread := false
+	makeLatestRelease := "false"
+	for releaseVersion, tagVersion := range releaseMetaData.ReleaseVersionTagMap {
+		if releaseMetaData.ReleaseVersionTagMap[releaseVersion] == releaseMetaData.TagReleaseVersionMap[tagVersion] {
+			if releaseVersion == releaseMetaData.LatestReleaseVersion {
```
strager, line 159: Nit: Make the `makeLatestRelease` variable as local as possible.
```suggestion
	for releaseVersion, tagVersion := range releaseMetaData.ReleaseVersionTagMap {
		if releaseMetaData.ReleaseVersionTagMap[releaseVersion] == releaseMetaData.TagReleaseVersionMap[tagVersion] {
			var makeLatestRelease string
			if releaseVersion == releaseMetaData.LatestReleaseVersion {
```

dist/update-release-notes.go
```diff
@@ -0,0 +1,449 @@
...
+			createReleaseWaitGroup.Add(1)
+			go updateOrCreateGitHubRelease(postRequest, requestURL, createReleaseChannel, createReleaseWaitGroup)
+			spawnedThread = true
+		}
+	}
+	if spawnedThread {
+		handleChannels(createReleaseChannel, createReleaseWaitGroup)
+	} else {
+		close(createReleaseChannel)
+	}
```
strager, line 184: Why do we have a channel at all? Can't we just call `waitGroup.Wait()` here instead, removing the need for the channels entirely?
```suggestion
	createReleaseWaitGroup.Wait()
```

dist/update-release-notes.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+func createMissingReleases(releaseMetaData releaseMetaData, authToken string, repoPath string) {
+	createReleaseWaitGroup := &sync.WaitGroup{}
+	createReleaseChannel := make(chan bool, len(releaseMetaData.ReleaseVersionTagMap))
+
+	spawnedThread := false
+	makeLatestRelease := "false"
+	for releaseVersion, tagVersion := range releaseMetaData.ReleaseVersionTagMap {
+		if releaseMetaData.ReleaseVersionTagMap[releaseVersion] == releaseMetaData.TagReleaseVersionMap[tagVersion] {
```
strager, line 158: What does this `if` mean? Please add an explanatory variable or at least a comment.

LeeWannacott: I don't think we actually need this if statement. It relates to how we validate tags. But everything we are looping over has been verified/validated anyway; including releaseVersion and tagVersion and we aren't using releaseMetaData.TagReleaseVersionMap[tagVersion] in the POST request. We still need to do the for loops for validating releaseMetaData.TagReleaseVersionMap[tagVersion] in the validateTags function because it will give us a useful warning  if we have a tag, but haven't got a release version in the changelog. The validating tags is complicated and can be confusing...

dist/update-release-notes.go
LeeWannacott: It was unnecessary. :sweat_smile:

dist/update-release-notes.go
LeeWannacott: I removed it; nothing broke :sweat_smile:

strager (approved):

Issue #662, opened 2022-03-09
10$: Report error on ()=>{}()


strager:
The following code has a parse error:
```js
() => { console.log('hello'); }()
```
quick-lint-js accepts this code without error. It should instead report an error stating that parentheses are required around the expression. Here is the desired code:
```js
(() => { console.log('hello'); }())
```

cgsdev0:
is the desired code here actually this?
```javascript
(() => { console.log('hello'); })()
```

I get an error when i paste the current code into my chrome console:
![image](https://user-images.githubusercontent.com/4583705/158919721-acb62634-ca8b-43b3-8a53-6e2a4bc5f0be.png)

strager:
> is the desired code here actually this?

Yes.

cgsdev0:
i'd like to claim this task!

I claim this for-hire task. I expect payment after I complete this task. I will email the quick-lint-js team if I am assigned this task.

PR #672, opened 2022-03-18
Add new error type to address #662

cgsdev0:
This PR addresses #662 by introducing a new error type `E0211` (I'm not sure if there's a convention for these numbers, so I just went with +1)

I decided not to address the following scenario just yet:
```javascript
function() { }()
```

the reason being is that this behaves slightly differently than the arrow functions with where the surrounding parentheses are acceptable. I think this should probably be its own class of error, so the wording doesn't get confusing between the two

if you want, I can add that error type as part of this PR as well (or I can just roll it into `E0211` if you prefer that)

this was fun! thanks for coming by the stream and helping out, much appreciated

### DIFF
--- a/docs/errors/E0211.md
+++ b/docs/errors/E0211.md
@@ -1,0 +1,19 @@
+# E0211: missing parentheses around self-invoked function
+
+Invoking an arrow function immediately without parentheses is a syntax error.
+For example:
+
+```
+() => {
+  console.log('hi');
+}()
+```
+
+To fix this error, add parentheses around the entire function definition, before
+the invocation:
+
+```
+(() => {
+  console.log('hi');
+})()
+```
--- a/src/quick-lint-js/error.h
+++ b/src/quick-lint-js/error.h
@@ -1047,6 +1047,17 @@
                unary_operator))                                                \
                                                                                \
   QLJS_ERROR_TYPE(                                                             \
+      error_missing_parentheses_around_self_invoked_function, "E0211",         \
+      {                                                                        \
+        source_code_span invocation;                                           \
+        source_code_span func_start;                                           \
+      },                                                                       \
+      ERROR(QLJS_TRANSLATABLE(                                                 \
+                "missing parentheses around self-invoked function"),           \
+            invocation)                                                        \
+          NOTE(QLJS_TRANSLATABLE("function starts here"), func_start))         \
+                                                                               \
+  QLJS_ERROR_TYPE(                                                             \
       error_missing_parentheses_around_unary_lhs_of_exponent, "E0194",         \
       {                                                                        \
         source_code_span unary_expression;                                     \
--- a/src/quick-lint-js/parse-expression-inl.h
+++ b/src/quick-lint-js/parse-expression-inl.h
@@ -1149,6 +1149,17 @@ next:
 
   // f(x, y, z)  // Function call.
   case token_type::left_paren:
+    if (binary_builder.last_expression()->kind() ==
+        expression_kind::arrow_function) {
+      // () => {}() // Invalid.
+      auto func_span = binary_builder.last_expression()->span();
+      auto func_start_span =
+          source_code_span(func_span.begin(), func_span.begin());
+      this->error_reporter_->report(
+          error_missing_parentheses_around_self_invoked_function{
+              .invocation = this->peek().span(),
+              .func_start = func_start_span});
+    }
     binary_builder.replace_last(this->parse_call_expression_remainder(
         v, binary_builder.last_expression()));
     goto next;
--- a/test/test-parse-function.cpp
+++ b/test/test-parse-function.cpp
@@ -851,6 +851,74 @@ TEST(test_parse, unnamed_function_expression_without_body) {
   }
 }
 
+TEST(test_parse, arrow_function_invoked_with_parens) {
+  {
+    padded_string code(u8"(() => {})()"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    p.parse_and_visit_module(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+}
+
+TEST(test_parse, async_arrow_function_invoked_with_parens) {
+  {
+    padded_string code(u8"(async () => {})()"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    p.parse_and_visit_module(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+    EXPECT_THAT(v.errors, IsEmpty());
+  }
+}
+
+TEST(test_parse, arrow_function_invoked_no_parens) {
+  {
+    padded_string code(u8"() => {}()"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    p.parse_and_visit_module(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_2_OFFSETS(
+            &code, error_missing_parentheses_around_self_invoked_function,  //
+            func_start, 0, u8"",                                            //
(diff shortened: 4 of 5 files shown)


docs/errors/E0211.md
```diff
@@ -0,0 +1,15 @@
+# E0063: missing parentheses around self-invoked function
+
+Invoking an arrow function immediately without parentheses is a syntax error.
+For example:
+
+```
+() => { console.log('hi'); }()
```
strager, line 7: Nit: I'd format the code to look more like real code:
```suggestion
() => {
  console.log('hi');
}();
```

src/quick-lint-js/error.h
```diff
@@ -789,6 +789,13 @@
       ERROR(QLJS_TRANSLATABLE("missing parameters for arrow function"),        \
             arrow))                                                            \
                                                                                \
+  QLJS_ERROR_TYPE(                                                             \
+      error_missing_parens_self_invoked_function, "E0211",                     \
```
strager, line 793: Nit: Sort alphabetically by name

src/quick-lint-js/parse-expression-inl.h
```diff
@@ -1148,6 +1148,15 @@ expression* parser::parse_expression_remainder(Visitor& v, expression* ast,
 
   // f(x, y, z)  // Function call.
   case token_type::left_paren:
+
+    switch (ast->kind()) {
```
strager, line 1152: Must fix: I think `ast` is wrong here. In most cases, `ast` is the beginning of the expression, not what is called. In the following example, `ast` refers to `x`, not `() => {}`:
```js
x - () => {}()
```
I think you want `binary_builder.last_expression()` instead.

src/quick-lint-js/error.h
```diff
@@ -789,6 +789,13 @@
...
+      { source_code_span func; },                                              \
```
strager, line 794: If the error spans the entire function, that could mean it spans hundreds of lines. Huge error spans lead to a bad user experience.

I think we should report an error like this instead:
```
    () => {}()
            ^ missing parentheses around self-invoked function
    ^ function begins here
```
This would mean storing two empty `source_code_span`s. Kinda dumb, but oh well.

(The CLI error reporting doesn't communicate in this way currently. You need to use an IDE integration to see the spans.)

test/test-parse-function.cpp
```diff
@@ -851,6 +851,37 @@ TEST(test_parse, unnamed_function_expression_without_body) {
...
+  {
+    padded_string code(u8"(() => {})()"_sv);
+    spy_visitor v;
+    parser p(&code, &v);
+    p.parse_and_visit_module(v);
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+    EXPECT_EQ(v.errors.size(), 0);
```
strager, line 864: Nit: The following will print the errors if the assertion fails:
```suggestion
    EXPECT_THAT(v.errors, IsEmpty());
```

test/test-parse-function.cpp
```diff
@@ -851,6 +851,37 @@ TEST(test_parse, unnamed_function_expression_without_body) {
...
+  }
+}
+
+TEST(test_parse, arrow_function_invoked_no_parens) {
+  {
```
strager, line 869: Can you also test `async` arrow functions?

test/test-parse-function.cpp
```diff
@@ -851,6 +851,37 @@ TEST(test_parse, unnamed_function_expression_without_body) {
...
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+
+    EXPECT_THAT(v.errors, ElementsAre(ERROR_TYPE_OFFSETS(
+                              &code, error_missing_parens_self_invoked_function,
+                              func, 0, u8"() => {}")));
+  }
+}
```
strager, line 883: Nit: I'd prefer these tests to be `parse_expression` tests in `test/test-parse-expression.cpp`. `parse_expression` tests can be a bit shorter. But what you wrote is good too.

src/quick-lint-js/parse-expression-inl.h
```diff
@@ -1148,6 +1148,15 @@ expression* parser::parse_expression_remainder(Visitor& v, expression* ast,
 
   // f(x, y, z)  // Function call.
   case token_type::left_paren:
+
```
strager, line 1151: Nit: Delete blank line.

src/quick-lint-js/parse-expression-inl.h
```diff
@@ -1148,6 +1148,15 @@ expression* parser::parse_expression_remainder(Visitor& v, expression* ast,
...
+    switch (ast->kind()) {
+    case expression_kind::arrow_function_with_statements:
+    case expression_kind::arrow_function_with_expression:
+      // () => {}()
```
strager, line 1155:
```suggestion
      // () => {}()  // Invalid.
```

strager (changes requested):
Overall, this looks good. 👍 

Must fix: See CI failures.

cgsdev0:
hi @strager 

sorry for the delay; addressed most of your comments. let me know if you have any more!

test/test-parse-function.cpp
cgsdev0: I didn't make a change for this - let me know if this is a blocker for you

src/quick-lint-js/error.h
```diff
@@ -1046,6 +1046,17 @@
...
+  QLJS_ERROR_TYPE(                                                             \
+      error_missing_parentheses_around_self_invoked_function, "E0211",         \
+      {                                                                        \
+        source_code_span invocation;                                           \
+        source_code_span func_start;                                           \
+      },                                                                       \
+      ERROR(QLJS_TRANSLATABLE(                                                 \
+                "missing parentheses around self-invoked function"),           \
+            invocation)                                                        \
+          NOTE(QLJS_TRANSLATABLE("Function starts here"), func_start))         \
```
strager, line 1058: Nit: lower case
```suggestion
          NOTE(QLJS_TRANSLATABLE("function starts here"), func_start))         \
```

test/test-parse-function.cpp
```diff
@@ -851,6 +851,73 @@ TEST(test_parse, unnamed_function_expression_without_body) {
...
+    EXPECT_THAT(v.visits, ElementsAre("visit_enter_function_scope",
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_2_OFFSETS(
+            &code, error_missing_parentheses_around_self_invoked_function,
+            func_start, 0, u8"", invocation, strlen(u8"() => {}"), u8"(")));
```
strager, line 897: Nit: Add `//` to break this into multiple lines for legibility:
```suggestion
            &code, error_missing_parentheses_around_self_invoked_function, //
            func_start, 0, u8"", //
            invocation, strlen(u8"() => {}"), u8"(")));
```

test/test-parse-function.cpp
```diff
@@ -851,6 +851,73 @@ TEST(test_parse, unnamed_function_expression_without_body) {
...
+                                      "visit_enter_function_scope_body",
+                                      "visit_exit_function_scope",
+                                      "visit_end_of_module"));
+
+    EXPECT_THAT(
+        v.errors,
+        ElementsAre(ERROR_TYPE_2_OFFSETS(
+            &code, error_missing_parentheses_around_self_invoked_function,
+            func_start, 0, u8"", invocation, strlen(u8"async () => {}"),
+            u8"(")));
```
strager, line 917: Nit: Add `//` to break this into multiple lines for legibility:
```suggestion
            func_start, 0, u8"", //
            invocation, strlen(u8"async () => {}"), u8"(")));
```

src/quick-lint-js/parse-expression-inl.h
```diff
@@ -1149,6 +1149,21 @@ expression* parser::parse_expression_remainder(Visitor& v, expression* ast,
 
   // f(x, y, z)  // Function call.
   case token_type::left_paren:
+    switch (binary_builder.last_expression()->kind()) {
+    case expression_kind::arrow_function: {
```
strager, line 1153: Nit: I think an `if` is better than a `switch`:
```suggestion
    if (binary_builder.last_expression()->kind() == expression_kind::arrow_function) {
```

strager:
This PR looks good to me. Let me know when I should merge.

> looks like there are a couple of CI issues that seem unrelated - not sure what to do for those

I fixed those issues in commit a98a5c5df1dcf0e7951ce200b819524d158ef55a. Rebase and you'll get the fix.

strager (approved):

cgsdev0:
Updated to address those comments - should be good to merge now 👍

Issue #514, opened 2021-11-17
5$: Website: remove empty <script> nodes


strager:
We write empty `<script>` nodes in a few places to get Prettier to format server-side JavaScript code in our .ejs.html files. We shouldn't serve the empty `<script>` nodes to clients. Let's strip them from the HTML.

krzysztof-grzybek:
I checked that, and it seems that `prettier` doesn't format javascript in `script` tags while running `yarn fmt`. So what's the requirement for this task?

strager:
> So what's the requirement for this task?

Either:

* Remove `<script>` in the source code (by teaching Prettier about EJS, or finding another workaround), or
* Remove `<script> // </script>` from the generated HTML (by post-processing).

Dreams-Happen:
Is https://github.com/quick-lint/quick-lint-js/blob/95795c6f808fdb8a7c13144db4cf8759fa54046b/website/public/common-nav.ejs.html the only file that has this empty <script>? This seems to be the only .ejs.html file which has one, unless I'm missing something.

strager:
@Dreams-Happen No. Server-only `<script>`s are on many other pages, such as common-footer-nav.ejs.html and blog/syntax-errors-2021/index.ejs.html.

strager:
Done in #717 (commit 9856993fae9809913f121c60b95f5f1c102b43e9).

PR #675, opened 2022-03-23
 5$: Website: remove empty <script> nodes #514 

Youness-Asserare:
the <script>//</script> lines exist in a lot of files mainly inside the tags <head>, <header> and <footer>.
this will prevent <script> nodes from appearing to clients.


strager (changes requested):
The goal of task #514 was to remove the `<script>` elements from the HTML file sent to browsers. Your patch moves in the opposite direction: it makes the browser run more JavaScript code.

Your approach is wrong. We need to remove the scripts on the server/generator, not in the browser/client.

PR #676, opened 2022-03-27
solution for issue#514

Youness-Asserare:
since there is no built in node module that i can use to solve this issue.I used a library called cheerio to remove the script tags.
I added the id's because there is no other way to get cheerio to delete the empty script tags specifically.

### DIFF
--- a/website/src/server.mjs
+++ b/website/src/server.mjs
@@ -8,6 +8,7 @@ import mime from "mime";
 import os from "os";
 import path from "path";
 import { Router, makeHTMLRedirect } from "./router.mjs";
+import * as cheerio from 'cheerio';
 
 export function makeServer({
   esbuildBundles = {},
@@ -50,6 +51,9 @@ export function makeServer({
             path.join(router.wwwRootPath, classifiedDirectory.path),
             { currentURI: request.path }
           );
+          const $ = cheerio.load(out);
+          $('script[id="emptyScript"]').remove();
+          out = $.html();
         } catch (error) {
           response.writeHeader(500, { "content-type": "text/plain" });
           response.end(error.stack);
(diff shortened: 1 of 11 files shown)


website/src/server.mjs
```diff
@@ -50,6 +51,9 @@ export function makeServer({
             path.join(router.wwwRootPath, classifiedDirectory.path),
             { currentURI: request.path }
           );
+          const $ = cheerio.load(out);
+          $('script[id="emptyScript"]').remove();
```
strager, line 55: To find the empty elements, could we do `$('script')` then ask each element's `textContent` to find the empty elements? (Does Cheerio support `textContent` or something similar?)

website/src/server.mjs
```diff
@@ -8,6 +8,7 @@ import mime from "mime";
 import os from "os";
 import path from "path";
 import { Router, makeHTMLRedirect } from "./router.mjs";
+import * as cheerio from 'cheerio';
```
strager, line 11: Nit: Format code by running `yarn fmt`.

website/src/server.mjs
```diff
@@ -50,6 +51,9 @@ export function makeServer({
             path.join(router.wwwRootPath, classifiedDirectory.path),
             { currentURI: request.path }
           );
+          const $ = cheerio.load(out);
```
strager, line 54: Must fix: This code only runs for the development server. It does not run when building the website for production (`yarn build`), but it should.

strager (changes requested):
Must fix: Update `yarn.lock` by running `yarn install`. You may need to install [Yarn](https://classic.yarnpkg.com/lang/en/).

strager:
Must fix: [`yarn test` is failing according to CI.](https://github.com/quick-lint/quick-lint-js/runs/5708470199?check_suite_focus=true)

PR #677, opened 2022-03-28
removed empty script tags for issue#514

Youness-Asserare:
did the following tasks as you asked:
-updated yarn.lock
-made the code that removes script tags better
-ran yarn fmt
-added the code to run-build.mjs so it runs in build


website/src/server.mjs
```diff
@@ -50,6 +51,16 @@ export function makeServer({
...
+          const $ = cheerio.load(out);
+          $("script")
+            .filter(function () {
+              return (
+                this.children[0].data === "\n  //\n" ||
+                this.children[0].data === "\n      //\n    "
+              );
+            })
+            .remove();
+          out = $.html();
```
strager, line 63: Must fix: We shouldn't copy-paste this code. It'll get out of sync easily.

website/src/server.mjs
```diff
@@ -50,6 +51,16 @@ export function makeServer({
             path.join(router.wwwRootPath, classifiedDirectory.path),
             { currentURI: request.path }
           );
+          const $ = cheerio.load(out);
+          $("script")
+            .filter(function () {
+              return (
+                this.children[0].data === "\n  //\n" ||
+                this.children[0].data === "\n      //\n    "
```
strager, line 59: This check doesn't look robust. What if there are more spaces?

Can we check for `//` regardless of the number of spaces?

website/src/server.mjs
```diff
@@ -50,6 +51,16 @@ export function makeServer({
             path.join(router.wwwRootPath, classifiedDirectory.path),
             { currentURI: request.path }
           );
+          const $ = cheerio.load(out);
+          $("script")
+            .filter(function () {
+              return (
+                this.children[0].data === "\n  //\n" ||
```
strager, line 58: Must fix: This code crashes when visiting the benchmarks page (http://127.0.0.1:9001/benchmarks/):
```
TypeError: Cannot read property 'data' of undefined
    at Node.<anonymous> (file:///Users/strager/Projects/quick-lint-js/website/src/server.mjs:58:34)
    at /Users/strager/Projects/quick-lint-js/website/node_modules/cheerio/lib/api/traversing.js:531:48
    at Array.filter (<anonymous>)
    at filterArray (/Users/strager/Projects/quick-lint-js/website/node_modules/cheerio/lib/api/traversing.js:548:17)
    at LoadedCheerio.filter (/Users/strager/Projects/quick-lint-js/website/node_modules/cheerio/lib/api/traversing.js:542:23)
    at serveDirectoryAsync (file:///Users/strager/Projects/quick-lint-js/website/src/server.mjs:56:14)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

Youness-Asserare:
i don't know that specific test is failing. do i have to build before pushing ?

strager (changes requested):
Don't commit `package-lock.json`. Remove it from the commit.

strager:
> i don't know that specific test is failing. do i have to build before pushing ?

That job runs `yarn test` in the `website/` directory. `yarn test` is failing:
```
Running server / serves index.ejs.html ...
Error: Expected '<html><head></head><body>hello 4</body></html>' to be 'hello 4'.
    at <Jasmine>
    at UserContext.<anonymous> (file:///__w/quick-lint-js/quick-lint-js/website/test/test-server.mjs:67:29)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
FAIL    server / serves index.ejs.html
```

website/src/server.mjs
Youness-Asserare: would you consider the following be robust:
` $("script:contains('//')").remove();`
this will remove all the script tags that contains "//"

strager: No. I think that would delete script elements like the following:
```html
<script>
// TODO(strager): Fix this for IE.
fetch("/").then(r => console.log(r));
</script>
```

Youness-Asserare: I think the best thing we can do is assign id's to the empty script tags and  use the following:
` $('script[id="emptyScript"]').remove();`
an exact search  won't work because each script has different number of spaces.

strager: Instead of checking for the presence of `//`, we should check if there is *only* `//` (and optionally surrounding whitespace).

strager:
> can you explain what does this part of the test do ?

1. Create an [EJS template](https://ejs.co/) page
2. Load that page
3. Assert that `<%= 2+2 %>` was interpreted as EJS/JavaScript and replaced with `4`

(I don't know how to explain it besides rephrasing the code.)

website/src/server.mjs
Youness-Asserare: I think this will do exactly what we want:
`this.children[0].data.trim() === "//"`
this will get rid of new lines and compare if the tag has only "//".
I have tested it against many script tags and it seems to work perfectly.

Youness-Asserare:
sorry for delaying this.
I am back working on it.
I only have to prevent that test from failing.
It would be super helpful if you have any information that might help me solve it .

strager:
> I only have to prevent that test from failing.
> It would be super helpful if you have any information that might help me solve it .

Three options:
1. Avoid cheerio
2. Tell cheerio to not create `<html>`, `<head>`, and `<body>` tags
3. Change the test

I prefer option 1.

PR #686, opened 2022-04-12
Fixed #635 and added unit tests

### LINKED ISSUE #635: 10$: Warn on return\n<div/>;

robminer6:
Previously, the code:
```
return
   <div>hi</div>
```
wouldn't report an error. It now correctly reports E0179 when returning an html tag on a new line. I also added unit tests.

### DIFF
--- a/src/quick-lint-js/parse-statement-inl.h
+++ b/src/quick-lint-js/parse-statement-inl.h
@@ -375,6 +375,7 @@ parse_statement:
         case token_type::slash_equal:  // Regular expression.
         case token_type::string:
         case token_type::tilde:
+        case token_type::less:
           if (statement_type == parse_statement_type::any_statement_in_block) {
             this->error_reporter_->report(
                 error_return_statement_returns_nothing{
--- a/src/quick-lint-js/parse-statement-inl.h
+++ b/src/quick-lint-js/parse-statement-inl.h
@@ -368,6 +368,7 @@ parse_statement:
         case token_type::left_curly:  // Object literal.
         case token_type::left_paren:
         case token_type::left_square:  // Array literal.
+        case token_type::less:
         case token_type::minus:
         case token_type::number:
         case token_type::plus:
@@ -375,7 +376,6 @@ parse_statement:
         case token_type::slash_equal:  // Regular expression.
         case token_type::string:
         case token_type::tilde:
-        case token_type::less:
           if (statement_type == parse_statement_type::any_statement_in_block) {
             this->error_reporter_->report(
                 error_return_statement_returns_nothing{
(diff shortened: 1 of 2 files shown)


robminer6:

src/quick-lint-js/parse-statement-inl.h
```diff
@@ -375,6 +375,7 @@ bool parser::parse_and_visit_statement(
         case token_type::slash_equal:  // Regular expression.
         case token_type::string:
         case token_type::tilde:
+        case token_type::less:
```
strager, line 378: Nit: Sort alphabetically; move above line 371.

test/test-parse-statement.cpp
```diff
@@ -155,17 +155,28 @@ TEST(test_parse, return_statement_disallows_newline) {
            u8"typeof banana",
            u8"{}",
            u8"~bits",
+           u8"<div>hi</div>",
+           u8"<a href=\"github.com/quick-lint/quick-lint-js\">link</a>",
+           u8"<p></p>",
+           u8"<div>this is <b>bold</b> and <i>italic</i> text</div>",
+           u8"<head><title>head and title</title></head>",
```
strager, line 162: Nit: I think we need only one or two examples of JSX, not five.

test/test-parse-statement.cpp
```diff
@@ -155,17 +155,28 @@ TEST(test_parse, return_statement_disallows_newline) {
...
            // a statement.)
-           // TODO(strager): JSX.
        }) {
     {
       padded_string code(u8"return\n"s + second_line);
       SCOPED_TRACE(code);
       spy_visitor v;
-      parser p(&code, &v);
-      p.parse_and_visit_module(v);
+      if (second_line[0] == '<') {
```
strager, line 171: Let's always enable JSX for these tests. Make the test logic simple, not complex.

test/test-parse-statement.cpp
```diff
@@ -155,17 +155,28 @@ TEST(test_parse, return_statement_disallows_newline) {
...
+        parser_options options;
+        options.jsx = true;
+        parser p(&code, &v, options);
```
strager, line 174: Minor: We can use the [`jsx_options` variable](https://github.com/quick-lint/quick-lint-js/blob/86cbabefe61baf2f6dd164f36cf362fa1dce76b2/test/quick-lint-js/parse-support.h#L119):
```suggestion
        parser p(&code, &v, jsx_options);
```

strager (approved):
Looks good to me, but the code could be shorter.

Let me know when I should merge.

strager (approved):
Looks great!

PR #698, opened 2022-04-25
Update German Translation files and add simple testing script for translation messages

herrhotzenplotz:
Updates the German po file and recompiled the translation tables.

Also, as discussed on IRC this adds the simple script that can be used to quickly test some of the translated error messages.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -48,59 +48,55 @@ msgstr ""
 
 #: src/diagnostic-formatter.cpp
 msgid "'do-while' loop"
-msgstr ""
+msgstr "do-while-Schleife"
 
 #: src/diagnostic-formatter.cpp
 msgid "'for' loop"
-msgstr ""
+msgstr "for-Schleife"
 
 #: src/diagnostic-formatter.cpp
-#, fuzzy
 msgid "'if' statement"
-msgstr "Körper der if-Anweisung fehlt"
+msgstr "if-Anweisung"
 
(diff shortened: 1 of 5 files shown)


po/de.po
```diff
@@ -40,67 +40,63 @@ msgstr "Variablendeklaration ist hier"
 
 #: benchmark/benchmark-translation.cpp
 msgid "~~~ invalid string, do not use outside benchmark ~~~"
-msgstr ""
+msgstr "~~~ Ungültiger String, nicht außerhalb des Benchmarks verwenden ~~~"
```
strager, line 43: Must fix: Don't translate this string.

po/de.po
```diff
@@ -40,67 +40,63 @@ msgstr "Variablendeklaration ist hier"
...
 
 #: benchmark/benchmark-translation.cpp
 msgid "another invalid string, do not use outside benchmark"
-msgstr ""
+msgstr "weiterer ungültiger String, nicht außerhalb eines Benchmarks verwenden"
```
strager, line 47: Must fix: Don't translate this string.

po/de.po
strager: (Perhaps I should call it "NO DO TRANSLATE: test string for benchmarks")

strager (changes requested):

strager (approved):

PR #700, opened 2022-04-26
Issue#514

Youness-Asserare:
I avoided cheerio as you advised.
I used regex to get rid of all the new lines and removed all the unwanted script tags.
I had to specify the three kinds of script tags (different number of spaces) since there seems to be no way of working around it.


website/package.json
```diff
@@ -13,7 +13,8 @@
     "ejs": "^3.1.6",
     "esbuild-wasm": "^0.13.2",
     "express": "^4.17.1",
-    "jsdom": "17.0.0",
+    "jasmine": "^4.1.0",
+    "jsdom": "^19.0.0",
```
strager, line 17: Are these changes necessary? If not, undo.

website/src/router.mjs
```diff
@@ -157,7 +157,12 @@ export class Router {
   async renderEJSFile(ejsFilePath, { currentURI }) {
     let childProcess = await renderEJSChildProcessPool.takeAsync();
     try {
-      return await childProcess.renderAsync({ currentURI, ejsFilePath });
+      let page = await childProcess.renderAsync({ currentURI, ejsFilePath });
+      let postProcess = page.replaceAll(/\r?\n|\r/g, "");
+      postProcess = postProcess.replaceAll("<script>  //</script>", "");
+      postProcess = postProcess.replaceAll("<script>    //    </script>", "");
+      postProcess = postProcess.replaceAll("<script>      //    </script>", "");
```
strager, line 164: Why not use `*` to match any number of spaces?
```suggestion
      postProcess = postProcess.replaceAll("<script> *//</script>", "");
```

website/src/router.mjs
```diff
@@ -157,7 +157,12 @@ export class Router {
   async renderEJSFile(ejsFilePath, { currentURI }) {
     let childProcess = await renderEJSChildProcessPool.takeAsync();
     try {
-      return await childProcess.renderAsync({ currentURI, ejsFilePath });
+      let page = await childProcess.renderAsync({ currentURI, ejsFilePath });
+      let postProcess = page.replaceAll(/\r?\n|\r/g, "");
```
strager, line 161: Must fix: This line breaks many code blocks. See http://127.0.0.1:9001/errors/E0007/ for example.

strager (changes requested):

PR #726, opened 2022-05-26
Remove annoying shadow in range-console-log.jpg

alex-huff:


alex-huff:

Issue #469, opened 2021-10-13
10$: Show release with notes on GitHub


strager:

strager:
Idea: A Go script which uses GitHub's API to sync docs/CHANGELOG.md to GitHub releases. Dunno how to do attachments, but we can start with just the release notes. Should be pretty self-contained.

LeeWannacott:
I will have a `Go` at this. I have been looking for an excuse to learn more Go and how releases work on GitHub anyway.

strager:
Any progress on this, doge?

pileofneurons:
Can I work on this?

LeeWannacott:
This issue is being worked on: https://github.com/quick-lint/quick-lint-js/pull/669

LeeWannacott:
Landed in: https://github.com/quick-lint/quick-lint-js/pull/669 :eyes:

strager:
Issue #469 fixed in #669 with a hotfix in #966 😅

PR #734, opened 2022-06-01
Automate Create Github "Release"

wilson1375:
Regarding issue [469](https://github.com/quick-lint/quick-lint-js/issues/469), just add a user token that can make releases under `AUTH_TOKEN` and it will create a "release" and "tag" for the latest release of quick-lint with a body/description of the changelog of that version and include the source code zip and tar as downloads. Looks as it does in changelog.

Prerequisites before running at a new released version:
1. docs/changelog.md in the repo should be updated appropriately
2. downloads via the website, specifically "c.quick-lint-js.com/releases/[version]" should be present

### DIFF
--- a/docs/CreateVersion.py
+++ b/docs/CreateVersion.py
@@ -1,0 +1,85 @@
+import requests
+import re
+import sys
+
+def main():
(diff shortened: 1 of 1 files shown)


strager:
Somebody else is already working on issue #469. I will choose his version because it has more features.

Next time, please wait for acknowledgement before picking up a task which has already been assigned to someone else.

PR #749, opened 2022-06-04
Website cache and github action improvement

k1ng440:


.github/workflows/build-website.yml
```diff
@@ -10,48 +10,78 @@ on:
 jobs:
   build:
     name: website
-    runs-on: ubuntu-latest
-    container: ghcr.io/quick-lint/quick-lint-js-github-builder:v3
+    runs-on: ubuntu-22.04
```
strager, line 13: Must fix: Use Docker. I've had enough issues using GitHub's host OS directly that Docker is a better default for this project.

k1ng440: the build is working fine on the host OS but sure if you want me to use docker for it, it will just slow down the workflow.

PR #754, opened 2022-06-16
added cursor change on hover for the paypal button

cmstoddard:
Currently, when you hover over the paypal button for buying the stickers, the cursor doesn't change. Here is the fix for that :)


cmstoddard:

cmstoddard:
recheck

Issue #750, opened 2022-06-08
Website: use HTTP/2


strager:
Our website used to support HTTP/2, but now it doesn't seem to work. Let's enable it again.

strager:
@k1ng440, you might be interested in this task. I think the issue is with mpm_prefork.

k1ng440:
@strager i will take it

k1ng440:
Sorry for the delay on this task. I will work on it later this week.

k1ng440:
After doing some research I found out that `mpm_prefork` is not a good choice anymore and it does not work with HTTP/2 thus `mpm_event` should be used.

k1ng440:
I have confirmed that HTTP/2 is working with `mpm_event`. do you want me to create a docker-compose to run those 2 containers?

strager:
@k1ng440 Why do we need two containers? Can't we edit the existing image?

k1ng440:
The best practice is to dedicate each container to running a single application because docker support only one `ENTRYPOINT` or  `CMD`. We can use `supervisord`, `runit` or something similar to accomplish it

strager:
I'm confused. We only have a single top-level application: Apache httpd. httpd spawns PHP as needed, right? What other top-level program do we need to run explicitly in a container?

k1ng440:
We have to switch to `php-fpm` because `npm_event` does not support mod_php like `npm_prefork` did. PHP-FPM is another process.

strager:
I deployed @k1ng440's patch.

PR #768, opened 2022-06-26
Enable HTTP/2

k1ng440:
Related to https://github.com/quick-lint/quick-lint-js/issues/750

* Switched to mpm_event from mpm_prefork
* Switched to php-fpm from mod_php

### DIFF
--- a/infrastructure/quick-lint-js-web/apache2/apache2.conf
+++ b/infrastructure/quick-lint-js-web/apache2/apache2.conf
@@ -229,5 +229,4 @@ IncludeOptional conf-enabled/*.conf
 # Include the virtual host configurations:
 IncludeOptional sites-enabled/*.conf
 
-
 # vim: syntax=apache ts=4 sw=4 sts=4 sr noet
--- a/infrastructure/quick-lint-js-web/apache2/apache2.conf
+++ b/infrastructure/quick-lint-js-web/apache2/apache2.conf
@@ -220,14 +220,6 @@ CustomLog ${APACHE_LOG_DIR}/matomo_analytics.log matomo_analytics
 
 FileETag INode MTime
 
-# FCGI
-AddHandler fcgid-script .fcgi .php .fpl
-FcgidConnectTimeout 20
-FcgidMaxRequestLen 268435456
-FcgidMaxProcessesPerClass 10
-FcgidIOTimeout 300
-
-
 # Include of directories ignores editors' and dpkg's backup files,
 # see README.Debian for details.
 
--- a/infrastructure/quick-lint-js-web/apache2/apache2.conf
+++ b/infrastructure/quick-lint-js-web/apache2/apache2.conf
@@ -220,6 +220,14 @@ CustomLog ${APACHE_LOG_DIR}/matomo_analytics.log matomo_analytics
 
 FileETag INode MTime
 
+# FCGI
+AddHandler fcgid-script .fcgi .php .fpl
+FcgidConnectTimeout 20
+FcgidMaxRequestLen 268435456
+FcgidMaxProcessesPerClass 10
+FcgidIOTimeout 300
+
+
 # Include of directories ignores editors' and dpkg's backup files,
 # see README.Debian for details.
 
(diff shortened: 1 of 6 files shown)


k1ng440:
@strager please review it whenever you have time.

website/public/phpinfo.php
```diff
@@ -0,0 +1,3 @@
+<?php 
+
+phpinfo();
```
strager, line 3: Must fix: This is a security issue. Delete.

We shouldn't even be serving PHP on https://quick-lint-js.com/. (But that's not your patch's fault.)

infrastructure/quick-lint-js-web/apache2/apache2.conf
```diff
@@ -220,6 +220,14 @@ CustomLog ${APACHE_LOG_DIR}/matomo_analytics.log matomo_analytics
 
 FileETag INode MTime
 
+# FCGI
+AddHandler fcgid-script .fcgi .php .fpl
```
strager, line 224: We only need .php, right?

infrastructure/quick-lint-js-web/run-apache2.sh
```diff
@@ -9,7 +9,11 @@ set -u
 printf '\n============================================================\n'
 printf '\nrunning web server: http://quick-lint-js.com%s/\n' "${DEV_TLD}"
 printf '\n============================================================\n'
-exec apache2ctl -D FOREGROUND
+
+php-fpm8.1 -c /etc/php/8.1/fpm/php-fpm.conf -F &
+apache2ctl -D FOREGROUND &
+
+wait
```
strager, line 16: If we want `wait -n`, we should install Bash in the container and use it for this script. It looks like POSIX sh doesn't support `-n` or an equivalent. =\

infrastructure/quick-lint-js-web/apache2/mods-available/mpm_event.conf
```diff
@@ -6,7 +6,7 @@
 # MaxRequestWorkers: maximum number of worker threads
 # MaxConnectionsPerChild: maximum number of requests a server process serves
 <IfModule mpm_event_module>
-	StartServers			 2
+	StartServers		 2
```
strager, line 9: Nit: Undo unnecessary whitespace changes like this.

infrastructure/quick-lint-js-web/apache2/apache2.conf
strager, line 224: Do we need this directive at all? We already have a similar directive in `fcgid.conf`.

infrastructure/quick-lint-js-web/apache2/mods-available/fcgid.conf
```diff
@@ -0,0 +1,9 @@
+<IfModule mod_fcgid.c>
```
strager, line 1: Is this file used? I don't see an `a2enconf fcgid`.

strager (changes requested):

infrastructure/quick-lint-js-web/apache2/mods-available/fcgid.conf
k1ng440: Yes, it's being used. we are overwriting the file.

infrastructure/quick-lint-js-web/run-apache2.sh
k1ng440: switched to bash and had to use pids with wait -n

website/public/phpinfo.php
k1ng440: done

strager:
I deployed this PR, and HTTP/2 seems to be working.

However, our [PHP app](https://admin.quick-lint-js.com/matomo/) is broken:
>  Deprecated: DateTime::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated in /var/www/admin.quick-lint-js.com/matomo/vendor/monolog/monolog/src/Monolog/Logger.php on line 324

I think a Matomo upgrade will fix it: https://github.com/matomo-org/matomo/issues/17686

strager:
I rolled back and Matomo is still broken. Hmm. Maybe it was always broken. :zany_face:

strager:
It looks like Matomo has been broken since June 8 (i.e. when I switched to the Docker container). I upgraded to Matomo 4.9.1 the UI seems to be working. (Logging is probably broken still.)

strager (approved):

PR #785, opened 2022-07-12
Warn on 'await await'

### LINKED ISSUE #776: 7$: Warn on 'await await'
In an async function, `await await p` is no different than `await p`. Mark the first `await` as redundant.

clegoz:
Resolves #776

### DIFF
--- a/src/quick-lint-js/diagnostic-types.h
+++ b/src/quick-lint-js/diagnostic-types.h
@@ -1381,6 +1381,11 @@
                   original_declaration))                                       \
                                                                                \
   QLJS_DIAG_TYPE(                                                              \
+      diag_redundant_await, "E0262", diagnostic_severity::warning,             \
+      { source_code_span await_operator; },                                    \
+      MESSAGE(QLJS_TRANSLATABLE("redundant 'await'"), await_operator))         \
+                                                                               \
+  QLJS_DIAG_TYPE(                                                              \
       diag_regexp_literal_flags_cannot_contain_unicode_escapes, "E0035",       \
       diagnostic_severity::error, { source_code_span escape_sequence; },       \
       MESSAGE(QLJS_TRANSLATABLE(                                               \
--- a/test/test-parse-expression.cpp
+++ b/test/test-parse-expression.cpp
@@ -888,6 +888,30 @@ TEST_F(test_parse_expression, await_unary_operator_inside_async_functions) {
   }
 }
 
+TEST_F(test_parse_expression, redundant_await) {
+  {
+    test_parser p(u8"await await p"_sv);
+    expression* ast = p.parse_expression();
+    EXPECT_EQ(summarize(ast), "await(await(var p))");
+    EXPECT_THAT(p.errors(),
+                ElementsAre(DIAG_TYPE_OFFSETS(p.code(), diag_redundant_await,
+                                              await_operator, 0, u8"await")));
+  }
+
+  {
+    test_parser p(u8"await await await p"_sv);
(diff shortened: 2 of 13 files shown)


clegoz:

src/quick-lint-js/diagnostic-types.h
```diff
@@ -1980,6 +1980,11 @@
       MESSAGE(QLJS_TRANSLATABLE("continue can only be used inside of a loop"), \
               continue_statement))                                             \
                                                                                \
+  QLJS_DIAG_TYPE(                                                              \
+      diag_redundant_await, "E0260", diagnostic_severity::warning,             \
```
strager, line 1984: Nit: Sort alphabetically by struct name (`diag_redundant_await`).

test/test-parse-expression.cpp
```diff
@@ -886,6 +886,28 @@ TEST_F(test_parse_expression, await_unary_operator_inside_async_functions) {
     EXPECT_EQ(summarize(ast), "await(paren(var x))");
     EXPECT_THAT(p.errors(), IsEmpty());
   }
+
+  {
```
strager, line 890: Nit: I'd put this in a separate `TEST` so it's better documented.

test/test-parse-expression.cpp
```diff
@@ -1126,7 +1149,9 @@ TEST_F(test_parse_expression,
       auto guard = p.parser().enter_function(function_attributes::async);
       expression* ast = p.parse_expression();
       EXPECT_EQ(summarize(ast), test.expected_async_function);
-      EXPECT_THAT(p.errors(), IsEmpty());
+
+      EXPECT_THAT(p.errors(), ::testing::Not(::testing::Contains(DIAG_TYPE(
+                                  diag_await_operator_outside_async))));
```
strager, line 1154: Must fix: Negative assertions are bad and we should avoid them if possible.

If the diagnostics change for only one test case, we should special-case that test case (similar to what we do on line 1126).

test/test-parse-expression.cpp
```diff
@@ -1136,7 +1161,8 @@ TEST_F(test_parse_expression,
       EXPECT_EQ(summarize(ast), test.expected_async_function
                                     ? test.expected_async_function
                                     : test.expected_normal_function);
-      EXPECT_THAT(p.errors(), IsEmpty());
+      EXPECT_THAT(p.errors(), ::testing::Not(::testing::Contains(DIAG_TYPE(
+                                  diag_await_operator_outside_async))));
```
strager, line 1165: Must fix: Same comment as on line 1153-4.

test/test-parse.cpp
```diff
@@ -750,7 +750,8 @@ TEST(test_no_overflow, parser_depth_limit_not_exceeded) {
     parser p(&code, &v);
     bool ok = p.parse_and_visit_module_catching_fatal_parse_errors(v);
     EXPECT_TRUE(ok);
-    EXPECT_THAT(v.errors, IsEmpty());
+    EXPECT_THAT(v.errors, ::testing::Not(::testing::Contains(
+                              DIAG_TYPE(diag_depth_limit_exceeded))));
```
strager, line 754: Note: Generally, negative assertions are bad. In this case, I think it's fine. You can leave this code as-is.

strager (changes requested):

strager:
I fixed the "Homebrew package / macOS 10.15" CI failure on master. Please rebase/merge.

strager (approved):

PR #811, opened 2022-08-01
Fix #741: Warn on 'x.toLowerCase() == "BANANA"'

### LINKED ISSUE #741: 10$: Warn on 'x.toLowerCase() == "BANANA"'
`String#toLowerCase()` (or any reasonable function named `toLowerCase` won't ever return characters in the range `/[A-Z]/`. Report a warning if the result of `toLowerCase` is compared with a string containing upper-case letters.

Similarly for `toUpperCase`.

I don't think we need to check that `x` has a string type.

rol1510:
fixes #741

### DIFF
--- a/src/quick-lint-js/fe/parse.cpp
+++ b/src/quick-lint-js/fe/parse.cpp
@@ -319,28 +319,6 @@ void parser::error_on_pointless_string_compare(
   auto is_comparison_operator = [](string8_view s) {
     return s == u8"=="sv || s == u8"==="sv || s == u8"!="sv || s == u8"!=="sv;
   };
-  auto check = [&](source_code_span op_span, string8_view call,
-                   string8_view literal) {
-    bool lower = call == u8"toLowerCase"sv;
-    bool upper = call == u8"toUpperCase"sv;
-
-    // Don't check if both or neither are found
-    if (lower == upper) {
-      return;
-    }
-
-    if (lower) {
-      if (hasupper(literal)) {
-        this->diag_reporter_->report(
-            diag_pointless_string_comp_contains_upper{op_span});
-      }
-    } else {
-      if (haslower(literal)) {
-        this->diag_reporter_->report(
-            diag_pointless_string_comp_contains_lower{op_span});
-      }
-    }
-  };
 
   for (size_t i = 0; i < ast->child_count() - 1; i++) {
     expression* lhs = ast->child(i);
@@ -362,10 +340,21 @@ void parser::error_on_pointless_string_compare(
         rhs = tmp;
       }
 
-      source_code_span call = lhs->child_0()->variable_identifier().span();
-      source_code_span literal = rhs->span();
+      string8_view call =
+          lhs->child_0()->variable_identifier().span().string_view();
+      string8_view literal = rhs->span().string_view();
 
-      check(op_span, call.string_view(), literal.string_view());
+      if (call == u8"toLowerCase"sv) {
+        if (hasupper(literal)) {
+          this->diag_reporter_->report(
+              diag_pointless_string_comp_contains_upper{op_span});
+        }
+      } else if (call == u8"toUpperCase"sv) {
+        if (haslower(literal)) {
+          this->diag_reporter_->report(
+              diag_pointless_string_comp_contains_lower{op_span});
+        }
+      }
     }
   }
 }
--- a/src/quick-lint-js/fe/parse.cpp
+++ b/src/quick-lint-js/fe/parse.cpp
@@ -314,6 +314,63 @@ void parser::error_on_sketchy_condition(expression* ast) {
   }
 }
 
+void parser::error_on_pointless_string_compare(
+    expression::binary_operator* ast) {
+  auto is_comparison_operator = [](string8_view s) {
+    return s.starts_with(u8"=="sv) || s.starts_with(u8"==="sv) ||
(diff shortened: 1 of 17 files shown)


src/quick-lint-js/fe/parse.cpp
```diff
@@ -314,6 +314,51 @@ void parser::error_on_sketchy_condition(expression* ast) {
...
+
+      // make sure the call is on the "left" and the literal on the "right"
+      if (lhs->kind() == expression_kind::literal) {
+        expression* tmp = lhs;
+        lhs = rhs;
+        rhs = tmp;
+      }
+
+      string8_view call =
+          lhs->child_0()->variable_identifier().span().string_view();
```
strager, line 344: Must fix: This is wrong. You're assuming that the call expression's callee is a dot expression or a variable reference. The following is legal code, but this call to `variable_identifier` will crash:

```javascript
// binary(call(index(dot(var o, m), literal)), literal)
o.m[0]() == 42
```

You should check that `lhs`'s callee (`child_0`) is a `dot` expression before calling `variable_identifier`.

src/quick-lint-js/fe/parse.cpp
```diff
@@ -314,6 +314,51 @@ void parser::error_on_sketchy_condition(expression* ast) {
...
+      string8_view literal = rhs->span().string_view();
+
+      if (call == u8"toLowerCase"sv) {
+        if (hasupper(literal)) {
```
strager, line 348: Must fix: `hasupper` is insufficient. A string literal might contain Unicode escapes. Your code will report a warning in the following code, but it shouldn't:
```js
let s = 'M';
console.log(s.toLowerCase() === "\u006D");
```

We currently don't have a way to get the normalized string (i.e. without escapes) from a string literal. So I think we should bail out if we see any `\`.

src/quick-lint-js/fe/parse.cpp
```diff
@@ -314,6 +314,51 @@ void parser::error_on_sketchy_condition(expression* ast) {
...
+      // make sure the call is on the "left" and the literal on the "right"
+      if (lhs->kind() == expression_kind::literal) {
+        expression* tmp = lhs;
+        lhs = rhs;
+        rhs = tmp;
+      }
+
+      string8_view call =
+          lhs->child_0()->variable_identifier().span().string_view();
+      string8_view literal = rhs->span().string_view();
```
strager, line 345: `rhs` might be a string literal, or it might be a number literal. We currently don't distinguish between the two. This means that your code will report a confusing warning.

I think the correct fix is to have two kinds of literals in our expression AST: string literals and number literals. As a hack, we can check if the first character is \` or `'` or `"`.

test/test-parse-warning.cpp
```diff
@@ -169,6 +169,162 @@ TEST_F(test_error_equals_does_not_distribute_over_or, non_constant) {
     EXPECT_THAT(p.errors, IsEmpty());
   }
 }
+
+TEST_F(test_parse_warning, warn_on_pointless_string_compare) {
+  {
+    padded_string code(u8"toLowerCase() == 'banana'"_sv);
```
strager, line 175: I don't think we want to check a bare call to a function named `toLowerCase`. We want to check a call to a *method* named `toLowerCase`. So I think this test should say `s.toLowerCase()` instead.

test/test-parse-warning.cpp
```diff
@@ -169,6 +169,162 @@ TEST_F(test_error_equals_does_not_distribute_over_or, non_constant) {
...
+    spy_visitor v;
+    parser p(&code, &v);
+    EXPECT_TRUE(p.parse_and_visit_statement(v));
+    EXPECT_THAT(v.errors, IsEmpty());
```
strager, line 179: Nit: I changed the style of tests recently. Use `test_parser` instead of creating a `parser`, `spy_visitor`, and `padded_string` manually.

test/test-parse-warning.cpp
```diff
@@ -169,6 +169,162 @@ TEST_F(test_error_equals_does_not_distribute_over_or, non_constant) {
...
+      EXPECT_TRUE(p.parse_and_visit_statement(v));
+      EXPECT_THAT(
+          v.errors,
+          ElementsAre(DIAG_TYPE_OFFSETS(
+              &code, diag_pointless_string_comp_contains_upper, span_operator,
+              strlen(u8"x.toLowerCase() "), string8(op))));
+    }
+  }
+  {
+    padded_string code(u8"toLowerCase() || 'BANANA'"_sv);
```
strager, line 235: :+1:

strager (changes requested):

test/test-parse-warning.cpp
```diff
@@ -169,6 +169,164 @@ TEST_F(test_error_equals_does_not_distribute_over_or, non_constant) {
     EXPECT_THAT(p.errors, IsEmpty());
   }
 }
+
+TEST_F(test_parse_warning, warn_on_pointless_string_compare) {
+  {
+    test_parser p(u8"s.toLowerCase() == 'banana'"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, IsEmpty());
```
strager, line 177: Nit: No need to capture diags if we are asserting no errors.
```suggestion
    test_parser p(u8"s.toLowerCase() == 'banana'"_sv);
    p.parse_and_visit_statement();
```

strager (approved):

Issue #821, opened 2022-08-12
10$: warn on 'x == []'


strager:
`x == []` is always false. Warn when comparing against an array literal, object literal, class literal, [arrow] function, or regexp literal.

clegoz:
I think this is wrong. It's not always false. 

```
const x = "";
console.log(x == []); // prints true
```

```
const x = 0;
console.log(x == []); // prints true
```

```
class Custom {
  static valueOf() {
    return 1;
  }
}
const x = 1;
console.log(x == Custom); // prints true
```

strager:
> I think this is wrong. It's not always false.

You are correct. We should only warn for `===` or `!==`.

Rebraws:
May I get this issue assigned to me, please? (I do not want the payment).

PR #837, opened 2022-09-10
Fix 821: Warn on x === [], x === {}, etc

Rebraws:
Fixes https://github.com/quick-lint/quick-lint-js/issues/821

Now quick-lint-js gives a warning when using strict equality or inequality operators for comparing against an array literal, object literal, class literal, arrow function, or regexp literal.

### DIFF
--- a/src/quick-lint-js/fe/diagnostic-types.h
+++ b/src/quick-lint-js/fe/diagnostic-types.h
@@ -2530,6 +2530,67 @@
           QLJS_TRANSLATABLE("upper case letters compared with toLowerCase"),    \
           span_operator))                                                       \
                                                                                 \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_strict_comp_against_array_literal, "E0341",                \
+      diagnostic_severity::warning, { source_code_span equals_operator; },      \
+      MESSAGE(                                                                  \
+          QLJS_TRANSLATABLE(                                                    \
+              "using '{0}' against an array literal does not compare items"),   \
+          equals_operator))                                                     \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_comp_against_arrow_function, "E0342",                      \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span equals_operator;                                       \
+        string8_view comparison_result;                                         \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against an arrow "                 \
+                                "function always returns '{1}'"),               \
+              equals_operator, comparison_result))                              \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_comp_against_class_literal, "E0343",                       \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span equals_operator;                                       \
+        string8_view comparison_result;                                         \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against a class "                  \
+                                "literal always returns '{1}'"),                \
+              equals_operator, comparison_result))                              \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_strict_comp_against_empty_array_literal, "E0344",          \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span equals_operator;                                       \
+        string8_view comparison_result;                                         \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("'{0} []' is always '{1}'"), equals_operator,   \
+              comparison_result))                                               \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_comp_against_object_literal, "E0345",                      \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span equals_operator;                                       \
+        string8_view comparison_result;                                         \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against an object "                \
+                                "literal always returns '{1}'"),                \
+              equals_operator, comparison_result))                              \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_pointless_comp_against_regular_expression_literal, "E0346",          \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span equals_operator;                                       \
+        string8_view comparison_result;                                         \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against a regular "                \
+                                "expression literal always returns '{1}'"),     \
+              equals_operator, comparison_result))                              \
   /* END */
 
 // QLJS_X_RESERVED_DIAG_TYPES lists reserved error codes. These codes were used
--- a/src/quick-lint-js/fe/parse.cpp
+++ b/src/quick-lint-js/fe/parse.cpp
@@ -408,6 +408,80 @@ void parser::error_on_invalid_as_const(expression* ast,
   }
 }
 
+void parser::error_on_pointless_compare_against_literal(
+    expression::binary_operator* ast) {
+  auto is_comparison_operator = [](string8_view s) -> bool {
+    return s == u8"=="sv || s == u8"==="sv || s == u8"!="sv || s == u8"!=="sv;
+  };
+
+  for (int i = 0; i < ast->child_count() - 1; i++) {
+    source_code_span op_span = ast->operator_spans_[i];
+    if (is_comparison_operator(op_span.string_view())) {
+      this->check_compare_against_literal(ast->child(i), ast->child(i + 1),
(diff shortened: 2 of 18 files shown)


test/test-parse-warning.cpp
```diff
@@ -327,6 +327,262 @@ TEST_F(test_parse_warning, warn_on_pointless_string_compare_literals) {
...
+  }
+  {
+    test_parser p(u8"x != []"_sv);
+    p.parse_and_visit_expression();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
+}
+
+TEST_F(test_parse_warning,
+       warn_on_pointless_non_strict_compare_against_class_literal) {
```
strager, line 346: Nit: I'd merge all the no-warn test cases together. Maybe call the merged test `non_strict_compare_against_literal_is_not_a_warning`. Then I'd loop over test data. See `test_parse_var.new_style_variables_cannot_be_named_let` in test/test-parse-var.cpp for example.

strager (approved):
Overall, I think your code looks good. But it's not warning in cases it should warn.

src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -2197,6 +2197,41 @@
           QLJS_TRANSLATABLE("upper case letters compared with toLowerCase"),   \
           span_operator))                                                      \
                                                                                \
+  QLJS_DIAG_TYPE(                                                              \
+      diag_pointless_strict_comp_against_array_literal, "E0290",               \
+      diagnostic_severity::warning, { source_code_span span_operator; },       \
```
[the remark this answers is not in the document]
Rebraws: operator is a C++ keyword, maybe comparison_operator could be better?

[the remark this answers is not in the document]
strager: Ah, right.

Yeah, `comparison_operator` is better. Or `equals_operator`, which is what we use for `diag_equals_does_not_distribute_over_or` (E0190).

docs/errors/E0290.md
```diff
@@ -0,0 +1,29 @@
+# E0290: using '===' or '!==' against an array literal always produces the same result
```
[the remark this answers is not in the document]
Rebraws: Yeah, it definitely looks better this way, but when I was fixing this, I was thinking if it could be a good idea to addionally give a warning like `this comparison seems unintentional` when the user uses a non-strict operator. I know that in some cases `x == []` might be true, but I don't think that there's a valid reason to write that (maybe there is, I'm not extremely familiar with js)


So we could have three diagnostics for arrays:
For strict operators (=== and !==):
  - `'{0} []' is always {1}` ('=== []' is always true; '!== []' is always false)
  -  `using '{0}' against an array literal does not compare items`

for non-strict operators (== and !=):
  - `this comparison seems unintentional` 

Do you think it could be a good idea?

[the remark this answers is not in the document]
strager: Let's not warn on `==` or `!=` for now. That can be a different task.

Rebraws:
I think I've addressed all the comments, let me know if anything else needs to be changed or improved

 -  Implemented the two diagnostics for array literals as suggested
 -  Now obj.prop === []  gives a proper warning 
 -  Removed the unnecesary lambda function and removed the overhead in the is_regexp_literal function
 -  Refactor the tests cases to use a for loop
 -  Provided better docs examples and removed the ones without fix suggestions 


I did not provide a warning for `==` and `!=` when compared against arrays literals

src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -2288,6 +2288,67 @@
...
+  QLJS_DIAG_TYPE(                                                              \
+      diag_pointless_comp_against_arrow_function, "E0302",                     \
+      diagnostic_severity::warning,                                            \
+      {                                                                        \
+        source_code_span equals_operator;                                      \
+        string8_view condition;                                                \
+      },                                                                       \
+      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against an arrow "                \
+                                "function always returns '{1}'"),              \
+              equals_operator, condition))                                     \
```
strager, line 2308: Nit: I think `result` or `comparison_result` is a better name than `condition`:
```suggestion
      {                                                                        \
        source_code_span equals_operator;                                      \
        string8_view result;                                                   \
      },                                                                       \
      MESSAGE(QLJS_TRANSLATABLE("using '{0}' against an arrow "                \
                                "function always returns '{1}'"),              \
              equals_operator, result))                                        \
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -402,6 +402,80 @@ void parser::error_on_invalid_as_const(expression* ast,
...
+    expression::binary_operator* ast) {
+  auto is_comparison_operator = [](string8_view s) -> bool {
+    return s == u8"=="sv || s == u8"==="sv || s == u8"!="sv || s == u8"!=="sv;
+  };
+  auto is_strict_operator = [](string8_view op) -> bool {
+    return op == u8"==="sv || op == u8"!=="sv;
+  };
+  auto is_regexp_literal = [](expression::literal* reg_exp) -> bool {
+    return !(reg_exp->span().begin() == reg_exp->span().end() ||
+             *reg_exp->span().begin() != u8'/');
```
strager, line 415: Nit: Avoid calling `span` many times:
```suggestion
    source_code_span span = reg_exp->span();
    return !(span.begin() == span.end() ||
             *span.begin() != u8'/');
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -402,6 +402,80 @@ void parser::error_on_invalid_as_const(expression* ast,
...
+              diag_pointless_comp_against_regular_expression_literal{
+                  .equals_operator = op_span, .condition = condition});
+          goto EXIT_LOOP;
+        }
+        break;
+      default:
+        break;
+      }
+    }
+  EXIT_LOOP:;
```
strager, line 475: Nit: Instead of `goto`, let's make a `parser` member function.

src/quick-lint-js/fe/parse.cpp
```diff
@@ -402,6 +402,80 @@ void parser::error_on_invalid_as_const(expression* ast,
...
+        goto EXIT_LOOP;
+        break;
+      case expression_kind::object:
+        this->diag_reporter_->report(diag_pointless_comp_against_object_literal{
+            .equals_operator = op_span, .condition = condition});
+        goto EXIT_LOOP;
+        break;
+      case expression_kind::literal:
+        if (is_regexp_literal(
+                static_cast<expression::literal*>(child->without_paren()))) {
```
strager, line 464: Minor: Avoid multiple calls to `without_paren`:
```suggestion
      child = child->without_paren();
      switch (child->kind()) {
      case expression_kind::_class:
        this->diag_reporter_->report(diag_pointless_comp_against_class_literal{
            .equals_operator = op_span, .condition = condition});
        goto EXIT_LOOP;
        break;
      case expression_kind::array:
        if (is_strict_operator(op_span.string_view())) {
          if (child->span().string_view() == u8"[]"sv) {
            this->diag_reporter_->report(
                diag_pointless_strict_comp_against_empty_array_literal{
                    .equals_operator = op_span, .condition = condition});
          } else {
            this->diag_reporter_->report(
                diag_pointless_strict_comp_against_array_literal{
                    .equals_operator = op_span});
          }
          goto EXIT_LOOP;
        }
        break;
      case expression_kind::arrow_function:
        this->diag_reporter_->report(diag_pointless_comp_against_arrow_function{
            .equals_operator = op_span, .condition = condition});
        goto EXIT_LOOP;
        break;
      case expression_kind::object:
        this->diag_reporter_->report(diag_pointless_comp_against_object_literal{
            .equals_operator = op_span, .condition = condition});
        goto EXIT_LOOP;
        break;
      case expression_kind::literal:
        if (is_regexp_literal(
                static_cast<expression::literal*>(child))) {
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -402,6 +402,80 @@ void parser::error_on_invalid_as_const(expression* ast,
...
+
+      switch (child->without_paren()->kind()) {
+      case expression_kind::_class:
+        this->diag_reporter_->report(diag_pointless_comp_against_class_literal{
+            .equals_operator = op_span, .condition = condition});
+        goto EXIT_LOOP;
+        break;
+      case expression_kind::array:
+        if (is_strict_operator(op_span.string_view())) {
+          if (child->without_paren()->span().string_view() == u8"[]"sv) {
```
strager, line 440: Must fix: This doesn't work if the array literal has a space or comment inside. Instead of comparing strings, query if the array is any children.
```suggestion
          if (child->without_paren()->child_count() == 0) {
```

strager (changes requested):
See 'must fix' comment. Otherwise, this looks good to me.

I can deal with the rebasing if you want.

Rebraws:
I did the rebase and I've addressed all the comments from the last review, I've also noticed that now the `expression::literal` class  has the `is_regexp` member function, so I deleted the `is_regexp_literal` lambda function and just used `static_cast<expression::literal*>(child)->is_regexp()`

There are some CI errors, but I think they are not related to the PR.

strager (approved):

PR #840, opened 2022-09-12
Fix: Improve check for E0286 and E0287

Rebraws:
Currently, quick-lint-js gives a warning when doing a pointless compare when using toUpperCase or toLowerCase against string literals, but it fails to do so if one of the expressions contains parenthesis e.g. 

`(s.toUpperCase()) === 'test'`

This PR intends to fix that issue and adds some tests to consider those scenarios.

### DIFF
--- a/src/quick-lint-js/fe/parse.cpp
+++ b/src/quick-lint-js/fe/parse.cpp
@@ -324,8 +324,8 @@ void parser::error_on_pointless_string_compare(
   };
 
   for (int i = 0; i < ast->child_count() - 1; i++) {
-    expression* lhs = ast->child(i);
-    expression* rhs = ast->child(i + 1);
+    expression* lhs = ast->child(i)->without_paren();
+    expression* rhs = ast->child(i + 1)->without_paren();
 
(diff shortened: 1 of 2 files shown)


src/quick-lint-js/fe/parse.cpp
```diff
@@ -322,10 +322,14 @@ void parser::error_on_pointless_string_compare(
   auto char_is_a_quote = [](const char8* s) {
     return *s == '"' || *s == '\'' || *s == '`';
   };
+  auto get_child_without_paren = [](expression* child) -> expression* {
+    return child->kind() == expression_kind::paren ? child->without_paren()
+                                                   : child;
+  };
```
strager, line 328: You don't need to check if the child is a paren before calling child->without_paren. Delete this function and just use without_paren directly.

strager (approved):
Thanks!

Rebraws:
Removed the unnecessary lambda function, so it should be ready for merging  👍

strager (approved):

PR #857, opened 2022-10-01
Troubleshoot how to get clangd to correctly find all headers

Guilherme-Vasconcelos:
After I compiled and opened quick-lint-js in my editor, I realized some headers could not be found by my LSP (clangd)

I found out I needed to use `-DCMAKE_EXPORT_COMPILE_COMMANDS=1`, and now clangd can find all headers correctly

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -66,11 +66,11 @@ dependencies:
 
 Run the following command to create a directory called `build`:
 
-    $ mkdir build ; cd build ; cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug .. ; cd "$OLDPWD"
+    $ mkdir build ; cd build ; cmake -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_BUILD_TYPE=Debug .. ; cd "$OLDPWD"
 
 For Ubuntu 18.04 Bionic, instead run the following command:
 
-    $ mkdir build ; cd build ; CC=gcc-8 CXX=g++-8 cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug .. ; cd "$OLDPWD"
+    $ mkdir build ; cd build ; CC=gcc-8 CXX=g++-8 cmake -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_BUILD_TYPE=Debug .. ; cd "$OLDPWD"
 
 #### 2. Build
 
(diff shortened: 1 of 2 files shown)


Guilherme-Vasconcelos:

docs/BUILDING.md
```diff
@@ -290,3 +294,18 @@ To run the tests, execute the test binary in the build directory:
...
+
+---
+
+### Troubleshooting
+
+#### Language server cannot find some files
+
+If your C++ language server cannot find certain headers, you may need to
+configure CMake to export the compile commands by using
+`-DCMAKE_EXPORT_COMPILE_COMMANDS=1`. E.g.:
```
strager, line 306: I think this is an odd place to put this documentation. The solution is already documented on clangd's website: https://clangd.llvm.org/installation#compile_commandsjson

Do you still think it's worth including this language-server-specific documentation in quick-lint-js' docs?

Guilherme-Vasconcelos: I agree that it is not ideal, but I thought it would make sense mostly because it alters the build command, so some people can find odd that "I executed the command to build quick-lint and it didn't work out-of-the-box"

I think it is also worth noticing that some users don't know what language-server they are using behind the curtain. As someone who is new to C++, I didn't know my editor was automatically using clangd, so it took me quite some research to find out how to get these headers to be found properly

Do you think it would make more sense if, instead of pointing exact flags, I changed the docs to make it more clear that there are some extra configurations the developer can do? Something like:

> Run the following command to create a directory called `build`:
>
>    $ mkdir build ; cd build ; cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug .. ; cd "$OLDPWD"
>
> Notice that, depending on your development environment, you may need some extra configurations. Here are some places to get you started:
> - For clangd users: \<insert docs you mentioned>

strager: > I thought it would make sense mostly because it alters the build command

That's a good point.

> some people can find odd that "I executed the command to build quick-lint and it didn't work out-of-the-box"

In that case, why not include `=DCMAKE_EXPORT_COMPILE_COMMANDS=1` in all documented `cmake` invocations, removing the need for this troubleshooting section? Is there any downside to enabling `=DCMAKE_EXPORT_COMPILE_COMMANDS=1` in all builds?

Guilherme-Vasconcelos: I didn't find any downsides that seem relevant to quick-lint in [cmake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html), so I believe including the flag in all invocations will be a better solution indeed

strager (approved):

strager:
Thanks for improving our docs!

Issue #859, opened 2022-10-07
Windows-x86 flaky test failure: internal check failed in flush_impl: false


strager:
https://github.com/quick-lint/quick-lint-js/actions/runs/3202094600/jobs/5230725143

```
test 1
    Start 1: quick-lint-js-test

1: Test command: D:\a\quick-lint-js\quick-lint-js\build\test\quick-lint-js-test.exe
1: Working Directory: D:/a/quick-lint-js/quick-lint-js/build/test
1: Test timeout computed to be: 1500
1: Running main() from gmock_main.cc
1: [==========] Running 1879 tests from 145 test suites.
1: [----------] Global test environment set-up.
1: [----------] 3 tests from test_write_integer
1: [ RUN      ] test_write_integer.common_integers
1: [       OK ] test_write_integer.common_integers (0 ms)
[snip]
1: [ RUN      ] typescript/test_parse_conditional_expression.conditional_expression_with_missing_colon_and_false_component/0
1: [       OK ] typescript/test_parse_conditional_expression.conditional_expression_with_missing_colon_and_false_component/0 (0 ms)
1: [----------] 5 tests from typescript/test_parse_conditional_expression (0 ms total)
1: 
1: [----------] Global test environment tear-down
1: [==========] 1879 tests from 145 test suites ran. (9726 ms total)
1: [  PASSED  ] 1879 tests.
1: 
1:   YOU HAVE 1 DISABLED TEST
1: 
1: D:/a/quick-lint-js/quick-lint-js/src/quick-lint-js/io/output-stream.cpp:101: internal check failed in flush_impl: false
1: quick-lint-js crashed. Please report this bug here:
1: https://quick-lint-js.com/crash-report/
1/2 Test #1: quick-lint-js-test ...............***Failed    9.77 sec
```

strager:
Saw it again: https://github.com/quick-lint/quick-lint-js/actions/runs/3203555922/jobs/5233792532

strager:
Release build failure: https://github.com/quick-lint/quick-lint-js/actions/runs/3212775439/jobs/5251931531

strager:
More info: https://github.com/quick-lint/quick-lint-js/actions/runs/3233883823/jobs/5330148207

> 1: fatal: file_output_stream::flush_impl failed to write data: The handle is invalid.
[...]

PR #862, opened 2022-10-09
fix: Disallow class and lexical declarations in label body

### LINKED ISSUE #193: LabelledStatement should allow only Statement or FunctionDeclaration, not ClassDeclaration or LexicalDeclaration
legal:
```javascript
mylabel: var x;
otherlabel: function f() {}
```

illegal:
```javascript
mylabel: const x = 3;
otherlabel: let y;
thirdlabel: class C {}
```

quick-lint-js thinks all of the above examples are legal.

We already check for `const`, `let`, and `class` in bodies of `with` statements. We could probably reuse that code for labels.

Guilherme-Vasconcelos:
Closes #193


test/test-parse-var.cpp
```diff
@@ -1986,6 +1986,37 @@ TEST_F(test_parse_var,
...
+          expected_body,
+          offsets_matcher(p.code, strlen(u8"l:"), u8""),
+          declaring_keyword,
+          offsets_matcher(p.code, strlen(u8"l: "), variable_kind))));
+  }
+}
+
+TEST_F(test_parse_var,
+       var_declaration_as_label_body_is_allowed) {
+  test_parser p(u8"l: var x = y;", capture_diags);
```
Guilherme-Vasconcelos, line 2012: question: Why don't tests emit "use of undeclared variable: y"? We used `x = y` in a bunch of tests, but I don't see `y` being declared anywhere (or E0057 being ignored)

strager: 'Use of undeclared variable' warnings don't get triggered by the `parser` class (which `test_parser` wraps). It's the `variable_analyzer` class which emits those warnings.

Your test is only running the `parser`, not the `variable_analyzer`, so such a warning shouldn't happen. That's normal for all the parser tests. (Most of the tests would be harder to follow if they did run the `variable_analyzer` too.)

docs/errors/E0149.md
```diff
@@ -8,7 +8,7 @@
 }
 ```
 
-`do`-`while` loops, `if` statements, `for` loops, `while` loops, and `with`
+`do`-`while` loops, `if` statements, `for` loops, `while` loops, `with` and labelled
 statements require a statement or list of statements for the body. It is a
 syntax error to write a class as the body of such a statement:
```
strager, line 13: Thinking aloud: Hmm, I think it's weird to talk about the 'body' for a labelled statement. We could make a new pair of diagnostic types to make the wording smoother, but I don't know if that's worth it.

test/test-parse-class.cpp
```diff
@@ -1174,6 +1174,24 @@ TEST_F(test_parse_class, class_statement_as_with_statement_body_is_disallowed) {
...
+                                      "visit_enter_class_scope_body",  // C
+                                      "visit_exit_class_scope",        // }
+                                      "visit_variable_declaration"));  // C
+    EXPECT_THAT(
+        p.errors,
+        ElementsAre(DIAG_TYPE_3_FIELDS(
+            diag_class_statement_not_allowed_in_body, kind_of_statement,
+            statement_kind::labelled_statement, expected_body,
+            offsets_matcher(p.code, strlen(u8"l:"), u8""), class_keyword,
+            offsets_matcher(p.code, strlen(u8"l: "), u8"class"))));
```
strager, line 1191: Nit: Add empty comments to improve word wrapping after running clang-format:
```suggestion
            diag_class_statement_not_allowed_in_body, //
            kind_of_statement, statement_kind::labelled_statement, //
            expected_body, offsets_matcher(p.code, strlen(u8"l:"), u8""), //
            class_keyword, offsets_matcher(p.code, strlen(u8"l: "), u8"class"))));
```

test/test-parse-function.cpp
```diff
@@ -1658,6 +1658,16 @@ TEST_F(test_parse_function, function_as_with_statement_body_is_disallowed) {
...
 }
 
+TEST_F(test_parse_function, function_as_label_body_is_allowed) {
+  test_parser p(u8"l: function f() {}"_sv, capture_diags);
+  p.parse_and_visit_statement();
+  EXPECT_THAT(p.visits, ElementsAre("visit_variable_declaration",
+                                    "visit_enter_function_scope",
+                                    "visit_enter_function_scope_body",
+                                    "visit_exit_function_scope"));
+  EXPECT_THAT(p.errors, IsEmpty());
```
strager, line 1668: Nit: Omit `capture_diags` if you're asserting no errros:
```suggestion
  test_parser p(u8"l: function f() {}"_sv);
  p.parse_and_visit_statement();
  EXPECT_THAT(p.visits, ElementsAre("visit_variable_declaration",
                                    "visit_enter_function_scope",
                                    "visit_enter_function_scope_body",
                                    "visit_exit_function_scope"));
```

test/test-parse-var.cpp
```diff
@@ -1986,6 +1986,32 @@ TEST_F(test_parse_var,
...
+            diag_lexical_declaration_not_allowed_in_body, kind_of_statement,
+            statement_kind::labelled_statement, expected_body,
+            offsets_matcher(p.code, strlen(u8"l:"), u8""), declaring_keyword,
+            offsets_matcher(p.code, strlen(u8"l: "), variable_kind))));
+  }
+}
+
+TEST_F(test_parse_var, var_declaration_as_label_body_is_allowed) {
+  test_parser p(u8"l: var x = y;", capture_diags);
+  SCOPED_TRACE(p.code);
```
strager, line 2008: Nit: No need to SCOPED_TRACE if the code is a string literal.
```suggestion
```

strager (approved):

strager:
This looks good to me. Let me know when I should merge.

MSVC test failure is not your fault. See https://github.com/quick-lint/quick-lint-js/issues/859

test/test-parse-var.cpp
Guilherme-Vasconcelos: That makes sense! Thanks

test/test-parse-function.cpp
Guilherme-Vasconcelos: Fixed! I actually had a question about this `capture_diags` too: I don't think I fully understand what it does and how it works. I looked at the code at `parse-support.h` and `parse-support.cpp` and realized it is present in two constructors, but I don't see it being actually used

docs/errors/E0149.md
Guilherme-Vasconcelos: I actually had the same line of thought, but I couldn't think of a word much better than 'body'. I thought about 'postfix', but maybe some users will find that more confusing

Guilherme-Vasconcelos:
> This looks good to me. Let me know when I should merge.
I just addressed your last comments. I think it should be ok to merge now

test/test-parse-function.cpp
strager: `capture_diags` is a tag without any state. If `capture_diags` is *not* set, reporting any diagnostic will fail the test. If `capture_diags` is set, diagnostics will be saved in the `p.errors` vector, so you should check them later with `EXPECT_THAT(p.errors, ...)`.

Guilherme-Vasconcelos: Oh, so I don't even need to assert that `p.errors` is empty, that's very neat!

I have changed the code to remove the assertions in these cases

Guilherme-Vasconcelos:
I have made one extra change (explained [here](https://github.com/quick-lint/quick-lint-js/pull/862#discussion_r991639002)), I think now it should actually be ok to merge

strager (approved):

PR #864, opened 2022-10-11
829 asan recommended strict flags

AlekLefebvre:
As suggested by https://github.com/google/sanitizers/wiki/AddressSanitizer#faq


strager:
Fixed in commit e6716d5bf7c050b0915d6baefc87718dfe2cd7c3. Thanks for being proactive with these changes.

Issue #744, opened 2022-06-02
10$: Warn on duplicate cases in switch


strager:
```js
switch (a) {
  case "one":
    break;
  case "one":  // warning
    break;
  case "three":
    break;
}
```

Rebraws:
Is using a `hash_set<string8_view>` a good approach for this issue?

I was thinking about using the hash_set and then in this else case https://github.com/quick-lint/quick-lint-js/blob/0f9d8c623cd5a8fc5c2848deb53641f49defdddd/src/quick-lint-js/fe/parse-statement.cpp#L1853 do something like this:

```
} else {  
        string8_view case_expression = this->peek().span().string_view();
        if (the_hash_set.contains(case_expression)) {
            // GIVE THE WARNING
         }
        the_hash_set.insert(case_expression);
        this->parse_and_visit_expression(
            v, precedence{
                   .colon_type_annotation = allow_type_annotations::never,
               });
        QLJS_PARSER_UNIMPLEMENTED_IF_NOT_TOKEN(token_type::colon);
        this->skip();
      }
```

strager:
> `string8_view case_expression = this->peek().span().string_view();`

This looks wrong. You're only comparing the token after `case`, but there might be multiple tokens. For example, this should not report a warning:
```javascript
switch (x) {
case MyEnum.A:
case MyEnum.B:
case MyEnum.C:
  // ...
}
```

I think we should instead call `parse_expression` and `visit_expression` separately (instead of `parse_and_visit_expression`), then use the span of the expression.

> Is using a `hash_set<string8_view>` a good approach for this issue?

I think so, assuming the string8_view refers to the span of the entire expression. That would work with string literals, number literals, and constant references.

Your idea won't work if the case expressions are spelled slightly differently. But because exact-copy-paste errors are more common than things like `case "x": case "\x78":`, I think comparing string spans is the way to go.

strager:
Did you want to be assigned this task? It looks like you're half way to finishing it already. =]

PR #866, opened 2022-10-12
Warn on duplicate cases in switch statements

Rebraws:
Now quick-lint-js gives a warning when a switch statement has duplicated cases.

closes  https://github.com/quick-lint/quick-lint-js/issues/744


src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -464,6 +464,13 @@
       MESSAGE(QLJS_TRANSLATABLE("missing property name between '.' and '.'"),   \
               dots))                                                            \
                                                                                 \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_duplicated_cases_in_switch_statement, "E0347",                       \
+      diagnostic_severity::warning,                                             \
+      { source_code_span duplicated_switch_case; },                             \
+      MESSAGE(QLJS_TRANSLATABLE("duplicated case clause in switch statement"),  \
+              duplicated_switch_case))                                          \
```
strager, line 472: Important: We should point to the original `case` too:

> \<location of second case>: duplicated case clause in switch statement
> \<location of first case>: this case will run instead

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1851,10 +1853,19 @@ void parser::parse_and_visit_switch(parse_visitor_base &v) {
...
                    .colon_type_annotation = allow_type_annotations::never,
                });
+
+        source_code_span expression_case_span = ast->span();
+        if (repeated_cases.contains(expression_case_span.string_view())) {
+          this->diag_reporter_->report(
+              diag_duplicated_cases_in_switch_statement{
+                  .duplicated_switch_case = expression_case_span});
+        }
+        repeated_cases.insert(expression_case_span.string_view());
```
strager, line 1867: Perf: We shouldn't do two lookups. Instead, use `insert` and see if it failed:
```suggestion
        auto [inserted, _it] = repeated_cases.insert(expression_case_span.string_view());
        if (!inserted) {
          this->diag_reporter_->report(
              diag_duplicated_cases_in_switch_statement{
                  .duplicated_switch_case = expression_case_span});
        }
```

test/test-parse-statement.cpp
```diff
@@ -949,6 +949,42 @@ TEST_F(test_parse_statement, switch_case_without_expression) {
...
+  {
+    test_parser p(
+        u8"switch (cond) {case MyEnum.A: break; case MyEnum.A: break;}"_sv,
+        capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.visits, ElementsAre("visit_variable_use",       // cond
+                                      "visit_enter_block_scope",  //
+                                      "visit_variable_use",       // MyEnum.A
+                                      "visit_variable_use",       // MyEnum.A
+                                      "visit_exit_block_scope"));
```
strager, line 978: Nit: No need to assert visits, since we're just checking for duplicated cases.
```suggestion
    EXPECT_THAT(p.errors,
                ElementsAre(DIAG_TYPE_OFFSETS(
                    p.code, diag_duplicated_cases_in_switch_statement,
                    duplicated_switch_case,
                    strlen(u8"switch (cond) {case x: case y: case "), u8"y")));
  }
  {
    test_parser p(
        u8"switch (cond) {case MyEnum.A: break; case MyEnum.A: break;}"_sv,
        capture_diags);
    p.parse_and_visit_statement();
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1834,6 +1835,7 @@ void parser::parse_and_visit_switch(parse_visitor_base &v) {
 
   bool keep_going = true;
   bool is_before_first_switch_case = true;
+  hash_set<string8_view> repeated_cases;
```
strager, line 1838: Nit: I think a better name is `cases` or `case_expressions`. `repeated_cases` sounds like it contains *only* the duplicates, but it actually contains *every* `case`.

strager (approved):
Looks nice. Much simpler than I thought it would be.

Rebraws:
I think I've addressed all the comments. I had to replace `hash_set<string8_view>` with `hash_map<string8_view, source_code_span>` to also keep track of all the `source_code_span` to point to the original `case` when giving the warning.

Please let me know if anything else needs to be changed.

strager:
>  I had to replace `hash_set<string8_view>` with `hash_map<string8_view, source_code_span>`

I think this is unnecessary. source_code_span can be converted into string8_view, and string8_view can be converted back to source_code_span. Just `hash_set<string8_view>` should be fine.

test/test-parse-statement.cpp
```diff
@@ -949,6 +949,31 @@ TEST_F(test_parse_statement, switch_case_without_expression) {
...
+TEST_F(test_parse_statement, switch_case_with_duplicated_cases) {
+  {
+    test_parser p(u8"switch (cond) {case x: case y: case y:}"_sv,
+                  capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors,
+                ElementsAre(DIAG_TYPE_OFFSETS(
+                    p.code, diag_duplicated_cases_in_switch_statement,
+                    duplicated_switch_case,
+                    strlen(u8"switch (cond) {case x: case y: case "), u8"y")));
```
strager, line 961: Please test `first_switch_case` too.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1851,10 +1853,21 @@ void parser::parse_and_visit_switch(parse_visitor_base &v) {
...
                    .colon_type_annotation = allow_type_annotations::never,
                });
+
+        source_code_span expression_case_span = ast->span();
+        auto [_it, inserted] = cases.insert(expression_case_span.string_view());
+        if (!inserted) {
+          this->diag_reporter_->report(
+              diag_duplicated_cases_in_switch_statement{
+                  .first_switch_case =
+                      source_code_span(_it->begin(), _it->end()),
```
strager, line 1867: `begin` and `end` return iterators, but `source_code_span` expects pointers. This is causing build failures with MSVC. You should use `_it->data()` and `_it->data() + _it->size()` instead.

Rebraws: Thank you, I didn't realized that the constructor expected pointers, now I was just looking to the `source_code_span` class and saw that there's a `source_code_span::unit` member function, since we are using this span only to point to the original `case` do you think it would be better to use that function like?
`.first_switch_case = source_code_span::unit(_it->data())`

Or should I just use the constructor

strager: No, you don't want to use `unit`. `unit` is for diagnostics which point between two characters. Use the two-pointer constructor.

Rebraws:
Fixed the CI error and added the tests for `first_switch_case` let me know if there's anything else to improve :)

test/test-parse-statement.cpp
```diff
@@ -949,6 +949,33 @@ TEST_F(test_parse_statement, switch_case_without_expression) {
...
+TEST_F(test_parse_statement, switch_case_with_duplicated_cases) {
+  {
+    test_parser p(u8"switch (cond) {case x: case y: case y:}"_sv,
+                  capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors,
+                ElementsAre(DIAG_TYPE_2_OFFSETS(
+                    p.code, diag_duplicated_cases_in_switch_statement,
+                    first_switch_case, strlen(u8"switch (cond) {case x: case "),
+                    u8"y", duplicated_switch_case,
```
strager, line 961: Nit: Add a blank comment to improve code formatting:
```suggestion
                    first_switch_case, strlen(u8"switch (cond) {case x: case "),
                    u8"y", //
                    duplicated_switch_case,
```

strager (approved):

test/test-parse-statement.cpp
Rebraws: Fixed it  in commit https://github.com/quick-lint/quick-lint-js/pull/866/commits/0172714a51675810905d3cd0f84e72daef5d1078 and I also did a rebase to fix the merge conflicts

strager:
Beautiful!

PR #868, opened 2022-10-15
feature: Add Brazilian Portuguese translation

Guilherme-Vasconcelos:


strager:
> error: string literal of length 73968 exceeds maximum length 65536 that C++ compilers are required to support [-Werror,-Woverlength-strings]

Welp, I guess we need to disable this warning. 😆 Please remove it in `.github/workflows/build-and-test.yml`.

po/pt_BR.po
```diff
@@ -0,0 +1,1657 @@
+# Code generated by tools/update-translator-sources. DO NOT EDIT.
+# source: (various)
```
strager, line 2: Must fix: This comment shouldn't exist. (I should update the instructions to remove this message...)

strager (changes requested):

Guilherme-Vasconcelos:
> error: string literal of length 73968 exceeds maximum length 65536 that C++ compilers are required to support [-Werror,-Woverlength-strings]
> 
> Welp, I guess we need to disable this warning. :laughing: Please remove it in `.github/workflows/build-and-test.yml`.

Fixed! By the way, do you know why GCC wasn't giving the same warning? Looks like both compilers had that flag, but only clang was failing

strager:
> do you know why GCC wasn't giving the same warning? Looks like both compilers had that flag, but only clang was failing

I experimented, and it seems that GCC only warns in C code, not in C++ code.

strager (approved):

PR #870, opened 2022-10-17
842 better diagnostics for async functions

AlekLefebvre:


docs/errors/E0326.md
```diff
@@ -0,0 +1,20 @@
+# E0326: 'async export' is not allowed; write 'export async' instead
+
+Class methods can have modifiers such as `export`, `static`, `async`, and `*`. It is a
```
strager, line 3: Class methods can be marked `export`...? I'm confused by this wording. It seems wrong and unrelated to `async export function f() {}`.

docs/errors/E0326.md
```diff
@@ -0,0 +1,20 @@
...
+
+To fix this error, replace `async export` with `export async`:
+
+```javascript
+export async function f() {
+  return 0;
+}
+```
+
+Introduced in quick-lint-js version 3.0.0.
```
strager, line 20: Must fix: Remove this line. I don't know what the next version number will be yet.

docs/errors/E0327.md
```diff
@@ -0,0 +1,21 @@
+# E0327: 'function async f()' is not allowed; write 'async function f()' instead
+
+Class methods can have modifiers such as `export`, `static`, `async`, and `*`. It is a
+syntax error to write `function async f()` in a function declaration because async will
+not be recognized as a keyword but as the function name:
```
strager, line 5: > async will not be recognized as a keyword but as the function name

Well, this isn't true, because it's a syntax error. I'd remove this part of the sentence.

I suggest this wording instead for the opening paragraph:

> Functions can be marked with the `async` keyword. It is a syntax error to write the `async` keyword after the `function` keyword:

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -273,7 +273,12 @@ bool parser::parse_and_visit_statement(
       this->skip();
       this->check_body_after_label();
       goto parse_statement;
-
+      // "async export f()" is not valid. It should be "export async f()"
+    case token_type::kw_export: {
+      this->diag_reporter_->report(
+          diag_async_export_method{async_token.span()});
+      break;
```
strager, line 280: Shouldn't we parse the exported function (line 231)? If we don't, then the following will also (incorrectly) report that we're using `await` in a not-async function:
```javascript
async export function f() {
  await myPromise;  // await should be an operator here.
}
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -273,7 +273,12 @@ bool parser::parse_and_visit_statement(
       this->skip();
       this->check_body_after_label();
       goto parse_statement;
-
+      // "async export f()" is not valid. It should be "export async f()"
```
strager, line 276: Nit: Shouldn't this comment say 'async export **function** f' and 'export async **function** f'?

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1651,6 +1656,19 @@ parser::parse_and_visit_function_parameters(
...
+    if (name->string_view() ==
+        u8"async"_sv) {  // In this context async is of type identifier and not
+                         // kw_async so we need to look at the name.
+      this->diag_reporter_->report(
+          diag_function_async_method{this->peek().span()});
+      this->skip();
+      return this->parse_and_visit_function_parameters(v, name);
+    }
+    [[fallthrough]];  // Should still call QLJS_PARSER_UNIMPLEMENTED if the
+                      // identifier is not "async"
```
strager, line 1670: Minor: Instead of falling through here, we should copy-paste the code:
```suggestion
    QLJS_PARSER_UNIMPLEMENTED();
    return function_parameter_parse_result::parsed_parameters;
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1651,6 +1656,19 @@ parser::parse_and_visit_function_parameters(
...
 
+    // function async f() {}  // Invalid. Should be async function f() {}
+  case token_type::identifier:
+    if (name->string_view() ==
+        u8"async"_sv) {  // In this context async is of type identifier and not
+                         // kw_async so we need to look at the name.
+      this->diag_reporter_->report(
+          diag_function_async_method{this->peek().span()});
+      this->skip();
+      return this->parse_and_visit_function_parameters(v, name);
```
strager, line 1667: Same comment as with `async export function`: we're not parsing the function body as if the function was `async`. But it's not easy here to fix that, so I think a TODO comment is fine.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1651,6 +1656,19 @@ parser::parse_and_visit_function_parameters(
     });
     return function_parameter_parse_result::missing_parameters_ignore_body;
 
+    // function async f() {}  // Invalid. Should be async function f() {}
+  case token_type::identifier:
+    if (name->string_view() ==
+        u8"async"_sv) {  // In this context async is of type identifier and not
+                         // kw_async so we need to look at the name.
```
strager, line 1663: Nit: This comment is misleading. The real issue is that we don't have the `token` object. I think we don't need a comment here, or if there is one, it should be something like:

> TODO: Make parse_and_visit_function_parameters accept a token instead of a source_code_span so we can compare the token type instead of strings.

src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -211,6 +211,21 @@
               "'async static' is not allowed; write 'static async' instead"),   \
           async_static))                                                        \
                                                                                 \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_async_export_method, "E0326", diagnostic_severity::error,            \
```
strager, line 215: This should be called `diag_async_export_function`, because it's for a function, not a class method.

src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -211,6 +211,21 @@
...
+      { source_code_span async_export; },                                       \
+      MESSAGE(                                                                  \
+          QLJS_TRANSLATABLE(                                                    \
+              "'async export' is not allowed; write 'export async' instead"),   \
+          async_export))                                                        \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_function_async_method, "E0327", diagnostic_severity::error,          \
```
strager, line 223: This should be called `diag_function_async_function`, because it's for a function, not a class method.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -273,7 +273,12 @@ bool parser::parse_and_visit_statement(
       this->skip();
       this->check_body_after_label();
       goto parse_statement;
-
+      // "async export f()" is not valid. It should be "export async f()"
+    case token_type::kw_export: {
+      this->diag_reporter_->report(
+          diag_async_export_method{async_token.span()});
```
strager, line 279: Must fix: Use designated initializers (i.e. named arguments). Once you do, you'll realize that `async_token.span()` is wrong; it only covers the `async` keyword, not both `async` and `export`. This should have been tested; see my comment in the tests.
```suggestion
          diag_async_export_method{.async_export = async_token.span()});
```
I think this code should look something like the following:
```suggestion
          diag_async_export_method{
            .async_export = source_code_span(async_token.begin(), this->peek().end),
          });
```

test/test-parse-function.cpp
```diff
@@ -204,6 +204,40 @@ TEST_F(test_parse_function, async_function_statement) {
...
+    ASSERT_EQ(p.variable_declarations.size(), 1);
+    EXPECT_EQ(p.variable_declarations[0].name, u8"f");
+  }
+}
+
+TEST_F(test_parse_function, async_keyword_order_diagnostic2) {
+  {
+    test_parser p(u8"async export function f() {};"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, ElementsAre(DIAG_TYPE(diag_async_export_method)));
```
strager, line 220: Must fix: Please assert the span of `async_export`. This testing would have caught a bug.

test/test-parse-function.cpp
```diff
@@ -204,6 +204,40 @@ TEST_F(test_parse_function, async_function_statement) {
...
+  }
+}
+
+TEST_F(test_parse_function, async_keyword_order_diagnostic4) {
+  {
+    test_parser p(u8"function async f() {};"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, ElementsAre(DIAG_TYPE(diag_function_async_method)));
+  }
+}
```
strager, line 239: Nit: Please merge these test cases together:
```suggestion
TEST_F(test_parse_function, export_async_function_keyword_order) {
  {
    test_parser p(u8"export async function f() {};"_sv);
    p.parse_and_visit_statement();
    ASSERT_EQ(p.variable_declarations.size(), 1);
    EXPECT_EQ(p.variable_declarations[0].name, u8"f");
  }

  {
    test_parser p(u8"async export function f() {};"_sv, capture_diags);
    p.parse_and_visit_statement();
    EXPECT_THAT(p.errors, ElementsAre(DIAG_TYPE(diag_async_export_method)));
  }

  {
    test_parser p(u8"async function f() {};"_sv);
    p.parse_and_visit_statement();
    ASSERT_EQ(p.variable_declarations.size(), 1);
    EXPECT_EQ(p.variable_declarations[0].name, u8"f");
  }

  {
    test_parser p(u8"function async f() {};"_sv, capture_diags);
    p.parse_and_visit_statement();
    EXPECT_THAT(p.errors, ElementsAre(DIAG_TYPE(diag_function_async_method)));
  }
}
```

test/test-parse-function.cpp
```diff
@@ -204,6 +204,40 @@ TEST_F(test_parse_function, async_function_statement) {
   }
 }
 
+TEST_F(test_parse_function, async_keyword_order_diagnostic1) {
+  {
+    test_parser p(u8"export async function f() {};"_sv);
+    p.parse_and_visit_statement();
+    ASSERT_EQ(p.variable_declarations.size(), 1);
+    EXPECT_EQ(p.variable_declarations[0].name, u8"f");
```
strager, line 212: Nit:
```suggestion
    EXPECT_THAT(p.variable_declarations, ElementsAre(function_decl(u8"f")));
```

test/test-parse-function.cpp
```diff
@@ -204,6 +204,40 @@ TEST_F(test_parse_function, async_function_statement) {
...
+  }
+}
+
+TEST_F(test_parse_function, async_keyword_order_diagnostic2) {
+  {
+    test_parser p(u8"async export function f() {};"_sv, capture_diags);
```
strager, line 218: Must fix: To ensure that the function was interpreted as `async`, we should write `await myPromise` in the body and check `p.variable_uses`. See the `await_in_async_function` in this file for an example.

strager (changes requested):
This looks like a good patch, but it needs some fixes before it's perfect.

docs/errors/E0327.md
vegerot: I get you want to tell users the fun fact that `async` can be a function name, and that the error is really two identifiers back-to-back; but I think teaching users this will just further confuse them.  sTRAGER's suggestion is better.  My suggestion is best

> Functions can be marked using the `async` keyword like `async function f() {`.  It is a syntax error to write the `async` keyword after the `function` keyword:

AlekLefebvre: Thank you both, much clearer like that.

src/quick-lint-js/fe/parse-statement.cpp
AlekLefebvre: Done. Why is it better to do it that way instead of falling through?

src/quick-lint-js/fe/parse-statement.cpp
AlekLefebvre: I'd like to work on that issue next if possible.

src/quick-lint-js/fe/parse-statement.cpp
strager: When we end up writing a good diagnostic, the code will almost certainly be different between the two. The `case` code and the `default` code do not need to be kept in sync. They can evolve independently.

A fallthrough also makes it harder to add more `case`s later.

AlekLefebvre:
I added a fix for #872 to this PR

strager (approved):

strager:
Thanks for the improvements!

PR #876, opened 2022-10-23
TypeScript: warn when constructor or constructor type has 'this' parameter

### LINKED ISSUE #853: 10$: error if constructor has TypeScript 'this' parameter
We should diagnose the error in the following lines:
```typescript
type F = new (this: string) => string;
class C {
  constructor(this: string) {}
}
```

Guilherme-Vasconcelos:
Closes #853

### DIFF
--- a/src/quick-lint-js/fe/diagnostic-types.h
+++ b/src/quick-lint-js/fe/diagnostic-types.h
@@ -2582,6 +2582,14 @@
       MESSAGE(QLJS_TRANSLATABLE("using '{0}' against a regular "                \
                                 "expression literal always returns '{1}'"),     \
               equals_operator, comparison_result))                              \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_this_parameter_not_allowed_in_typescript_constructor, "E0348",       \
+      diagnostic_severity::error, { source_code_span this_keyword; },           \
+      MESSAGE(                                                                  \
+          QLJS_TRANSLATABLE("in TypeScript, 'this' keyword is not allowed "     \
+                            "in constructors"),                                 \
+          this_keyword))                                                        \
   /* END */
 
 // QLJS_X_RESERVED_DIAG_TYPES lists reserved error codes. These codes were used
--- a/src/quick-lint-js/fe/parse-class.cpp
+++ b/src/quick-lint-js/fe/parse-class.cpp
@@ -403,9 +403,17 @@ void parser::parse_and_visit_class_or_interface_member(
       QLJS_CASE_CONTEXTUAL_KEYWORD:
       case token_type::identifier:
       case token_type::reserved_keyword_with_escape_sequence: {
-        identifier property_name = p->peek().identifier_name();
-        p->skip();
-        parse_and_visit_field_or_method(property_name);
+        auto parse_and_visit = [this](auto p) {
+          identifier property_name = p->peek().identifier_name();
+          p->skip();
+          parse_and_visit_field_or_method(property_name);
+        };
+        if (p->peek().type == token_type::kw_constructor) {
+          constructor_guard g = p->enter_constructor();
+          parse_and_visit(p);
+        } else {
+          parse_and_visit(p);
+        }
         break;
       }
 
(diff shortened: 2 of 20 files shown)


src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -2582,6 +2582,14 @@
       MESSAGE(QLJS_TRANSLATABLE("using '{0}' against a regular "                \
                                 "expression literal always returns '{1}'"),     \
               equals_operator, comparison_result))                              \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_this_parameter_not_allowed_in_typescript_constructor, "E0348",       \
+      diagnostic_severity::error, { source_code_span this_keyword; },           \
+      MESSAGE(                                                                  \
+          QLJS_TRANSLATABLE("in TypeScript, 'this' keyword is not allowed "     \
+                            "in constructors"),                                 \
```
strager, line 2591: Nit: I think saying "in TypeScript" is confusing. Mentioning TypeScript implies that 'this' is allowed in JavaScript, but it is not allowed in JavaScript either. I think a better message is: "'this' keyword is not allowed in constructors"

Nit: Saying "'this' keyword is not allowed in constructors" is misleading. 'this' *is* allowed in constructors, but it is not allowed as a constructor *parameter*. I think a better message is: "'this' parameters are now allowed in constructors"

src/quick-lint-js/fe/parse-class.cpp
```diff
@@ -403,9 +403,17 @@ void parser::parse_and_visit_class_or_interface_member(
...
+          identifier property_name = p->peek().identifier_name();
+          p->skip();
+          parse_and_visit_field_or_method(property_name);
+        };
+        if (p->peek().type == token_type::kw_constructor) {
+          constructor_guard g = p->enter_constructor();
+          parse_and_visit(p);
+        } else {
+          parse_and_visit(p);
+        }
```
strager, line 416: Nit: Why is `p` a parameter? I think it's already accessible through `this`:
```suggestion
        auto parse_and_visit = [this]() {
          identifier property_name = p->peek().identifier_name();
          p->skip();
          parse_and_visit_field_or_method(property_name);
        };
        if (p->peek().type == token_type::kw_constructor) {
          constructor_guard g = p->enter_constructor();
          parse_and_visit();
        } else {
          parse_and_visit();
        }
```

src/quick-lint-js/fe/parse-class.cpp
```diff
@@ -1183,7 +1191,10 @@ void parser::parse_and_visit_typescript_interface(
   case token_type::kw_assert:
   case token_type::kw_asserts:
   case token_type::kw_async:
-  case token_type::kw_constructor:
+  case token_type::kw_constructor: {
+    constructor_guard g = this->enter_constructor();
+    [[fallthrough]];
```
strager, line 1196: Must fix: This cannot work. The `constructor_guard` is destructed before the fallthrough, and this code also runs if the token is `kw_async` (for example), not just for `kw_constructor`. I think you didn't test this code path.

strager (changes requested):
I'm sorry I took so long to look at your PR.

I like the `is_constructor_` idea. I didn't think it'd be this simple (relatively). However...

Must fix: You don't clear the `is_constructor_` flag when entering a nested function. That means the following code will produce false positive diagnostics because `is_constructor_` is true when it should be false:
```typescript
class C {
  constructor() {
    let o = {
      method(this) {}  // is_constructor_ is true here, but should be false
    };

    function f(this) {}  // is_constructor_ is true here, but should be false
  }
}

type T = new (func: (this: any /* is_constructor_ is true here, but should be false */) => any) => any;
```

src/quick-lint-js/fe/parse-class.cpp
Guilherme-Vasconcelos: Thanks for the tip, I hadn't noticed that

Guilherme-Vasconcelos:
I'll convert this PR to a draft until I finish the requested changes

PR #902, opened 2023-01-07
feat(lint): ignore QuickJS scriptArgs pre defined global

wgrr:
currently, this is the only global variable non standard defined by quickjs:
https://bellard.org/quickjs/quickjs.html#Global-objects

### DIFF
--- a/src/quick-lint-js/fe/global-variables.cpp
+++ b/src/quick-lint-js/fe/global-variables.cpp
@@ -232,6 +232,11 @@ constexpr const char8 global_variables_node_js_es[] =
     u8"setTimeout\0"
     u8"unescape\0";
 
+constexpr const char8 global_variables_quickjs[] =
+    u8"scriptArgs\0"
+    u8"console\0"
+    u8"print\0";
+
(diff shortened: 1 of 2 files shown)


wgrr:
maybe there is something to mention about this change in docs/config.adoc, i'm not sure

strager (approved):

src/quick-lint-js/fe/global-variables.cpp
```diff
@@ -232,6 +232,8 @@ constexpr const char8 global_variables_node_js_es[] =
     u8"setTimeout\0"
     u8"unescape\0";
 
+constexpr const char8 global_variables_quickjs[] = u8"scriptArgs\0";
```
strager, line 235: I think this should include `console` and `print` too. Otherwise, if someone configures quick-lint-js for ecmascript+quickjs, `console.log` would warn.

strager (approved):

wgrr:
this change was tested with a file containg:
```js
scriptArgs()
console.log()
print()
```
not errors or warnings as expected

PR #911, opened 2023-01-18
feature : added null and undefined in E0109

### LINKED ISSUE #836: 7$: warn on 'x == a || null' and 'x == a || undefined'

HarshitAghera:
- null was working already, added check for undefined
- added tests for null and undefined

closes #836


strager (approved):
Good patch. Thanks!

PR #914, opened 2023-01-21
feat(cli): don't print escape sequences in dumb terminals

wgrr:


src/quick-lint-js/cli/main.cpp
```diff
@@ -59,7 +59,8 @@ bool stderr_supports_terminal_escapes() {
 #if defined(_WIN32)
   return false;
 #else
-  return isatty(STDERR_FILENO);
+  const char *term = std::getenv("TERM");
+  return isatty(STDERR_FILENO) && term && term != "dumb"sv;
```
strager, line 63: Your code currently treats unset `TERM` differently than empty `TERM`. (Unset returns false; empty returns true.) Is that intentional?

test/test-cli.py
```diff
@@ -326,6 +327,26 @@ def test_language_typescript_warns_about_undeclared_variables_despite_eval(
...
+                os.execve(qljs, [qljs, str(test_file)], {"TERM": "dumb"})
+            else:
+                (_, rc) = os.waitpid(pid, 0)
+                stderr = os.read(fd, 4096)
+                returncode = 0
+                if os.WIFEXITED(rc):
+                    returncode = os.WEXITSTATUS(rc)
+                self.assertEqual(returncode, 1)
+                self.assertNotIn(
+                    b"[\x1b]8;;https://quick-lint-js.com/errors/E0017/\x1b\E0017\x1b]8;;\x1b\]",
```
strager, line 347: I think it's safer to assert that just `\x1b` is not present. We might change the exact formatting or the URL, and doing so would cause this test to always pass.

strager (approved):
Looks good to me. I'll need to fix Windows, etc.

src/quick-lint-js/cli/main.cpp
wgrr: yes, that's what the gcc reference does, if TERM is set, and is not equal to "dumb", then it assumes to be able to emit escape sequences (at least on posix systems), it could be more accurate by looking up terminfo databases, but i feel that's too much trouble for little gain.

strager:
I fixed the test on Windows (by disabling it), then landed the patch as commit 709304f3ef8e1f1147cdaf9278f84ce7927f9f3e.

Issue #735, opened 2022-06-01
Stack overflow issues with MSVC


strager:
Our stack limit is 150. Adding one `source_code_span` variable in `parser::parse_and_visit_class_or_interface_member` was the final straw and caused one of our stack depth tests to fail. 😕 

I'm going to slower the stack limit for now, but we should find a better solution.

strager:
I did some investigation with a x64 Debug build.

* `parser::parse_and_visit_statement` consumes about 2 KiB of stack space.
* `parser::parse_and_visit_class_or_interface_member ` consumes about 1.5 KiB of stack space per call.

* It looks like there is significant padding between most stack variables. For example, in `parser::parse_and_visit_statement`:
  * [1012,1012] `bool`
  * [1013,1013] `bool`
  * [1014,1031] *18 bytes of padding*
  * [1032,1047] `parser::bool_guard`
  * [1048,1079] *32 bytes of padding*
  * [1080,1095] `parser::bool_guard`
  * [1096,1127] *32 bytes of padding*
  * [1128,1143] `parser::bool_guard`
  * [1144,1155] *12 bytes of padding*
  * [1156,1156] `bool`
  * [1157,1157] `bool`
* Variables with nonoverlapping lifetimes are allocated on the stack separately. This is a big problem because we have a lot of `switch` cases.

strager:
I think I addressed the MSVC issues by refactoring parse_and_visit_class to use a helper class instead of a bunch of lambda functions. I'm testing with CI to see: 6eb7a7572908bd6d5830e7719aebb096f89e2a5b

strager:
MSVC stackoverflowed with my patch. =\

strager:
I think the main issue is that, in unoptimized builds, MSVC allocates all locals on the stack irrespective of liveness. So the following code allocates two booleans on the stack:

```c++
bool foo(int p) {
  if (p > 0) {
    bool ok = false;
    return ok;
  } else {
    bool ok = true;  // Note: does not share stack space with 'ok' above.
    return ok;
  }
}
```

I don't know why there's all that padding between certain variables. I tried disabling some security switches (like `/RTC`) but they didn't affect stack allocation significantly.

[...]

PR #924, opened 2023-01-28
Fix crash in quick-lint-js-benchmark-parse

save-buffer:
There was one extra scope going in which was causing the depth to exceed 130. Removing just one of the curly braces causes this limit to not be reached.

### DIFF
--- a/benchmark/benchmark-parse.cpp
+++ b/benchmark/benchmark-parse.cpp
@@ -67,7 +67,7 @@ function f() {
   ()=>{{{{{{{await/
   ()=>{{{{{{{await/
   ()=>{{{{{{{await/
-  ()=>{{{{{{{await/
+  ()=>{{{{{{ await/ 
 }
 )"_sv);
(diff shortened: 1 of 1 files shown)


save-buffer:

save-buffer:
recheck

strager:
Thanks for looking into this.

The regression in this benchmark's stability was introduced in commit c2e9c9bcb777304fc7d53dd550acf9d9ea300261. Since then, I think we fixed the code which was causing MSVC to use excessive stack. I'll investigate bumping the stack limit.

strager:
Increasing the stack limit isn't an option right now (#735), so your patch is good.

strager:
FYI: I reverted your PR in commit 59c3ea4c7245ff50e836296ff5c075b1cb4fade6, after I implemented a (somewhat) proper fix.

Ideally, we'd keep benchmarks the same over time, so we can track performance of the production code over time (rather than the performance of the benchmarks). So in the interest in having good numbers, I rolled back the benchmark to its original state.

save-buffer:
Sounds good, I agree that’s the way to go

PR #928, opened 2023-01-29
added suggestion for colon in type decleration

cmstoddard:


strager:
Landed as commit 0cb7ef5dc590ec0151639ede91e22cff08972af1.

PR #940, opened 2023-02-27
Add warning diagnostic for comma operator usage in indexing operation

### LINKED ISSUE #939: 10$: warn on comma operator in indexing operation: array[i, j]

yunusey:
This commit adds a new diagnostic to point out misleading use of comma operators in indexing operation (e.g., let b = a[i, j]).
Closes #939

### DIFF
--- a/docs/errors/E0450.md
+++ b/docs/errors/E0450.md
@@ -1,0 +1,26 @@
+# E0450: misleading use of ',' operator in index
+
+Using comma operator within an indexing operation is misleading, as it
+can be confused with a new array definition, or with access to multiple elements
+of the container.
+
+For example:
+
+```javascript
+let matrix = [
+  [1, 2, 3],
+  [4, 5, 6],
+  [7, 8, 9],
+];
+let topLeft = matrix[0, 0];
+```
+Here, the value of `topLeft` will be equal to `matrix[0] = [1, 2, 3]`.
+If the intention is to find the value at the top-left corner of `matrix`, the correct syntax would be:
+```javascript
+let matrix = [
+  [1, 2, 3],
+  [4, 5, 6],
+  [7, 8, 9],
+];
+let topLeft = matrix[0][0]; // topLeft = 1
+```
--- a/src/quick-lint-js/fe/diagnostic-types.h
+++ b/src/quick-lint-js/fe/diagnostic-types.h
@@ -559,6 +559,17 @@
           comma))                                                               \
                                                                                 \
   QLJS_DIAG_TYPE(                                                               \
+      diag_misleading_comma_operator_in_index_operation, "E0450",               \
+      diagnostic_severity::warning,                                             \
+      {                                                                         \
+        source_code_span comma;                                                 \
+        source_code_span left_square;                                           \
+      },                                                                        \
+      MESSAGE(QLJS_TRANSLATABLE("misleading use of ',' operator in index"),     \
+              comma)                                                            \
+          MESSAGE(QLJS_TRANSLATABLE("index starts here"), left_square))         \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
       diag_expected_as_before_imported_namespace_alias, "E0126",                \
       diagnostic_severity::error,                                               \
       {                                                                         \
--- a/test/test-parse-warning.cpp
+++ b/test/test-parse-warning.cpp
@@ -343,6 +343,54 @@ TEST_F(test_parse_warning,
 }
 
 TEST_F(test_parse_warning,
+       warn_on_comma_between_member_array_subscript_operators) {
+  {
+    test_parser p(u8"a[1, 2, 3]"_sv, capture_diags);
+    p.parse_and_visit_expression();
+    EXPECT_THAT(
+        p.errors,
+        ElementsAreArray({
+            DIAG_TYPE_2_OFFSETS(
+                p.code, diag_misleading_comma_operator_in_index_operation,
+                comma, strlen(u8"a[1, 2"), u8","_sv, left_square, strlen(u8"a"),
+                u8"["_sv),
+        }));
+  }
+
+  {
+    test_parser p(u8"a[pow(1, 2), 2]"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(
+        p.errors,
+        ElementsAreArray({
+            DIAG_TYPE_2_OFFSETS(
+                p.code, diag_misleading_comma_operator_in_index_operation,
+                comma, strlen(u8"a[pow(1, 2)"), u8","_sv, left_square,
+                strlen(u8"a"), u8"["_sv),
+        }));
+  }
+
+  {
+    test_parser p(u8"a[b[1967, 1975]]"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(
+        p.errors,
+        ElementsAreArray({
+            DIAG_TYPE_2_OFFSETS(
+                p.code, diag_misleading_comma_operator_in_index_operation,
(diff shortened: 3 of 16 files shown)


yunusey:

yunusey:
Addresses the issue #939.

yunusey:
Closes #939

docs/errors/E0450.md
```diff
@@ -0,0 +1,20 @@
...
+let b = a[0, 1, 2];
+```
+Here, the value of `b` will be equal to `a[1] = 5`.
+If the intention is to create a new array `b` with the first three elements of array `a` the correct syntax would be:
+```javascript
+let a = [1, 2, 3, 4];
+let b = a[a[0], a[1], a[2]];
+```
+
+Introduced in quick-lint-js version 2.11.0.
```
strager, line 20: Must fix: This is wrong. Version 2.11.0 has already been released.

I'd just delete this paragraph for now:
```suggestion
```

src/quick-lint-js/fe/diagnostic-types.h
```diff
@@ -558,6 +558,12 @@
           QLJS_TRANSLATABLE("extra ',' is not allowed between enum members"),   \
           comma))                                                               \
                                                                                 \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_misleading_comma_operator_in_index_operation, "E0450",               \
+      diagnostic_severity::warning, { source_code_span comma; },                \
+      MESSAGE(QLJS_TRANSLATABLE("misleading use of ',' operator in in index"),  \
```
strager, line 564: Fix typo:
```suggestion
      MESSAGE(QLJS_TRANSLATABLE("misleading use of ',' operator in index"),  \
```

docs/errors/E0450.md
```diff
@@ -0,0 +1,20 @@
+# E0450: misleading use of comma operator in indexing operation
```
strager, line 1: This doc title should match the diagnostic's message ("misleading use of ',' operator in in index"). Either change this title or change the diagnostic in diagnostic-types.h.

test/test-parse-warning.cpp
```diff
@@ -342,6 +342,86 @@ TEST_F(test_parse_warning,
...
+    test_parser p(u8"a = [1, 2, 3]"_sv, capture_diags);
+    p.parse_and_visit_expression();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
+  {
+    test_parser p(u8"for (let i = 0, j = 1; ; ++i, ++j) {}"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
+}
```
strager, line 423: Nitpick: I think this testing is excessive for how simple the implementation is. I'd deduplicate some tests, such as the `a[b[1967, 1975]]` and `a[0][-1, 1]` and `let b = a[-1, 1]` tests which cover the same thing.

src/quick-lint-js/fe/diagnostic-types.h
strager, line 564: Minor: I think we should also point to the `[` in a secondary message. This could help people understand what 'index' means exactly.

    foo[bar, baz]
    //     ^ misleading use of ',' operator in index
    // ^ index starts here

strager (changes requested):
Overall, this patch looks good. 👍

src/quick-lint-js/fe/diagnostic-types.h
yunusey: I have added an additional diagnostic note as in the example above, and also addressed the other issues you pointed out. Thanks!

strager (approved):

strager:
The 'check error docs' CI job detected some issues: https://github.com/quick-lint/quick-lint-js/actions/runs/4289586245/jobs/7477197134

The existing example in docs/errors/E0062.md is triggering for your code, but shouldn't. It looks like our error recovery is bad here:

https://github.com/quick-lint/quick-lint-js/blob/7863d029128d7562134171c7f54e152bf46b0d1c/src/quick-lint-js/fe/parse-statement.cpp#L1501-L1502

The problem is that `parse_expression_remainder` is called which parses `[` as an index operation. I think a good fix for this is to only call `parse_expression_remainder` if the current token is `(` (IIFE).

docs/errors/E0450.md
```diff
@@ -0,0 +1,18 @@
...
+
+```javascript
+let a = [1, 3, 5, 7, 9, 11];
+let b = a[0, 1, 2];
+```
+Here, the value of `b` will be equal to `a[1] = 5`.
+If the intention is to create a new array `b` with the first three elements of array `a` the correct syntax would be:
+```javascript
+let a = [1, 2, 3, 4];
+let b = a[a[0], a[1], a[2]];
```
strager, line 17: Must fix: This line triggers E0450. CI catches this mistake: https://github.com/quick-lint/quick-lint-js/actions/runs/4289586245/jobs/7477197134

> /__w/quick-lint-js/quick-lint-js/docs/errors/E0450.md: error: expected no error in code block #2 but found errors

I think you mean this?
```suggestion
let b = [a[0], a[1], a[2]];
```

Looking at it more closely, though, I think the comma syntax is more often used for multidimensional arrays, so I suggest you change your examples here:
```javascript
let matrix = [
  [1, 2, 3],
  [4, 5, 6],
  [7, 8, 9],
];
let topLeft = matrix[0, 0]; // should be: matrix[0][0]
```

strager (changes requested):

docs/errors/E0450.md
yunusey: Yes, I fixed that typo, and enhanced the error recovery around IIFE as you suggested.

strager (approved):

yunusey:
I just saw that the CI running on the documentation failed again due to having both E0062 and E0061 in the documentation of E0062. But with the change we just made to better handle IIFE, I think this is expected, and we should remove the second half of the first block in [```docs/errors/E0062.md```.](https://github.com/quick-lint/quick-lint-js/blob/master/docs/errors/E0062.md) It is practically identical to the test we have in [```docs/errors/E0061.md```](https://github.com/quick-lint/quick-lint-js/blob/master/docs/errors/E0061.md), so should produce the same error.

strager:
> But with the change we just made to better handle IIFE, I think this is expected, and we should remove the second half of the first block in [docs/errors/E0062.md.](https://github.com/quick-lint/quick-lint-js/blob/master/docs/errors/E0062.md) It is practically identical to the test we have in [docs/errors/E0061.md](https://github.com/quick-lint/quick-lint-js/blob/master/docs/errors/E0061.md), so should produce the same error.

I agree!

Can you do this (and the related fix) in a separate PR? If you don't want to deal with the Git-fu, that's fine; you can put the fixes in this PR and I can split them out when I merge.

Unrelated: I fixed the 'test Emacs plugin' and 'test Nix install' issues in master, so if the failures bother you, you can rebase or merge master to get those fixed.

yunusey:
> 
> I agree!
> 
> Can you do this (and the related fix) in a separate PR? If you don't want to deal with the Git-fu, that's fine; you can put the fixes in this PR and I can split them out when I merge.
> 
> Unrelated: I fixed the 'test Emacs plugin' and 'test Nix install' issues in master, so if the failures bother you, you can rebase or merge master to get those fixed.

Hi @strager, I created a separate PR (#942) for IIFE related changes, and squashed all of my other changes into a single commit here after merging master. Hope I have not broken the git history :)

strager (approved):

strager:
Landed as commit 35d94b9d6c65e6b645189ae22c918e1f6ac526db.

PR #945, opened 2023-03-03
Warn on comma operator in conditional statements

### LINKED ISSUE #585: 8$: Warn when comma operator used in 'if' or 'while' or 'for' condition
e.g. https://old.reddit.com/r/learnprogramming/comments/rh3ez8/continue_in_nested_while_loop_problem_javascript/

yunusey:
This commit adds a warning diagnostic to catch misleading comma operator usage in conditional statements of the if, while, for, and switch control blocks.

Closes #585

### DIFF
--- a/docs/errors/E0451.md
+++ b/docs/errors/E0451.md
@@ -1,0 +1,23 @@
+# E0451: misleading use of ',' operator in conditional statement
+
+Using comma operator within a conditional statement of a control block (if, for, while, switch) is misleading, because only the last expression will be used to evaluate the condition.
+
+For example:
+
+```javascript
+let N = 10;
+for (let i = 0, j = N - 1; i < j, j >= 0; ++i, --j) {
+  console.log(i, j);
+}
+```
+
+Here, the statement `i < j, j >= 0` will be evaluated as `j >= 0`, disregarding
+the first part of the statement `i < j`. In order to have both parts evaluated,
+a conditional operator should be used instead of the comma operator.
+
+```javascript
+let N = 10;
+for (let i = 0, j = N - 1; i < j && j >= 0; ++i, --j) {
+  console.log(i, j);
+}
+```
--- a/docs/errors/E0451.md
+++ b/docs/errors/E0451.md
@@ -1,6 +1,6 @@
 # E0451: misleading use of ',' operator in conditional statement
 
-Using comma operator within a conditional statement of a control block (if, for, while, switch) is misleading, because only the last expression will be used to evaluate the condition.
+Using comma operator within a conditional statement of a control block (`if`, `for`, `while`, `switch`) is misleading, because only the last expression will be used to evaluate the condition.
 
 For example:
 
@@ -12,8 +12,8 @@ for (let i = 0, j = N - 1; i < j, j >= 0; ++i, --j) {
 ```
 
 Here, the statement `i < j, j >= 0` will be evaluated as `j >= 0`, disregarding
-the first part of the statement `i < j`. In order to have both parts evaluated,
-a conditional operator should be used instead of the comma operator.
+the first part (`i < j`) of the statement. In order to have both parts evaluated,
+a logical operator should be used instead of the comma operator.
 
 ```javascript
 let N = 10;
--- a/src/quick-lint-js/fe/parse.cpp
+++ b/src/quick-lint-js/fe/parse.cpp
@@ -328,6 +328,23 @@ void parser::error_on_sketchy_condition(expression* ast) {
   }
 }
 
+void parser::warn_on_comma_operator_in_conditional_statement(expression* ast) {
+  if (ast->kind() != expression_kind::binary_operator) return;
+
+  auto is_comma = [](string8_view s) -> bool { return s == u8","_sv; };
+
+  auto* binary_operator = static_cast<expression::binary_operator*>(ast);
+  for (span_size i = binary_operator->child_count() - 2; i >= 0; i--) {
+    source_code_span op_span = binary_operator->operator_spans_[i];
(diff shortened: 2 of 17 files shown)


docs/errors/E0451.md
```diff
@@ -0,0 +1,23 @@
+# E0451: misleading use of ',' operator in conditional statement
+
+Using comma operator within a conditional statement of a control block (if, for, while, switch) is misleading, because only the last expression will be used to evaluate the condition.
```
strager, line 3: Nitpick: Surround code with \`:

```suggestion
Using comma operator within a conditional statement of a control block (`if`, `for`, `while`, `switch`) is misleading, because only the last expression will be used to evaluate the condition.
```

docs/errors/E0451.md
```diff
@@ -0,0 +1,23 @@
...
+
+```javascript
+let N = 10;
+for (let i = 0, j = N - 1; i < j, j >= 0; ++i, --j) {
+  console.log(i, j);
+}
+```
+
+Here, the statement `i < j, j >= 0` will be evaluated as `j >= 0`, disregarding
+the first part of the statement `i < j`. In order to have both parts evaluated,
```
strager, line 15: Nitpick: Technically, `i < j` is just an expression here, not a statement.
```suggestion
the first part of the expression `i < j`. In order to have both parts evaluated,
```

docs/errors/E0451.md
```diff
@@ -0,0 +1,23 @@
...
+a conditional operator should be used instead of the comma operator.
```
strager, line 16: Nitpick: I think the term 'conditional operator' usually applies to `? :`, not to logical operators like `&&` and `||`.
```suggestion
a logical operator should be used instead of the comma operator.
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -328,6 +328,23 @@ void parser::error_on_sketchy_condition(expression* ast) {
...
+  for (span_size i = binary_operator->child_count() - 2; i >= 0; i--) {
+    source_code_span op_span = binary_operator->operator_spans_[i];
+    if (is_comma(op_span.string_view())) {
+      this->diag_reporter_->report(
+          diag_misleading_comma_operator_in_conditional_statement{.comma =
+                                                                      op_span});
+      return;
+    }
+  }
+}
```
strager, line 346: This looks almost identical to `warn_on_comma_operator_in_index`. We should reuse the code using a C++ template. See `consume_semicolon` for an example of what I want.

test/test-parse-warning.cpp
```diff
@@ -390,6 +390,68 @@ TEST_F(test_parse_warning,
...
+        p.errors,
+        ElementsAreArray({
+            DIAG_TYPE_OFFSETS(
+                p.code, diag_misleading_comma_operator_in_conditional_statement,
+                comma, strlen(u8"if(false"), u8","_sv),
+        }));
+  }
+
+  {
+    test_parser p(u8"do{i++}while((i < 0), true)"_sv, capture_diags);
```
strager, line 407: Are the parentheses necessary to trigger the warning? Could we write the test like this instead:
```suggestion
    test_parser p(u8"do{i++}while(i < 0, true)"_sv, capture_diags);
```

test/test-parse-warning.cpp
```diff
@@ -390,6 +390,68 @@ TEST_F(test_parse_warning,
...
+                p.code, diag_misleading_comma_operator_in_conditional_statement,
+                comma, strlen(u8"do{i++}while((i < 0)"), u8","_sv),
+        }));
+  }
+
+  {
+    test_parser p(u8"do{i++}while(i < (0, true))"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
```
strager, line 422: 👍

test/test-parse-warning.cpp
```diff
@@ -390,6 +390,68 @@ TEST_F(test_parse_warning,
...
+                p.code, diag_misleading_comma_operator_in_conditional_statement,
+                comma, strlen(u8"for(; i < 5"), u8","_sv),
+        }));
+  }
+
+  {
+    test_parser p(u8"for(let i = 0, j = 0;;){}"_sv, capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
```
strager, line 440: We should also test the following cases which should report no diagnostic:
```javascript
for(i = 0, j = 0;;){}
for(;; ++i, ++j){}
```

strager (approved):

src/quick-lint-js/fe/parse.cpp
yunusey: I could not make use of the structure in the `consume_semicolon` function, because these two diagnostic functions take different arguments. Besides, their corresponding diagnostic classes have different shapes. So, instead, I put the common logic here into a separate function `find_comma_operator`.  Let me know if you would still prefer it the other way.

yunusey:
> Let me know when I should merge this.

If all looks good to you now, please go ahead and merge it.

src/quick-lint-js/fe/parse.cpp
strager: > these two diagnostic functions take different arguments

Ah, I did not see that `warn_on_comma_operator_in_index` needed `left_square_span`.

> instead, I put the common logic here into a separate function `find_comma_operator`. Let me know if you would still prefer it the other way.

I prefer the way you had it before.

yunusey: Hi @strager, reverted the related changes.

Issue #897, opened 2022-12-25
Website: improve mobile experience on home page


strager:
- [ ] Remove install CTAs
- [ ] Shrink features table

strager:
ctyjr complained: https://imgur.com/a/dtFDTMe

yasserare:
can you assign me this issue ?

PR #952, opened 2023-03-15
Website: improve mobile experience on home page #897

yasserare:
-Created a responsive Table for mobile only (max-width 684 which is the point where the original table starts being not responsive)
-Implemented the Design approved by starger.


yasserare:

yasserare:
recheck

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+            </tr>
+          </tbody>
+        </table>
```
strager, line 902: This information seems to be duplicated. For example, when we implement TypeScript support in quick-lint-js, I need to remember to update two tables. (I'll probably forget to update the mobile version!)

Can we reuse the existing markup? Alternatively, could we generate both tables from a single source of data? (We can use EJS here.)

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+                Rules
+              </td>
+              <td class="meh">
+                <a href="https://jshint.com/docs/options/">some</a><br />Lint
+                Rules
+              </td>
+            </tr>
+            <tr>
+              <td class="good quick-lint-js">
+                <span aria-label="stable">✅</span><br />Stable
```
strager, line 892: I think aria-label is unnecessary because we have the label immediately after in the markup.

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+              <td class="meh quick-lint-js">
+                <a href="errors/">minimal</a><br />Lint Rules
+              </td>
+              <td class="good">
+                <a href="https://eslint.org/docs/rules/">zillions</a><br />Lint
+                Rules
+              </td>
+              <td class="meh">
+                <a href="https://jshint.com/docs/options/">some</a><br />Lint
+                Rules
```
strager, line 887: I'd adjust the capitalization and make the phases grammatically correct:
```suggestion
                <a href="errors/">minimal</a><br />lint rules
              </td>
              <td class="good">
                <a href="https://eslint.org/docs/rules/">zillions</a><br />of lint
                rules
              </td>
              <td class="meh">
                <a href="https://jshint.com/docs/options/">some</a><br />lint
                rules
```

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+
+            <tr>
+              <td class="good quick-lint-js">free<br />GPLv3</td>
+              <td class="meh">
+                open<br />
+                source<br />MIT
+              </td>
+              <td class="meh">
+                open<br />
+                source<br />MIT
```
strager, line 851: "open source" should not have a forced line break:
```suggestion
                open source<br />MIT
              </td>
              <td class="meh">
                open source<br />MIT
```

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+              <th><a href="https://eslint.org/">ESLint</a></th>
+              <th><a href="https://jshint.com/">JSHint</a></th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td class="good quick-lint-js has-link">
+                <a href="benchmarks/"
+                  ><span aria-labelledby="feature-speed-qljs-quick">🚀</span
+                  ><br /><span id="feature-speed-qljs-quick">quick</span></a
```
strager, line 820: This link has underlines in the mobile version, which look ugly. The underlines don't appear in the desktop version. Remove the underlines.

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+                source<br />MIT
+              </td>
+            </tr>
+            <tr>
+              <td class="good quick-lint-js">
+                <span aria-label="yes">✅<br />JSX</span>
+              </td>
+              <td class="meh">
+                <span aria-label="yes">✅<br />JSX</span>
+              </td>
```
strager, line 860: The quick-lint-js and ESLint columns contain the same text and icons, but have different background colors. This is confusing.

The desktop version has a † symbol on the ESLint version explaining why ESLint is yellow and not green. We should do something similar for the mobile version.

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+                ><br /><span id="feature-speed-jshint-slow">slow</span>
+              </td>
+            </tr>
+            <tr>
+              <td class="good quick-lint-js">
+                <span aria-label="easy">✅</span><br />zero config
+              </td>
+              <td class="meh">
+                <span aria-label="hard">❌</span><br />config or<br />
+                <code>init</code>
```
strager, line 838: Why the `<br>`?
```suggestion
                <span aria-label="hard">❌</span><br />config or
                <code>init</code>
```

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
         </figcaption>
       </figure>
+
+      <figure class="comparison-mobile">
+        <table>
+          <thead>
+            <tr class="header">
+              <th class="quick-lint-js">quick-lint-js</th>
+              <th><a href="https://eslint.org/">ESLint</a></th>
+              <th><a href="https://jshint.com/">JSHint</a></th>
```
strager, line 812: I think we should compare against ESLint and Rome instead of ESLint and JSHint.

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+            </tr>
+            <tr>
+              <td class="good quick-lint-js">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
```
strager, line 898: Use lower case for consistency:
```suggestion
                <span aria-label="stable">✅</span><br />stable
              </td>
              <td class="good">
                <span aria-label="stable">✅</span><br />stable
              </td>
              <td class="good">
                <span aria-label="stable">✅</span><br />stable
```

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+                <span aria-label="no">❌<br />JSX</span>
+              </td>
+            </tr>
+            <tr>
+              <td class="bad quick-lint-js">
+                <span aria-label="no">❌<br />TypeScript</span>
+              </td>
+              <td class="meh">
+                <span aria-label="yes">✅<br />TypeScript</span>
+              </td>
```
strager, line 871: (Same comment as for JSX.)

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+              <td class="good quick-lint-js">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+              <td class="good">
+                <span aria-label="stable">✅</span><br />Stable
+              </td>
+            </tr>
```
strager, line 900: I don't think we need to bother with a "stable" column if all of the linters we show are stable. What do you think?

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+          </tbody>
+        </table>
+      </figure>
+      <figure id="editors-support">
+        <h2>Support</h2>
```
strager, line 905: This should be titled "Editor support", not just "Support".

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+        <div id="supported-editors">
+          <a href="install/vscode/"
```
strager, line 907: I think these buttons look lame. Maybe a background, making them look like buttons, would help?

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+            ><qljs-icon name="vscode" size="19" /><br /><br />VScode</a
```
strager, line 908: It's spelled "Visual Studio Code" or "VS Code".

website/public/index.ejs.html
```diff
@@ -802,6 +802,128 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+          >
+          <a href="install/vim/"
+            ><qljs-icon name="vim" size="19" /><br /><br />Vim</a
+          >
+          <a href="install/neovim/"
+            ><qljs-icon name="neovim" size="19" /><br /><br />Neovim</a
+          >
+          <a href="install/sublime/"
+            ><qljs-icon name="sublime-text" size="19" /><br /><br />Sublime</a
```
strager, line 917: It's spelled "Sublime Text" if we have room.

website/public/main.css
```diff
@@ -296,15 +296,18 @@ nav.breadcrumbs ol > li {
   background-color: #ffeecc;
 }
 @media (prefers-color-scheme: dark) {
-  .comparison td.bad {
+  .comparison td.bad,
+  .comparison-mobile td.bad {
```
strager, line 300: Must fix: These rules only apply in dark mode, not light mode. The table is missing colors in light mode.

I think you should fix this by adding the `comparison` class to the table rather than updating all of the individual CSS selectors here.

website/public/main.css
```diff
@@ -718,7 +721,48 @@ kbd.keystroke {
...
+    display: initial;
+  }
+  .comparison-mobile table {
+    width: 100%;
+    text-align: center;
+    border-collapse: collapse;
+    border: 1px solid gray;
+  }
+  .comparison-mobile table th,
+  td {
```
strager, line 744: Must fix: Do not apply rules to all `td` elements. I think you meant this
```suggestion
  .comparison-mobile table td {
```

website/public/main.css
```diff
@@ -718,7 +721,48 @@ kbd.keystroke {
...
+    padding: 0.5em 0em;
+    border: 1px solid gray;
+  }
+  #editors-support {
+    display: initial;
+  }
+  #supported-editors {
+    display: grid;
+    grid-template-columns: repeat(3, 1fr);
```
strager, line 753: I would have expected `display: flex;` here. What is the advantage of `display: grid;`?

website/public/index.ejs.html
yasserare: That seems logical.

website/public/index.ejs.html
yasserare: I thought it would look good ! I will remove it

website/public/main.css
yasserare: there is no huge advantage, besides that grid requires lines of css to achieve that layout than flex does. Do you want me to change it to flex, I think grid is fine.

yasserare:
In the most recent commit, the major changes are :
-Used EJS to display table so we can avoid updating one and forgetting the other (only them exist in  `comparison-table.ejs.html`)
-Compared with ROME and ESLINT 
I did many other changes but the above are the major ones.

website/public/main.css
strager: Grid is fine.

PR #956, opened 2023-03-22
Website: improve mobile experience on home page issue #897

yasserare:
-Removed CTAs on Mobile
-Shrank The first Table on Mobile
-Replaced The second table with the "Editors Support" section

### DIFF
--- a/website/public/index.ejs.html
+++ b/website/public/index.ejs.html
@@ -442,70 +442,69 @@ const language = "JavaScript";
       <p><a href="demo/">Try it out in your browser.</a></p>
 
       <%- await spriteSheet.makeInlineHTMLAsync() %>
-
-      <h2>Install quick-lint-js</h2>
-      <p>
-        Latest version:
-        <a href="releases/" title="Release history"
-          ><%= qljsVersionInfo.version %></a
-        >
-        (released
-        <time datetime="<%= qljsVersionInfo.releaseDate %>"
-          ><%= new Intl.DateTimeFormat("en-US", {dateStyle: "long", timeZone:
-          "UTC"}).format(new Date(qljsVersionInfo.releaseDate)) %></time
-        >)
-      </p>
-
-      <ul class="install-options">
-        <li>
-          <a
-            href="https://marketplace.visualstudio.com/items?itemName=quick-lint.quick-lint-js"
-            ><qljs-icon
-              name="vscode"
-              size="19"
-              aria-labelledby="install-vs-code"
-            />
-            <span id="install-vs-code">install VS Code plugin</span></a
+      <div id="CTA">
+        <h2>Install quick-lint-js</h2>
+        <p>
+          Latest version:
+          <a href="releases/" title="Release history"
+            ><%= qljsVersionInfo.version %></a
           >
-        </li>
-        <li>
-          <a href="install/neovim/"
-            ><qljs-icon
-              name="neovim"
-              size="19"
-              aria-labelledby="install-neovim"
-            />
-            <span id="install-neovim">install Neovim plugin</span></a
-          >
-        </li>
-        <li>
-          <a href="install/vim/"
-            ><qljs-icon name="vim" size="19" aria-labelledby="install-vim" />
-            <span id="install-vim">install Vim plugin</span></a
-          >
-        </li>
-        <li>
-          <a href="install/cli/"
-            ><qljs-icon
-              name="cli-and-lsp-server"
-              size="19"
-              aria-labelledby="install-cli"
-            />
-            <span id="install-cli">install CLI</span></a
-          >
-        </li>
-        <li>
-          <a href="https://www.npmjs.com/package/quick-lint-js"
-            ><qljs-icon name="npm" size="19" aria-labelledby="install-npm" />
-            <span id="install-npm">install CLI on npm</span></a
-          >
-        </li>
-      </ul>
-      <p>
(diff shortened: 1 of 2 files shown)


website/public/index.ejs.html
```diff
@@ -616,24 +619,32 @@ <h2 id="linter-feature-comparison">Features</h2>
...
               <th>lint rules</th>
-              <td class="meh quick-lint-js"><a href="errors/">minimal</a></td>
+              <td class="meh quick-lint-js">
+                <a href="errors/">minimal</a
+                ><span class="show-mobile"><br />lint rules</span>
+              </td>
               <td class="good">
-                <a href="https://eslint.org/docs/rules/">zillions</a>
+                <a href="https://eslint.org/docs/rules/">zillions</a
+                ><span class="show-mobile"><br />of lint rules</span>
```
strager, line 640: This "of lint rules" label seems to show even on the desktop view. Shouldn't this be visible only in the mobile view?

website/public/main.css
```diff
@@ -718,7 +718,73 @@ kbd.keystroke {
...
+}
+@media only screen and (max-width: 43em) {
+  /* Hide the feature, JShint and Deno columns */
+  .comparison tr th:first-child,
+  .comparison table th:nth-child(4),
+  .comparison table th:nth-child(6),
+  .comparison table th:nth-child(7),
+  .comparison table td:nth-child(4),
+  .comparison table td:nth-child(6),
+  .comparison table td:nth-child(7),
```
strager, line 732: Hmm, there's got to be a better way...

`col.whatever { visibility: collapse; }` works in Blink but not Gecko. =\ From MDN:

> Firefox doesn't hide borders when hiding [<col>](https://developer.mozilla.org/docs/Web/HTML/Element/col) and [<colgroup>](https://developer.mozilla.org/docs/Web/HTML/Element/colgroup) elements if border-collapse: collapse is set.
https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

I think we want `border-collapse: collapse;`, so I guess we are stuck with either your solution to adding a class to each `th` and `td`. =\

website/public/main.css
```diff
@@ -718,7 +718,73 @@ kbd.keystroke {
...
+  .hide-in-mobile {
+    display: none;
+  }
+
+  /* Hide stable, easy npm install rows */
+  .comparison tbody tr:nth-child(3),
+  .comparison tbody tr:nth-child(8) {
+    display: none;
+  }
+  /*how elements that are only for mobile */
```
strager, line 742: Typo?
```suggestion
  /* Show elements that are only for mobile */
```

website/public/main.css
```diff
@@ -718,7 +718,73 @@ kbd.keystroke {
...
+    display: grid;
+    grid-template-columns: repeat(3, 1fr);
+    row-gap: 2em;
+    text-align: center;
+    padding: 2em;
+  }
+  .comparison-mobile figcaption p {
+    margin-top: 0.25em;
+    margin-bottom: 0.25em;
+  }
```
strager, line 759: The `comparison-mobile` class doesn't exist. Delete?
```suggestion
```

website/public/index.ejs.html
```diff
@@ -802,6 +813,38 @@ <h2 id="linter-feature-comparison">Features</h2>
...
+        <h2>Editor support</h2>
+        <div id="supported-editors">
+          <a href="install/vscode/" class="editors-support-buttons"
+            ><qljs-icon name="vscode" size="19" /><br /><br /><span
+              >VS code</span
+            ></a
+          >
+
+          <a href="install/vim/" class="editors-support-buttons"
+            ><qljs-icon name="vim" size="19" /><br /><br /><span>Vim</span></a
```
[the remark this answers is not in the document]
yasserare: using <br><br> instead of just <br> make the buttons look better in this particular case.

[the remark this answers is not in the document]
strager: I think we should use CSS to make it look better.

website/public/main.css
```diff
@@ -718,6 +718,59 @@ kbd.keystroke {
...
+@media only screen and (max-width: 43em) {
+  /* Hide the feature, JShint, Deno columns and stable, easy npm install rows*/
+  .hide-in-mobile {
+    display: none;
+  }
+
+  /* show elements that are only for mobile */
+  .show-in-mobile {
+    display: initial;
+  }
```
strager, line 756: Nitpick: These comments seem unnecessary.
```suggestion
  .hide-in-mobile {
    display: none;
  }
  .show-in-mobile {
    display: initial;
  }
```

website/public/main.css
```diff
@@ -718,6 +718,59 @@ kbd.keystroke {
...
+  .comparison table {
+    border: 1px solid gray;
+  }
+  /*Hide Calll to action section*/
```
strager, line 760: Nitpick: Fix typos:
```suggestion
  /*Hide call to action section*/
```

strager (approved):
Looks good to me aside from some typos. Let me know when I should land this.

website/public/index.ejs.html
```diff
@@ -382,6 +382,13 @@
           width: auto;
         }
       }
+      /* to ge rid of  horizontal navigation bar */
```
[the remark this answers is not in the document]
yasserare: My bad, I meant _scroll bar_ not a _nav bar_. decreasing the padding made the  horizontal scroll bar disappear.

Issue #950, opened 2023-03-12
10$: Warn on ==, !, literals on left-hand side of ??


strager:
The `??` operator does nothing useful if the left-hand side is never null. We can report a diagnostic in some easy cases.

Some examples of code which should warn:
```typescript
true ?? false
x == y ?? false
!b ?? false
"hi" ?? false
```

Inspiration: @Boshen's linter: https://twitter.com/boshen_c/status/1634813504204181504/photo/2

applepie23:
Hi! I'm a student developer with the University of Michigan, and I, along with another student, are hoping to get involved in an open source project. We would love to take this issue!

strager:
@applepie23 I assigned the task to you. Feel free to ask any questions if you need help.

applepie23:
Hi, I tried to ask this question in Libera.Chat but I think my question disappeared. I am having a lot of trouble getting clang-format version 9.0.X to work. I was wondering if I could manually use clang-format version 16 and the quick-lint-provided .clang-format file to style all the files I edited instead of using the ./tools/clang-format executable. Thank you.

strager:
> Hi, I tried to ask this question in Libera.Chat but I think my question disappeared.

I saw it. But you left within 5 minutes, so I couldn't respond.

> I was wondering if I could manually use clang-format version 16 and the quick-lint-provided .clang-format file to style all the files I edited instead of using the ./tools/clang-format executable.

No, please don't do that!

If you can't set up the right clang-format version, don't worry about formatting. I can format your patch before merging.

strager:
Implemented in #969.

Issue #557, opened 2021-12-04
16$: Fix precedence bugs


strager:
I added a test for precedence, `test_parse_expression.precedence`. There are some known bugs, and the test is incomplete (so there might be more bugs). Let's fix everything once and for all, at least for normal expressions.

strager:
Idea for binary precedence:

As I've found binary precedence bugs, I've been adding new AST nodes. One idea is to keep the AST flat, as originally planned, but when we do need precedence, we re-parse using an efficient algorithm.

For example:
```javascript
selection === "agree" || "strongly agree" === selection
//        ^^^         ^^                  ^^^
```
Let's say we want to check for [E0190: missing comparison; '===' does not extend to the right side of '||'](https://quick-lint-js.com/errors/E0190/). A binary AST node has a list of operators: `===`, `||`, and `===`.

We ask the algorithm to tell us about `||` operators. It scans the list of operators for a `||`. When it finds the `||`, the algorithm scans left of the `||` until it finds something with an equal or lower precedence level. Then it scans right of the `||` until it finds something with an equal or lower precedence level. The result of the algorithm is the stuff to the left (`selection === "agree"`) and the stuff to the right (`"strongly agree" === selection`). The algorithm actually would output a list of such things. I'm thinking of an interface like this:

```c++
// maybe also iterate_binary_deep
template <class OperatorPredicate>
void iterate_binary(
  expression::binary* ast,
  OperatorPredicate is_interesting_operator,
  function_ref<void(expression::binary_slice lhs, token_type operator, expression::binary_slice rhs)>);

// Example call:
iterate_binary(ast, [](token_type t) { return t == token_type::pipe_pipe; },
  [&](expression::binary_slice lhs, token_type, expression::binary_slice rhs) {
    /* if lhs is an === but rhs is a literal, report diagnostic */
  });
```

It might also be useful to have a helper function to query the highest precedence of a binary AST node or slice. This would make it easier for E0190's check to look see whether the lhs of `||` is a `==` or something uninteresting (e.g. `+`):
```javascript
class expression::binary {
  ...
  // Returns an index into operators_.
  int highest_precedence() const;
  // A slice would return std::optional<int>, because a slice might be have no operators.
};
```

To make the left+right searching algorithm efficient, we could encode precedence levels inside `token_type`. For example:
```c++
enum class token_type {
  ...

  // lower precedence level
  equal_equal = 0x10,  // ==
  bang_equal = 0x11, // !=

[...]

PR #969, opened 2023-04-05
Issue #950: Add nullish_coalescing_operator diagnostic

applepie23:
This new diagnostic feature is a warning regarding an expression. The nullish coalescing operator resolves to the right operand when the left operand is null. However, if the left operand is never null, it is pointless since the overall expression (LHS ?? RHS) will always resolve to the left operand. Typically, only variables or function call return values have the ability to be null in Javascript. When parse-expression.cpp recognizes a binary operator, we check that the operator is the nullish operator and that the left child/children are not a variable or function call and report the diagnostic.


applepie23:

src/quick-lint-js/fe/parse.cpp
```diff
@@ -734,6 +734,55 @@ void parser::consume_semicolon() {
...
+
+void parser::check_lhs_for_null_potential(expression* lhs, expression* rhs,
+                                           source_code_span op_span) {
+  switch(lhs->kind()){
+    case expression_kind::variable:
+        return;
+    case expression_kind::call:
+        return;
+    default:
+        // expressions that are not variables or function calls are never null
```
strager, line 779: Must fix: This is incorrect. There are other kinds of expressions, such as property lookups and parenthesized expressions, which might return null or undefined. The following uses of `??` all yield `true` but your code incorrectly issues warnings about them:
```javascript
async function f() {
  "use strict";

  console.log(null ?? true);
  var foo = null;
  console.log((foo) ?? true);
  console.log({}.missingProp ?? true);
  console.log({}['missingProp'] ?? true);
  console.log(await foo ?? true);
  console.log(void 42 ?? true);
  function bar() { return null; }
  console.log(bar`hello` ?? true);
  console.log(this ?? true);
}
f();
```
I think you should use a *whitelist* of never-null/undefined expressions rather than a *blacklist* of sometimes-null/undefined expressions.

test/test-parse-warning.cpp
```diff
@@ -672,6 +672,68 @@ TEST_F(test_parse_warning,
...
+  {
+    test_parser p(u8"s.toLowerCase() ?? false"_sv, capture_diags);
+    p.parse_and_visit_expression();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
+  {
+    test_parser p(u8"s ?? false"_sv, capture_diags);
+    p.parse_and_visit_expression();
+    EXPECT_THAT(p.errors, IsEmpty());
+  }
```
strager, line 735: Nitpick: You can make these tests shorter if you remove `capture_diags`:
```suggestion
  {
    test_parser p(u8"s.toLowerCase() ?? false"_sv);
    p.parse_and_visit_expression();
  }
  {
    test_parser p(u8"s ?? false"_sv);
    p.parse_and_visit_expression();
  }
```

strager (changes requested):

src/quick-lint-js/fe/parse.cpp
```diff
@@ -734,6 +734,55 @@ void parser::consume_semicolon() {
...
+    return s == u8"??"_sv;
+  };
+
+  auto no_left_operand = [](string8_view s) -> bool {
+    return s == u8"??"_sv;
+  };
+
+  if (no_left_operand(ast->span().string_view())){
+    return;
+  }
```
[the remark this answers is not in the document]
applepie23: test_parse_expression_statement.statement_starting_with_binary_only_operator failed if I don't include this check. This makes sure we don't report the nullish diagnostic when the expression is missing a left operand since its more relevant to report that the lhs is missing rather than it is always null. However, changing from blacklist to whitelist should make this check unnecessary.

src/quick-lint-js/fe/parse.cpp
applepie23: This was my approach at first, but I switched to blacklist since I could not figure out how to deal with multichild lhs expression when there were no parentheses. However, when trying to find examples to go off from, i found that other diagnostics like error_on_pointless_string_compare seem to assume that the left and right operands will be wrapped in parentheses, so i'm not sure its necessary for me to account for that.

test/test-parse-warning.cpp
applepie23: I removed those and the tests passed the same so you're definitely right. So is the capture_diags parameter only necessary for false negative testing? I ask since other tests in this file that test for IsEmpty() include capture_diags. I also see that you deleted EXPECT_THAT(p.errors, IsEmpty()); I thought that was neccessary too, but it seems to work exactly the same without it, so I can delete those statements as well. I only ask since I based my test cases off of other test cases in this file which follow this format.

strager: > So is the capture_diags parameter only necessary for false negative testing?

If you add `capture_diags`, then `p.errors` is filled with whatever diagnostics are emitted by the code. You can then write an assertion for `p.errors`.

If you do not add `capture_diags`, then any diagnostic will cause the test to fail.

> I ask since other tests in this file that test for IsEmpty() include capture_diags. [...] I only ask since I based my test cases off of other test cases in this file which follow this format.

Yeah, those tests must be old and not written in the new style.

test/test-parse-warning.cpp
```diff
@@ -672,6 +672,97 @@ TEST_F(test_parse_warning,
...
+    p.parse_and_visit_expression();
+  }
+  {
+    test_parser p(u8"bar`hello` ?? false"_sv);
+    p.parse_and_visit_expression();
+  }
+  {
+    test_parser p(u8"this ?? false"_sv);
+    p.parse_and_visit_expression();
+  }
```
strager, line 764: Nitpick: To shrink this code further, we could use a loop:
```suggestion
  for (string8_view code : {
           u8"s.toLowerCase() ?? false"_sv,
           u8"s ?? false"_sv,
           u8"null ?? false"_sv,
           u8"(foo) ?? false"_sv,
           u8"{}.missingProp ?? false"_sv,
           u8"{}['missingProp'] ?? false"_sv,
           u8"await foo ?? false"_sv,
           u8"void 42 ?? false"_sv,
           u8"bar`hello` ?? false"_sv,
           u8"this ?? false"_sv,
       }) {
    SCOPED_TRACE(out_string8(code));
    test_parser p(code);
    p.parse_and_visit_expression();
  }
```

strager (changes requested):

src/quick-lint-js/fe/parse.cpp
```diff
@@ -734,6 +734,81 @@ void parser::consume_semicolon() {
...
+        return;
+        }
+    case expression_kind::_typeof:
+        this->diag_reporter_->report(diag_pointless_nullish_coalescing_operator{
+            .question_question = op_span});
+        return;
+    case expression_kind::binary_operator:{
+        auto* operator_lhs = static_cast<expression::binary_operator*>(lhs);
+        source_code_span binary_op_span = operator_lhs->operator_spans_[0];
+        if (binary_operator_is_never_null(binary_op_span.string_view())){
```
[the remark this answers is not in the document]
applepie23: I will try to look into this, but I'm not entirely sure how to differentiate the main operator of the expression and the sub-operators of the left and right operands. I checked expression.h and there doesn't seem to already be a class method to return this. Are there any existing helper functions that do this? Or do you think it would be a good solution to write a helper function that returns the lowest priority operator, and breaking ties based on which is the leftmost operator?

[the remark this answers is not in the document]
strager: > I'm not entirely sure how to differentiate the main operator of the expression and the sub-operators of the left and right operands.

Yeah. quick-lint-js hasn't needed to deal with precedence much before, so the support for it is weak.

> Are there any existing helper functions that do this?

Not right now. I do have an idea for how it'd look though; see #557.

> So i'm thinking that it might just be enough to return false if 1 of the operators are one of these 4 && ?? , ||

Sounds good!

If you can't find a reliable method (without writing lots of code), then I say don't worry about it for now and just ignore binary operators to the left of `??`.

strager:
Sorry for the delay. I didn't realize you updated this PR with new commits.

I landed your changes in commit edec0892d64aad3dc76f5d1e4fa4f4931ecf9993. That commit includes formatting with clang-format and fixes to some warnings.

Thanks for contributing!

Issue #979, opened 2023-04-11
Website: copy-pasting example on home page copies diagnostic text


strager:
https://quick-lint-js.com/
Copy-pasting from Firefox gives me this:
```
const occupation = language + " coder";
const language = "JavaScript";
occupation = "Senior " + occupation;

E0057 - use of undeclared variable: ocupationconsole.log(`Welcome, ${ocupation}!`);
```

The "E0057 - use of undeclared variable: ocupation" part shouldn't be included in the paste.

PR #980, opened 2023-04-11
BugFix: Fix copy and paste of code from demo.

LeeWannacott:
Issue: #979 we use `user-select:none` on CSS with id `error-box`
gets rid of this message "0057 - use of undeclared variable: ocupationconsole.log(`Welcome, ${occupation}!`);note: creating worker"

### DIFF
--- a/website/public/main.css
+++ b/website/public/main.css
@@ -532,6 +532,7 @@ kbd.keystroke {
   padding: 5px;
   font-size: 0.8rem;
   max-width: 300px;
+  user-select: none;
   z-index: 100;
 }
 


website/public/main.css
```diff
@@ -532,6 +532,7 @@ kbd.keystroke {
   padding: 5px;
   font-size: 0.8rem;
   max-width: 300px;
+  user-select: none;
```
strager, line 535: I think this deserves a comment explaining why we want to disable text selection.

strager:
Another approach: `display: none;` for the offending element after the animation is finished. What do you think about that instead of disabling text selection?

strager (approved):

website/public/main.css
LeeWannacott: @strager You will never be able to select the error box if it's to the right of your cursor; this special case is just because you set the error box span and do an animation. We could just set it inline CSS on this specific span :thinking: and undo it in main.css . Up to you..., or I can put a comment...

LeeWannacott:
> Another approach: `display: none;` for the offending element after the animation is finished. What do you think about that instead of disabling text selection?

Possibly, But you would still be able to copy it, while the error message is present... Another-words you have to wait for the animation to play out.

I haven't looked at the code for animation yet...

Issue #937, opened 2023-02-08
add to Webi package repository


strager:
https://webinstall.dev/

yasserare:
hi ! can you assign me this one ?

strager:
This video might help: https://www.youtube.com/watch?v=RDLyJtiyQHA

yasserare:
I am almost finished, I am just busy with school these days, I will be finishing this issue in the next few weeks.
Sorry for the delay.

coolaj86:
@strager Any particular reason to not use GitHub releases for the binary assets?
https://github.com/quick-lint/quick-lint-js/releases

That makes it much simpler - it's what most projects use.

It appears that https://c.quick-lint-js.com/webinstaller/quick-lint-js.json is broken.

strager:
> @strager Any particular reason to not use GitHub releases for the binary assets?
> https://github.com/quick-lint/quick-lint-js/releases

I prefer my own infrastructure be the source of truth, not GitHub. I could certainly mirror assets to GitHub releases.

> It appears that https://c.quick-lint-js.com/webinstaller/quick-lint-js.json is broken.

Yes, this is expected. @yasserare did not finish his version merging script. I did publish a JSON file for one version: https://c.quick-lint-js.com/releases/2.14.0/webinstaller/quick-lint-js.json

yasserare:
@strager   I apologize for disappearing and not completing the contribution; I will resume work on it.

PR #982, opened 2023-04-13
Issue#937 - step one: script to create version-specific webinstaller/…

yasserare:
In order to add quick-lint-js to webi, we need a file called releases.js that needs a bunch of informations about each QLJS release.
the > make-webinstaller-manifest.js
create a json file that contains all the informations releases.js will need based on quick-lint-js-template.json.
this version of the script is better than the one I showed you in stream, this version takes in arguments, checks if they are valid and display an error message as appropriate.

### DIFF
--- a/dist/webinstaller/make-webinstaller-manifest.js
+++ b/dist/webinstaller/make-webinstaller-manifest.js
@@ -1,47 +1,63 @@
-const fs = require('fs');
+// Copyright (C) 2020  Matthew "strager" Glazar
+// See end of file for extended copyright information.
 
-const [, , ...args] = process.argv;
-const [baseUriArg, BaseURI, releaseDateArg, releaseDate, outputArg, outputFile] = args;
+const fs = require("fs");
 
-function validateArg(args, argName, regex) {
-    const argIndex = args.indexOf(argName);
-    if (argIndex < 0) {
-      console.log(`error: missing ${argName}`);
-      return false;
-    }
-  
-    const argValue = args[argIndex + 1];
-    if (!regex.test(argValue)) {
-      console.log(`error: invalid ${argName}; must match regular expression: ${regex}`);
-      return false;
-    }
-  
-    return true;
-  }
-  
-  if (!validateArg(args, '-BaseURI', /^https?:\/\/.*\/$/)) {
-    return 0;
-  }
-  
-  if (!validateArg(args, '-releaseDate', /^\d{4}-\d{2}-\d{2}$/)) {
-    return 0;
-  }
-  
+const args = process.argv.slice(2);
+const releaseVersion = process.argv[3];
+const releaseDate = process.argv[5];
+const outputFile = process.argv[7];
 
-//get the version
-const releasesRegex = /\/(\d+\.\d+\.\d+)\/$/;
-const releasesVersion = BaseURI.match(releasesRegex);
-
-const template = fs.readFileSync('quick-lint-js-template.json', 'utf8');
-
-const data = JSON.parse(template);
-
-// set version and insert download with the approriate value 
-for (const release of data.release) {
-    release.version = releasesVersion[1];
-    release.download = release.download.replace('${version}', releasesVersion[1]);
-    release.date = releaseDate;
+if (!args.includes("-releaseVersion")) {
+  console.log("error: missing -releaseVersion");
+  return 0;
+}
+if (!args.includes("-releaseDate")) {
+  console.log("error: missing -releaseDate");
+  return 0;
+}
+if (!args.includes("-Out")) {
+  console.log("error: missing -Out");
+  return 0;
 }
 
-// Write the modified data back to a file
-fs.writeFileSync(outputFile, JSON.stringify(data, null, 2), 'utf8');
+const releaseVersionRegex = /^\d+\.\d+\.\d+$/;
+if (!releaseVersionRegex.test(releaseVersion)) {
+  console.log(
+    "error: invalid -releaseVersion; must match regular expression: " +
+      releaseVersionRegex
+  );
+  return 0;
+}
+
+const releasesDateRegex = /^\d{4}-\d{2}-\d{2}$/;
+if (!releasesDateRegex.test(releaseDate)) {
+  console.log(
+    "error: invalid -releaseDate; must match regular expression: " +
+      releasesDateRegex
+  );
+  return 0;
+}
+
+let data = fs.readFileSync("quick-lint-js-template.json", "utf8");
+data = data.replace(/{version}/g, releaseVersion);
+data = data.replace(/\${releaseDate}/g, releaseDate);
+fs.writeFileSync(outputFile, data, "utf8");
+
+// quick-lint-js finds bugs in JavaScript programs.
+// Copyright (C) 2020  Matthew "strager" Glazar
+//
+// This file is part of quick-lint-js.
+//
+// quick-lint-js is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// quick-lint-js is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


dist/webinstaller/quick-lint-js-template.json
```diff
@@ -0,0 +1,103 @@
+{
+  "release": [ 
+      {
+      "name": "linux.tar.gz",
```
strager, line 4: Nitpick: Use Prettier to format this file.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,47 @@
...
+    const argValue = args[argIndex + 1];
+    if (!regex.test(argValue)) {
+      console.log(`error: invalid ${argName}; must match regular expression: ${regex}`);
+      return false;
+    }
+  
+    return true;
+  }
+  
+  if (!validateArg(args, '-BaseURI', /^https?:\/\/.*\/$/)) {
```
strager, line 22: This indentation is confusing. Use Prettier to reformat this code.

strager:
Must fix: See the "check copyright" CI job failures.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,47 @@
...
+
+// set version and insert download with the approriate value 
+for (const release of data.release) {
+    release.version = releasesVersion[1];
+    release.download = release.download.replace('${version}', releasesVersion[1]);
+    release.date = releaseDate;
+}
+
+// Write the modified data back to a file
+fs.writeFileSync(outputFile, JSON.stringify(data, null, 2), 'utf8');
```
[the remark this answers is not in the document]
yasserare: I will try this ! this seems much better than what i wrote.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+  
+  if (!validateArg(args, '-releaseDate', /^\d{4}-\d{2}-\d{2}$/)) {
+    return 0;
+  }
+  
+
+//get the version
+const releasesRegex = /\/(\d+\.\d+\.\d+)\/$/;
+const releasesVersion = BaseURI.match(releasesRegex);
```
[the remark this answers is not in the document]
yasserare: I wanted to emulate the scoop script (make-manifest.go) as much as possible after you pointed me to it, I also assumed that this script would be ran in every release folder by another script that gives it the URL of the folder at that moment.
that seems inaccurate given that we included the flags ( -releaseDate ) and now this.
That seems better so I will do it.

yasserare:
in commit #2:
        [X] correct release to relases.
        [X] changed how the script takes input(used slice and saved each input in a variabe).
        [X]  fixed (This webinstaller Markdown link is broken.).
        [X] replace README.md  typos.
        [X] included copyright.
        [X]  simple string replacements.
        [X] use version flag instead of BaseURI.
        [X] formatted the Files using prettier.

dist/webinstaller/make-webinstaller-manifest.js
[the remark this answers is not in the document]
strager: > I wanted to emulate the scoop script (make-manifest.go) as much as possible

Nitpick: dist/scoop/make-manifest.go hard-codes the version number inside the template. (That's not the best idea...) It doesn't parse the URL.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,63 @@
...
+  return 0;
+}
+if (!args.includes("-releaseDate")) {
+  console.log("error: missing -releaseDate");
+  return 0;
+}
+if (!args.includes("-Out")) {
+  console.log("error: missing -Out");
+  return 0;
+}
```
strager, line 22: Must fix: This code still does the wrong thing if arguments are not given in a particular order. We should allow *any* order.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,63 @@
...
+
+const releaseVersionRegex = /^\d+\.\d+\.\d+$/;
+if (!releaseVersionRegex.test(releaseVersion)) {
+  console.log(
+    "error: invalid -releaseVersion; must match regular expression: " +
+      releaseVersionRegex
+  );
+  return 0;
```
strager, line 30: Must fix: We should `process.exit(1);` or something instead. Failures should be signaled with a non-zero process exit code.

dist/webinstaller/make-webinstaller-manifest.js
```diff
@@ -0,0 +1,63 @@
...
+  console.log(
+    "error: invalid -releaseDate; must match regular expression: " +
+      releasesDateRegex
+  );
+  return 0;
+}
+
+let data = fs.readFileSync("quick-lint-js-template.json", "utf8");
+data = data.replace(/{version}/g, releaseVersion);
+data = data.replace(/\${releaseDate}/g, releaseDate);
```
strager, line 44: Why do we use `{}` in one case and `${}` in another case? We should be consistent.

strager (changes requested):
This is looking much better than before. :+1:

yasserare:
in commit#3:
- Made the code take the arguments in no particular order.
- Replaced `retrun(0)` with `process.exit(1)` in failures.
-  Replaced `/{version}/g` with `/\${version}/g` for consistency.

dist/webinstaller/quick-lint-js-template.json
```diff
@@ -0,0 +1,103 @@
+{
+  "releases": [
+    {
+      "name": "linux.tar.gz",
+      "version": "${version}",
+      "lts": false,
+      "channel": "stable",
+      "date": "${releaseDate}",
+      "os": "linux",
+      "arch": "",
```
strager, line 10: Is an empty arch allowed? This is an x86_64/amd64 package.

strager (approved):
The installer script looks good to me. Let me know when I should land your patch.

dist/webinstaller/quick-lint-js-template.json
strager: > I think it would be better if we add the arch: x86_64/amd64 to those empty values.

Yes, please do this.

yasserare: Done.

yasserare:
> The installer script looks good to me. Let me know when I should land your patch.

Land it ASAP if you can, so we can finish off this issue "https://github.com/quick-lint/quick-lint-js/issues/937"

dist/webinstaller/quick-lint-js-template.json
```diff
@@ -7,7 +7,7 @@
       "channel": "stable",
       "date": "${releaseDate}",
       "os": "linux",
-      "arch": "",
+      "arch": "x86_64/amd64",
```
strager, line 10: Must fix: This doesn't look right. I would expect either "x86_64" or "amd64", not both. Did you test this? What does webinstaller expect? What do other webinstaller packages use?

strager (changes requested):

dist/webinstaller/quick-lint-js-template.json
yasserare: You are right. We can't really do that, I will create an object for each of the 3 architectures for the 3 OS's.

yasserare: > You are right. We can't really do that, I will create an object for each of the 3 architectures for the 3 OS's.

@strager  does this seems like the right thing to do ?

yasserare:
- Added the appropriate architectures (aarch64 for arm64), (armv7l for armhf and arm )...
- Webi extensions don't recognize (tar.gz) so I used (tar) which is what was used by other packages.
- Fixed the formatting
I tested on linux mint and windows and it worked with success.

strager (changes requested):
Looks good to me.

strager (approved):

Issue #740, opened 2022-06-02
8$: TypeScript: warn on 'x! == y'


strager:
`x! == y` is legal TypeScript but might have been a typo for `x !== y`. Issue a warning, suggesting parentheses around `x!` if a null assertion was intended.

daethtech:
Hello,
As mentioned earlier in an email, I am able to take up this task to replace the other one I had, let me know if you need anything else in terms of assigning me to this task.
-G W

PR #984, opened 2023-04-13
#740-TypeScript-warn-on-'x!-==-y'

daethtech:
Having issues with build,
Main changes are in diagnostic-types.h and test-parse-warning.cpp Cannot build for some reason.


test/test-parse-warning.cpp
```diff
@@ -25,655 +25,770 @@ using ::testing::IsEmpty;
...
+        EXPECT_THAT(
+            p.errors,
+            ElementsAreArray({
+                DIAG_TYPE_OFFSETS(
+                    p.code, diag_pointless_comp_against_object_literal,
+                    equals_operator, strlen(u8"x === y || ({}) "), u8"!="_sv),
+            }));
+      }
+
+      TEST_F(test_parse_warning, warn_on_mistyped_strict_inequality_operator)
```
strager, line 706: Issue 1: You need to run `./tools/update-translator-sources` again.

Issue 2: You added this `TEST_F` in the middle of an existing `TEST_F`. Put it outside (e.g. move the `}` on line 788 to above line 705).

strager (changes requested):

test/test-parse-warning.cpp
daethtech: Thank you, that worked.
Next question, I am trying to find the part in the code that you described as "In parser::parse_expression_remainder, when we create an expression::non_null_assertion." could you help me locate this part in the code? Is it around line 207?

strager: > Is it around line 207?

No. Creation of a expression::non_null_assertion is on line 1783 of parse-expression.cpp:
https://github.com/quick-lint/quick-lint-js/blob/b9013cd85bfe3c7fd2b754d76a1aa9cf68aec1f6/src/quick-lint-js/fe/parse-expression.cpp#L1782-L1784

daethtech: Ah okay. So, there I would check for the left-hand side is not in parenthesis, and that to the right of the "!" is "==", right? Then we would call the warning.

strager: try it out

daethtech:
I do not know how to do the following, but once I do I can likely implement the diagnostic detection easily:

-How do I check if the left hand side is in parenthesis? I know the function without_paren() gets rid of them, but how would I check to see if the lhs is in them?
-How do I check if the right hand side is '=='? would I check if rhs.string_view() is equal to '==' or should I be checking if it is equivalent to equals_operator or?
-Should the .equals_operator in the report call be op_span?

Any help for these methods is apprecaited


my diagnostic detection in parse.cpp begins on line 554

daethtech:

strager:
It looks like you reformatted the code, making it difficult to find your differences. This will be a problem when we ultimately merge your code. Do you need help fixing the formatting?

> -How do I check if the left-hand side is in parenthesis? I know the function without_paren() gets rid of them, but how would I check to see if the lhs is in them?

Right before your call to `warn_on_mistyped_strict_inequality_operator`, we **just** skipped over the `!` (`token_type::bang`). If you see a `==` after the `!`, then the left-hand-side of `==` was not parenthesized. (If it was, you would see a `)` (`token_type::right_paren`).) Therefore I don't think it makes sense to check for parentheses.

>  -How do I check if the right-hand side is '=='? would I check if rhs.string_view() is equal to '==' or should I be checking if it is equivalent to equals_operator or?

Where you are adding code, we are working with tokens, not with expression ASTs. (That code is *creating* the expression ASTs.)

You can check if the next token is a `==` by writing `if (this->peek().type == token_type::equal_equal)`.

>  -Should the .equals_operator in the report call be op_span?

It should be the `==`. It looks like you're giving the span of the `!`, which is incorrect.

src/quick-lint-js/fe/parse.cpp
```diff
@@ -36,833 +36,1023 @@
...
+    // if (lhs is enclosed within parentheses) {
+    //   then null assertion was intended.
+    //   return
+    // }
+    //
+    // else, then check if rhs is "==" operator (equals)
+    // if (rhs is '==' ) {
+    // then report error
+
+    // print(op_span.string_view()); **CANT USE** (how do I know what op_span is??)
```
strager, line 573: op_span is a source_code_span object. We don't have code for printing a source_code_span... (I normally use gdb to debug this kind of thing, so I haven't added printing support.)

I added support in commit 518915e2b1b13d9ec1945c0d2003c825ff728779. You can apply that patch by running the following Git commands:

```bash
# In the following command, 'upstream' is your name
# for the https://github.com/quick-lint/quick-lint-js/ fork.
git fetch upstream
git cherry-pick 518915e2b1b13d9ec1945c0d2003c825ff728779
```

Then you can write this if you want to printf debug:
```
std::cout << op_span << '\n';
```

daethtech:
> you reformatted the code, making it difficult to find your differences. This will be a problem when we ultimately merge your code. Do you need help fixing the formatting?

Yes, it's an issue with my Visual Studio Code, as every time I save a file it automatically reformates it.
Additionally, I could not quite get the clang-format tool to work. Let me know what you can do/what I should do to fix the formatting problem.

> It should be the `==`. It looks like you're giving the span of the `!`, which is incorrect.

I see, that being the case, would I set equals operator to be `this->peek()`?
Additionally, you originally asked for the red underlines to be under the space between the `!` and the `==` or the entire `! ==`, but I am not sure what I would pass into the diagnostic report to do either of those. But I can just use the `==` if that is okay with you, let me know.

> Then you can write this if you want to printf debug:
> 
> ```
> std::cout << op_span << '\n';
> ```

Thanks! that's very helpful for my print-based style of coding/debugging.

I will implement your suggestions and make another pull request by tomorrow night at the latest.

src/quick-lint-js/fe/parse.cpp
daethtech: > I added support in commit [518915e](https://github.com/quick-lint/quick-lint-js/commit/518915e2b1b13d9ec1945c0d2003c825ff728779). You can apply that patch by running the following Git commands:
> 
> ```shell
> # In the following command, 'upstream' is your name
> # for the https://github.com/quick-lint/quick-lint-js/ fork.
> git fetch upstream
> git cherry-pick 518915e2b1b13d9ec1945c0d2003c825ff728779
> ```

Due to Key issues, I can only use the desktop client for GitHub, Do you know how I could apply that patch on the Desktop client?

strager:
> Yes, it's an issue with my Visual Studio Code, as every time I save a file it automatically reformates it.
>  Additionally, I could not quite get the clang-format tool to work. Let me know what you can do/what I should do to fix the formatting problem.

I suggest disabling this feature of VS Code for now.

I can ultimately fix the formatting when I land your patch, but I think no automatic formatting is better than very broken automatic formatting.

>> It should be the `==`. It looks like you're giving the span of the `!`, which is incorrect.
>
> I see, that being the case, would I set equals operator to be `this->peek()`?

`this->peek()->span()`, yes.

> Additionally, you originally asked for the red underlines to be under the space between the `!` and the `==` or the entire `! ==`, but I am not sure what I would pass into the diagnostic report to do either of those.

A `source_code_span` has a begin and an end. I think you can take the begin of the `!` and the end of the `==`. https://github.com/quick-lint/quick-lint-js/blob/edec0892d64aad3dc76f5d1e4fa4f4931ecf9993/src/quick-lint-js/fe/source-code-span.h#L19-L24

src/quick-lint-js/fe/parse.cpp
```diff
@@ -553,51 +553,11 @@ namespace quick_lint_js
 
   void parser::warn_on_mistyped_strict_inequality_operator(source_code_span op_span)
```
strager, line 554: Minor: Because this function depends on non-obvious state, I think you should inline it into its call in `parse_expression_remainder`.

test/test-parse-warning.cpp
```diff
@@ -774,7 +774,7 @@ namespace quick_lint_js
             }));
```
strager, line 774: Must fix: This test is failing on my machine:
```
/Users/strager/Projects/quick-lint-js/test/test-parse-warning.cpp:774: Failure
Value of: p.errors
Expected: has 1 element that has type diag_mistyped_strict_inequality_operator
  Actual: {}
```
I think it's failing because you're parsing an `if` statement with `parse_and_visit_expression`, but `parse_and_visit_statement` is the function you should be calling.

strager (changes requested):

src/quick-lint-js/fe/parse.cpp
daethtech: You mean as in `void parse_expression_remainder::warn_on_mistyped_strict_inequality_operator(source_code_span op_span)`?

test/test-parse-warning.cpp
daethtech: I will fix this and see if the  tests pass, if so I think we're golden

daethtech:
> > Yes, it's an issue with my Visual Studio Code, as every time I save a file it automatically reformates it.
> > Additionally, I could not quite get the clang-format tool to work. Let me know what you can do/what I should do to fix the formatting problem.
> 
> I suggest disabling this feature of VS Code for now.
> 
> I can ultimately fix the formatting when I land your patch, but I think no automatic formatting is better than very broken automatic formatting.

I was able to disable the auto formatting feature. However, the code is still formatted incorrectly, and I am not sure how I would re-format it. I can however have the code that I wrote be formatted in the form you were looking for. I can try to make a fresh clone of the repository and then try to commit the changes.

daethtech:
I was able to satisfy all the changes you requested, with the exception of the formatting. If you let me know how to go about reformatting, I can take care of that. Otherwise, I can leave it up to you.
Please let me know if you have any further requests/changes.

src/quick-lint-js/fe/parse.cpp
strager: I don't understand your question. (Copy paste mistake?)

I'm saying: you have the following:
```c++
void parse_expression_remainder() {
  ...
  warn_on_mistyped_strict_inequality_operator(x);
  ..
}

void warn_on_mistyped_strict_inequality_operator(op_span) {
  (do stuff with op_span)
}
```
but I think you should write the following instead:
```c++
void parse_expression_remainder() {
  ...
  (do stuff with x)
  ..
}

// The warn_on_mistyped_strict_inequality_operator function no longer exists.
```

strager: (I checked 6219bcc6499d6431d9e08a6a75557bb3b7f7f126 and it looks like you understood what I meant. :+1:)

docs/errors/E0373.md
```diff
@@ -0,0 +1,31 @@
...
+  alert("Not equal!");
+}
+```
+
+Alternatively, if a non-null assertion was the intention, put parenthesis around the left hand side:
+
+```javascript
+let x = 17;
+let y = 38;
+if ((x)! == y) {
```
strager, line 28: I think this example is incorrect. It'll still trigger the warning, right?

strager (approved):
Looks good to me aside from the documentation (maybe; I'm unsure).

strager:
If this is ready to land, let me know.

daethtech:
> Looks good to me aside from the documentation (maybe; I'm unsure).

That's fair, What about the documentation would you want me to change?

strager:
> That's fair, What about the documentation would you want me to change?

I left a comment inline: https://github.com/quick-lint/quick-lint-js/pull/984#discussion_r1173168635

docs/errors/E0373.md
daethtech: Ahhhh good point, I was originally thinking this was `(x)! == y` because that is what's being suggested if a non null assertion is intended, this suggestion still is in the diagnostic but does not prevent the warning from triggering, so this example can be removed.

daethtech:
> If this is ready to land, let me know.

If you mean merge into the project, I'm all for it (after I edit the documentation and make another pull request)

strager:
> after I [...] make another pull request

There's no need to make another pull request. You can push over the existing Git branch on your fork on GitHub (`Issue-#740-TypeScript-warn-on-'x!-==-y'`), and that will reuse this pull request.

But if you're more comfortable making a new pull request, I'm fine with it.

daethtech:
> There's no need to make another pull request. You can push over the existing Git branch on your fork on GitHub (`Issue-#740-TypeScript-warn-on-'x!-==-y'`), and that will reuse this pull request.
> 
> But if you're more comfortable making a new pull request, I'm fine with it.

What I meant was making another commit to this pull request, but just to clarify, how would I be able to merge if there are still formatting issues? Or did you have something else in mind?

strager:
> just to clarify, how would I be able to merge if there are still formatting issues?

I would fix the formatting myself then land your commits.

docs/errors/E0373.md
```diff
@@ -0,0 +1,21 @@
+# E0373: mistyped strict inequality operator
+
+While `x! == y` is legal TypeScript for a non-null assertion, it may be a typo for a strict inequality, as in `x !== y`:
+
+```javascript
+let x = 17;
+let y = 38;
+if (x! == y) {
```
strager, line 8: Must fix: CI caught an issue with this code: This example reports E0373 (as intended), but also reports E0261 (TypeScript non-null assertions are not allowed in JavaScript).

Please avoid reporting E0261 (diag_typescript_non_null_assertion_not_allowed_in_javascript) if E0373 (diag_mistyped_strict_inequality_operator) is reported.

strager (changes requested):

docs/errors/E0373.md
daethtech: Should I move my implementation of the check before the JavaScript vs TypeScript check?
Or is there a way to change the JavaScript in the documentation into typescript as to not trigger the error?

strager:
Landed as commit 62ad0366ebba886952b0d64a7e4de7b2287e25b1. Thanks for contributing. 👍

daethtech:
> Landed as commit [62ad036](https://github.com/quick-lint/quick-lint-js/commit/62ad0366ebba886952b0d64a7e4de7b2287e25b1). Thanks for contributing.

Ah I was just about to make another change since I noticed an error I left, I was literally just about to push it and I noticed your message. Long story short, I accidentally let the warning trigger when the code is JavaScript, but the placement of `==` after the `!` shouldn't matter in JavaScript since the non-null-assertion isn't allowed in the first place. I will make the push anyway since I finished it, but it's okay if you don't see it/put it in the master.

PR #994, opened 2023-04-21
new diag: mistaken do-while loop

footgod368:


strager (changes requested):
This pull request looks incomplete. There are some test cases and documentation, but no code which reports the new diagnostic.

Are you asking for help?

Issue #976, opened 2023-04-09
Improve diagnostic for '(args: number) => {}' in JavaScript mode


strager:
```typescript
let f = (args: number) => {};
```
Expected:
> E0224: TypeScript type annotations are not allowed in JavaScript code

Actual:
> hello.js:1:9: error: unmatched parenthesis [E0056]
> hello.js:1:14: error: missing semicolon after statement [E0027]
> hello.js:1:14: error: unexpected token [E0054]
> hello.js:1:22: error: unmatched parenthesis [E0056]
> hello.js:1:23: error: missing semicolon after statement [E0027]
> hello.js:1:24: error: missing parameters for arrow function [E0105]
> hello.js:1:10: warning: use of undeclared variable: args [E0057]
> hello.js:1:16: warning: use of undeclared variable: number [E0057]

leszek888:
I'd like to tackle this one.

PR #995, opened 2023-04-22
fix(fp): improve diagnostics for type annotations in arrow function

leszek888:
#976 
I've set `.colon_type_annotation = allow_type_annotations::always` when parsing arrow function's params.
This seems to solve the problem.
If I understand it correctly, we force parsing type annotations regardless of TypeScript/JavaScript to avoid all of those syntax errors and then in function `parse_typescript_colon_for_type()` error is reported when not in JavaScript mode.


strager:
Looks great. Thanks!

strager:
Landed in commit 326ee1885322e2aab7b6d386bbcedfa73f33d733.

PR #996, opened 2023-04-26
Add Nix Flake

countoren:
This PR contains Nix flakes with:

quick-lint-js cli,
quick-lint-js vimPlugin,
and template to initialize neovim that has quick-lint-js both plugin and cli ready to work


flake.nix
```diff
@@ -0,0 +1,25 @@
+{
+  inputs.flake-utils.url = github:numtide/flake-utils;
+  outputs = { self, nixpkgs, flake-utils }:
+  flake-utils.lib.eachSystem [ "x86_64-linux" ] (system:
```
strager, line 4: quick-lint-js should work on every platform:
```suggestion
  flake-utils.lib.eachSystem flake-utils.lib.allSystems (system:
```

flake.nix
```diff
@@ -0,0 +1,25 @@
...
+    packages.default = packages.quick-lint-js;
+
+  }) // rec {
+    # Used by `nix flake init -t <flake>`
+    templates.simple-neovim = { 
+      description = "simple neovim with quick-lint-js activated in it"; 
+      path = ./dist/nix/templates/simple-neovim; 
+    };
+    templates.default = templates.simple-neovim;
+  };
```
strager, line 24: I don't think the templates make sense. Please remove the templates.

dist/nix/templates/simple-neovim/flake.nix
```diff
@@ -0,0 +1,27 @@
...
+          start = with pkgs.vimPlugins;[
+            # loaded on launch
+              ale
+              quick-lint-js.packages.${system}.vimPlugin 
+            ];
+            # manually loadable by calling `:packadd $plugin-name`
+            opt = [ ];
+        };
+      };
+    };
```
strager, line 25: Must fix: Now that I learned more about Nix flakes, I think this makes sense. However, I think this belongs inside `/flake.nix`, not in a template. Please move this code into `/flake.nix`.

flake.nix
```diff
@@ -0,0 +1,25 @@
+{
```
strager, line 1: Is it possible to avoid placing flake.nix or flake.lock in the root of the repository? I want to reduce the number of files in the repo root if possible. (default.nix is already annoying IMO...)

strager (changes requested):
Thanks for being patient with me. =]

flake.nix
countoren: I will say it will defeat the purpose of flakes.
the way I would like to look it as flakes are used at the "edge" of the repo to "report" what it contains.
it should be top level by design. for people that run:
```
nix run quick-lint/quick-lint-js 
#or
nix show quick-lint/quick-lint-js 
```
to easily find the implementations.
so you have guaranty that it will be the top level flake.nix

dist/nix/templates/simple-neovim/flake.nix
countoren: Why nix flakes? as a package?
this will make it harder to modified by the user, we want to declare that this is meant to user to modify.
now we might want to create some more templates for people that has vim already and we just create them a file with additional customRC and the plugins adding a comment how they can import it. to their vim configurations.

countoren:
> Thanks for being patient with me. =]

I am glad I can help too. thank you for being patient with me :)

dist/nix/templates/simple-neovim/flake.nix
strager: > as a package?

Yes. That way I can manually test quick-lint-js's Vim plugin with `nix run .#quick-lint-js-vim-ale` or `nix run .#quick-lint-js-neovim-lspconfig`. (I thought that was the point of adding a pre-configured Vim.)

> this will make it harder to modified by the user, we want to declare that this is meant to user to modify.

I think plain documentation is a better solution to this problem. (With your approach, we would need to document how to use the flake template *and* document how to modify the template.)

> now we might want to create some more templates for people that has vim already and we just create them a file with additional customRC and the plugins adding a comment how they can import it. to their vim configurations.

I thought that was what this template was. You are saying we would need *more* templates? That doesn't seem like a good plan.

countoren: Ok maybe you have a point, I moved the template to packages,
I was able to build and run simple-neovim but you might to want confirm yourself it is fully working.
I think the template one was fully working.

flake.nix
```diff
@@ -0,0 +1,35 @@
...
+  inputs.flake-utils.url = github:numtide/flake-utils;
+  outputs = { self, nixpkgs, flake-utils }:
+  flake-utils.lib.eachSystem flake-utils.lib.defaultSystems (system:
```
countoren, line 4: note that I had to change to defaultSystems,
allSystems does not work for you, you are not yet building to all possible systems.

strager (approved):
Looks good!

Please sign the CLA: https://github.com/quick-lint/quick-lint-js/pull/996#issuecomment-1522674116

If you do not want to sign the CLA for any reason, let me know.

countoren:

PR #1005, opened 2023-05-10
feat(typescript): new diagnostic for keywords containing escape in typescript

### LINKED ISSUE #1003: 8$: error on '\u{63}onstructor' in TypeScript
The following code is legal in JavaScript but is illegal in TypeScript:
```javascript
class C {
  \u{63}onstructor() {}  // equivalent to: constructor() {}
}
```
We should either report this issue to upstream TypeScript or make quick-lint-js diagnose this code is illegal TypeScript code.

jaitjacob:
Fixes #1003


jaitjacob:
> **CLA Assistant Lite bot:** Thank you for your contribution! Like many free software projects, you must sign our [Contributor License Agreement](https://github.com/quick-lint/quick-lint-js/blob/master/docs/CLA-v1.md) before we can accept your contribution. You can agree to the Contributor License Agreement by copy-pasting the following message into a new comment on this GitHub pull request:
> 
> _**I have read and hereby agree to quick-lint-js' Contributor License Agreement (CLA-v1.md).**_
> 
> You can retrigger this bot by commenting **recheck** in this Pull Request

jaitjacob:
recheck

jaitjacob:

strager:
Note: diag_keywords_cannot_contain_escape_sequences (E0023) already exists. However, I think a new diagnostic is warranted because the `constructor` problem is TypeScript-specific thus deserves its own separate documentation.

jaitjacob:
the diag report i added was based on my assumption that `parse-class.cpp` guarantees that code within
[`case token_type::reserved_keyword_with_escape_sequence:{...}`](https://github.com/quick-lint/quick-lint-js/blob/d20bb8693493009fe8780adf5ca0e5a46134b9ba/src/quick-lint-js/fe/parse-class.cpp#L420)  would never execute if current keyword does not contain an escape sequence. this assumption turned out to be false. so I've reverted to my failing test commit for a fresh start.

**what next:** as suggested by Strager on the stream I could check for `escape_sequence_list* identifier_escape_sequences` in the `token` type returned by `peek()` to trigger the diag.

jaitjacob:
the diag report i added was based on my assumption that `parse-class.cpp` guarantees that code within
[`case token_type::reserved_keyword_with_escape_sequence:{...}`](https://github.com/quick-lint/quick-lint-js/blob/d20bb8693493009fe8780adf5ca0e5a46134b9ba/src/quick-lint-js/fe/parse-class.cpp#L420)  would never execute if current keyword does not contain an escape sequence. this assumption turned out to be false. so I've reverted to my failing test commit for a fresh start.

**what next:** as suggested by Strager on the stream I could check for `escape_sequence_list* identifier_escape_sequences` in the `token` type returned by `peek()` to trigger the diag.

strager:
> @strager this one weird edge case makes little sense to me,

Wow, that is weird. It looks like TypeScript reports an error only if there is exactly one escape sequence. 🤨

I think you could do something like this:
```c++
if (std::count(p->peek().begin, p->peek().end, u8'\\') == 1) {
  report(...);
}
```

jaitjacob:
thanks @strager nice & easy fix. all tests gucci. please review.

test/test-parse-typescript-class.cpp
```diff
@@ -1933,6 +1933,26 @@ TEST_F(test_parse_typescript_class,
...
+  EXPECT_THAT(
+      p.errors,
+      ElementsAreArray({
+          DIAG_TYPE_OFFSETS(
+              p.code,
+              diag_keyword_contains_escape_characters,  //
+              escape_character_in_keyword, strlen(u8"class C {\n  "), 
+              u8"\\u{63}onstructor"_sv),
+      }));
+}
```
strager, line 1955: Please include a test for JavaScript.

Please include a test for the `\u{63}onstructo\u{72}` example.

Please include a test for `interface`.

jaitjacob: Added following tests,
1. A test to ensure no diag raised for JavaScript code where keyword contains escape.
2. A test for the \u{63}onstructo\u{72} example.
3. Two tests for interface,
        - a test for interface with escaped identifier
        - a test for interface with escaped keyword

test/test-parse-typescript-class.cpp
```diff
@@ -1933,6 +1933,80 @@ TEST_F(test_parse_typescript_class,
...
+  {
+    test_parser p(
+        u8"interface A {\n"_sv
+        u8"  a:\\u{61}ny"_sv
+        u8"}"_sv,
+        typescript_options);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.errors,IsEmpty());
+  }
+}
```
strager, line 2009: Nit: These two tests belong in `test/test-parse-typescript-interface.cpp`.

strager, line 2009: These interface tests don't seem relevant to the patch.

When I ask for a test for interfaces, I had something like this in mind:
```typescript
interface I {
      \u{63}onstructor();  // allowed (no error)
}
```

test/test-parse-typescript-class.cpp
```diff
@@ -1933,6 +1933,80 @@ TEST_F(test_parse_typescript_class,
         }));
   }
 }
+
+TEST_F(test_parse_typescript_class,
+       class_keyword_with_escape_sequence) {
```
strager, line 1938: Nit: I think this test is misnamed:
```suggestion
       constructor_keyword_with_escape_sequence) {
```

strager, line 1938: Because this is strange behavior, I think it deserves a comment. Something like this:
```c++
// NOTE[typescript-constructor-escape]: The TypeScript compiler complains if
// 'constructor' in a class contains an identifier escape sequence, despite
// escapes being allowed in vanilla JavaScript. TypeScript does not complain if
// there is more than one sequence, though.
```

And for the other tests (no_diag_for_more_than_one_escape and class_keyword_with_escape_sequence_legal_in_js), you can reference:
```c++
// See NOTE[typescript-constructor-escape].
```

strager: And maybe reference the note in the implementation code too, 'cus why not? =]

test/test-parse-typescript-class.cpp
```diff
@@ -1933,6 +1933,80 @@ TEST_F(test_parse_typescript_class,
...
+  EXPECT_THAT(
+      p.visits,
+      ElementsAreArray(
+          {"visit_enter_class_scope", "visit_enter_class_scope_body",
+           "visit_property_declaration", "visit_enter_function_scope",
+           "visit_enter_function_scope_body", "visit_exit_function_scope",
+           "visit_exit_class_scope", "visit_variable_declaration"}));
+}
+
+TEST_F(test_parse_typescript_class, class_keyword_with_escape_sequence_legal_in_js) {
```
strager, line 1974: Nit: I think this test is misnamed:

Nit: Spell out 'javascript':
```suggestion
TEST_F(test_parse_typescript_class, constructor_keyword_with_escape_sequence_legal_in_javascript) {
```

strager (approved):

test/test-parse-typescript-class.cpp
jaitjacob: Resolved:

1. Modify test for interfaces with `\u{63}onstructor` as keyword.
2. Remove the redundant test(identifier).
3. Move test from to `test/test-parse-typescript-class.cpp` to `test/test-parse-typescript-interface.cpp.`

test/test-parse-typescript-class.cpp
jaitjacob: Resolved.

jaitjacob: Resolved,
1. Added NOTE above test.
2. Reference 'See NOTE' right above implementation if condition.

test/test-parse-typescript-class.cpp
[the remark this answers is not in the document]
jaitjacob: Keeping as is.

jaitjacob:
@strager Please take a look.

jaitjacob:
@strager sorry I had few merge conflicts resolved incorrectly. I've reverted that commit. Might have to re-run tests.

strager:
Landed as commit 54638656ea45ec92f78f7bad2cbfd37b69e0ccb4.

PR #1011, opened 2023-05-15
fix(fe): show diagnostics when access specifier is not first modifier

leszek888:


strager:
@leszek888 Are you looking for feedback for this PR?

leszek888:
Hi @strager, I haven't had much time last week to continue to work on this PR, that's why nothing has changed in the last week.
It's still work in progress, I've got first working version, and I was planning to make some more refinements, but if you could provide some feedback now, that would be much appreciated.
I've added diagnostics for public/readonly not only in the constructor, but also directly in class body.
I'm checking if `public/protected/private` precedes `static` and `readonly` for class members, and `static`/`async` for methods.

strager:
> I've added diagnostics for public/readonly not only in the constructor, but also directly in class body.

I recommend sticking to one PR per task. I think the code for class properties is very different than the code for constructor parameters, so I view the two scenarios as different tasks.

src/quick-lint-js/diag/diagnostic-types.h
```diff
@@ -2976,6 +2976,15 @@
       diag_unexpected_colon_after_generic_definition, "E0331",                  \
       diagnostic_severity::error, { source_code_span colon; },                  \
       MESSAGE(QLJS_TRANSLATABLE("':' should be 'extends' instead"), colon))     \
+                                                                                \
+  QLJS_DIAG_TYPE(                                                               \
+      diag_access_specifier_must_proceed_other_modifiers, "E0369",              \
+      diagnostic_severity::error, {                                             \
+        source_code_span specifier;                                             \
+        source_code_span modifier;                                              \
```
strager, line 2984: Nit: I think these are confusing names. I don't know what the difference between a specifier and modifier is.

Perhaps 'second_modifier' and 'first_modifier' are better names.

src/quick-lint-js/fe/parse-class.cpp
```diff
@@ -1126,6 +1129,40 @@ void parser::parse_and_visit_class_or_interface_member(
       }
     }
 
+    void error_if_access_specifier_not_first_in_field() {
+      if (p->options_.typescript && modifiers.size()) {
```
strager, line 1133: Nit: Check .empty rather than .size:
```suggestion
      if (p->options_.typescript && !modifiers.empty()) {
```

src/quick-lint-js/fe/parse-class.cpp
```diff
@@ -1126,6 +1129,40 @@ void parser::parse_and_visit_class_or_interface_member(
...
+    void error_if_access_specifier_not_first_in_method() {
+      if (p->options_.typescript && modifiers.size()) {
+        check_if_access_specifier_preceeds_given_modifiers({
+          token_type::kw_static,
+          token_type::kw_async
+        });
+      }
+    }
+
+    void check_if_access_specifier_preceeds_given_modifiers(std::vector<token_type> tokens) {
```
strager, line 1150: Must fix: `std::vector<token_type>` is expensive and its features are unnecessary. Use `span<const token_type>` (from `<quick-lint-js/port/span.h>`) instead.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1952,6 +1952,48 @@ void parser::parse_and_visit_function_parameters(parse_visitor_base &v,
         parameter_property_keyword = readonly_span;
         this->lexer_.commit_transaction(std::move(transaction));
       }
+
+      // consturctor(readonly public value) {}  // Invalid
```
strager, line 1956: Nit: Fix spelling:
```suggestion
      // constructor(readonly public value) {}  // Invalid
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1952,6 +1952,48 @@ void parser::parse_and_visit_function_parameters(parse_visitor_base &v,
         parameter_property_keyword = readonly_span;
         this->lexer_.commit_transaction(std::move(transaction));
       }
+
```
strager, line 1955: This code has since been refactor. Please merge or rebase `master`.

test/test-parse-typescript-class.cpp
```diff
@@ -1709,6 +1751,28 @@ TEST_F(test_parse_typescript_class, parameter_property_in_constructor) {
...
+    SCOPED_TRACE(p.code);
+    p.parse_and_visit_module();
+
+    EXPECT_THAT(
+        p.errors,
+        ElementsAreArray({
+            DIAG_TYPE_OFFSETS(
+                p.code,
+                diag_access_specifier_must_proceed_other_modifiers,
+                specifier, strlen(u8"class C {\n  constructor(readonly "), access_specifier),
```
strager, line 1773: We should use `DIAG_TYPE_2_OFFSETS` to check both `modifier` and `specifier`.

test/test-parse-typescript-class.cpp
```diff
@@ -1033,6 +1033,48 @@ TEST_F(test_parse_typescript_class,
...
+      test_parser p(code.string_view(), typescript_options, capture_diags);
+      p.parse_and_visit_statement();
+
+      EXPECT_THAT(
+          p.errors,
+          ElementsAreArray({
+              DIAG_TYPE_OFFSETS(
+                  p.code,
+                  diag_access_specifier_must_proceed_other_modifiers,
+                  specifier, concat(u8"class C { "_sv, modifier, u8" "_sv).size(), access_specifier),
```
strager, line 1052: We should use `DIAG_TYPE_2_OFFSETS` to check both `modifier` and `specifier`.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1952,6 +1952,48 @@ void parser::parse_and_visit_function_parameters(parse_visitor_base &v,
...
+            switch (this->peek().type) {
+              QLJS_CASE_STRICT_ONLY_RESERVED_KEYWORD:
+              QLJS_CASE_CONTEXTUAL_KEYWORD:
+              case token_type::kw_await:
+              case token_type::dot_dot_dot:
+              case token_type::identifier:
+              case token_type::kw_yield:
+              case token_type::left_curly:
+              case token_type::left_square:
+              case token_type::reserved_keyword_with_escape_sequence: {
```
strager, line 1978: Is this copy-pasted from somewhere? I don't see comprehensive tests justifying each of these cases.

If it's copy-pasted, could we avoid the copy-paste somehow? If we can't reasonbly avoid the copy-paste, I'd like a test for each of these cases.

strager (commented):
I think your solutions overall look good. There are some must-fix issues though.

leszek888:
> I recommend sticking to one PR per task. I think the code for class properties is very different than the code for constructor parameters, so I view the two scenarios as different tasks.

Sure, I can split it into two PRs, one branching off from the other one. I've combined it because they share the same diagnostic, but yeah, the code that actually triggers the diagnostics is different.

strager:
I wrote my comment recommending two PRs before I read the code for this PR. I think one PR is fine if you've already implemented it.

src/quick-lint-js/diag/diagnostic-types.h
leszek888: Yep, specifier and modifier wasn't clear enough. Changed to 'second_modifier' and 'first_modifier'

src/quick-lint-js/fe/parse-statement.cpp
leszek888: Yeah, that was a dirty copy paste which I wanted to refactor and I've found a simpler way to check if access modifier after "readonly" is used as a keyword or identifier by simply skipping a token and checking with already existing lambda function 'is_after_parameter_name' if access modifier was used as an identifier.

src/quick-lint-js/fe/parse-class.cpp
leszek888: Fixed

leszek888:
All suggested fixes applied, PR is no longer marked as WIP.

strager (approved):

strager:
Landed as commit 88fd3e63586ac494026a4c5e182ff2d1f06dee0e.

PR #1019, opened 2023-05-26
Website: Table fix missing thead and move it before tbody

LeeWannacott:
#SemanticHTML. Please don't merge this yet. I need to fix this edge case (having multiple tbody that have no thead) in table-sort-js...

### DIFF
--- a/website/public/index.ejs.html
+++ b/website/public/index.ejs.html
@@ -683,7 +683,7 @@ const language = "JavaScript";
               <td class="good"><span aria-label="stable">✅</span></td>
             </tr>
           </tbody>
-          <thead>
+          <thead class="hide-in-mobile">
             <tr class="header">
               <th>feature</th>
               <th class="quick-lint-js">quick-lint-js</th>
--- a/website/public/index.ejs.html
+++ b/website/public/index.ejs.html
@@ -683,7 +683,6 @@ const language = "JavaScript";
               <td class="good"><span aria-label="stable">✅</span></td>
             </tr>
           </tbody>
-
           <thead>
             <tr class="header">
               <th>feature</th>
--- a/website/public/index.ejs.html
+++ b/website/public/index.ejs.html
@@ -684,7 +684,7 @@ const language = "JavaScript";
             </tr>
           </tbody>
 
-          <tbody class="editors hide-in-mobile">
+          <thead>
             <tr class="header">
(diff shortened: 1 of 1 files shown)


website/public/index.ejs.html
```diff
@@ -683,8 +683,7 @@ <h2 id="linter-feature-comparison">Features</h2>
               <td class="good"><span aria-label="stable">✅</span></td>
             </tr>
           </tbody>
-
-          <tbody class="editors hide-in-mobile">
+          <thead class="hide-in-mobile">
```
strager, line 686: There can be only one thead per table [1], and the thead must not be after a tbody. There's a thead already on line 419, and there's a tbody immediately above this thead. So I don't think this HTML is legal.

[1] https://html.spec.whatwg.org/multipage/tables.html#the-thead-element

strager (changes requested):

strager:
What is the motivation for this change?

website/public/index.ejs.html
LeeWannacott: legal in what sense? It works! you can have thead followed by tbody followed by thead followed by tbody. Maybe its the browser that makes it work even if its not in the spec?

LeeWannacott: > What is the motivation for this change?

I was trying to make the table on quick-lint sortable with firefox plugin. I know you don't care :crying_cat_face:

LeeWannacott: image
But you really don't!

LeeWannacott: Do I take the first row of each subsequent tbody and make it clickable then?

LeeWannacott:
> What is the motivation for this change?
The last commit on quick-lint was a week ago.

LeeWannacott:
I'll close. I guess I should support multiple tbodies differently even though it works : /

website/public/index.ejs.html
strager: If a `tbody` has only one `tr`, and that `tr` only has `th`s, then I think you can treat that `tbody` as a `thead` for table-sort-js.

LeeWannacott: I think you mean if a `tbodys` first `tr` is only `th` `s` . Why would a tbody only have one tr big oof; this is complicated...

PR #1031, opened 2023-06-21
Warn on variable assignment to self being no-op

### LINKED ISSUE #1016: 8$: warn on assignment to rhs
quick-lint-js should warn about code like `x = x;`, i.e. a variable assigned to a variable.

`x.p = x.p;` should *not* warn because there might be side effects. For example, `element.innerHTML = element.innerHTML;` is not a no-op; it removes event listeners.

`x += x;`, `x &&= x;`, etc. should *not* warn.

Docs for creating a diagnostic: https://quick-lint-js.com/contribute/create-diagnostic/

holychowders:
Resolves #1016.

### DIFF
--- a/docs/errors/E0383.md
+++ b/docs/errors/E0383.md
@@ -1,0 +1,21 @@
+# E0383: variable assignment to self is no-op
+
+The following contains a no-op (`next = next;`):
+
+```javascript
+let next = 0;
+if (previous !== null) {
+  next = next;
+}
+switchToPage(next);
+```
+
+To fix this, change the assignment so it's not assigning a variable to itself:
+
+```javascript
+let next = 0;
+if (previous !== null) {
+  next = previous;
+}
+switchToPage(next);
+```
--- a/src/quick-lint-js/fe/parse-expression.cpp
+++ b/src/quick-lint-js/fe/parse-expression.cpp
@@ -85,6 +85,22 @@ void parser::visit_expression(expression* ast, parse_visitor_base& v,
     expression* lhs = ast->child_0();
     expression* rhs = ast->child_1();
     this->visit_assignment_expression(lhs, rhs, v);
+
+    expression* lhs_without_paren = lhs->without_paren();
+    expression* rhs_without_paren = rhs->without_paren();
+
+    bool both_sides_are_variables =
+        lhs_without_paren->kind() == expression_kind::variable &&
+        rhs_without_paren->kind() == expression_kind::variable;
+    if (both_sides_are_variables) {
+      string8_view lhs_variable_name = lhs_without_paren->variable_identifier().normalized_name();
+      string8_view rhs_variable_name = rhs_without_paren->variable_identifier().normalized_name();
+      if (lhs_variable_name == rhs_variable_name) {
+        this->diag_reporter_->report(diag_variable_assigned_to_self_is_noop {
+          .assignment_statement = ast->span(),
+        });
+      }
+    }
     break;
   }
   case expression_kind::compound_assignment:
--- a/test/test-parse-warning.cpp
+++ b/test/test-parse-warning.cpp
@@ -737,8 +737,66 @@ TEST_F(test_parse_warning, warn_on_pointless_nullish_coalescing_operator) {
     p.parse_and_visit_expression();
   }
 }
-}
-}
+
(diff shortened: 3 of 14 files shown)


holychowders:

docs/errors/E0383.md
```diff
@@ -0,0 +1,15 @@
+# E0383: variable assignment to self is no-op
+
+The following is a no-op and will warn:
```
strager, line 3: Nit: "will warn" is misleading. When run with a JavaScript engine, there is no warning. I would say just "The following is a no-op:".

docs/errors/E0383.md
```diff
@@ -0,0 +1,15 @@
...
+
+```javascript
+x = x;
+```
+
+The following is not necessarily a no-op and will not warn:
+
+```javascript
+i.x = i.x;
+```
```
strager, line 13: I think this example is confusing. It has nothing to do with the diagnostic or how to fix the diagnostic.

In our docs, the second code example shows what a fixed program looks like. I'd replace this example with what the user should write instead of the first example. In this case, though, that would be an empty code example.

Perhaps the following is a better example:
```javascript
let next = 0;
if (previous !== null) {
  next = next;
}
switchToPage(next);
```
```javascript
let next = 0;
if (previous !== null) {
  next = previous;
}
switchToPage(next);
```

src/quick-lint-js/fe/parse-expression.cpp
```diff
@@ -85,6 +85,12 @@ void parser::visit_expression(expression* ast, parse_visitor_base& v,
     expression* lhs = ast->child_0();
     expression* rhs = ast->child_1();
     this->visit_assignment_expression(lhs, rhs, v);
+
+    if (lhs->span().string_view() == rhs->span().string_view()) {
```
strager, line 89: Must fix: Here, you are checking that the source code of the left and right side have the same text. This means that `x.p = x.p;` will report `diag_variable_assigned_to_self_is_noop`. It also means that `x = \u{78};` (which is equivalent to `x = x;`) will not report a diagnostic, even though it should.

To fix this, I would check that both `lhs` and `rhs` have `expression_kind::variable`. If so, check `lhs->variable_identifier().normalized_name() == rhs->variable_identifier().normalized_name()`.

It might be a good idea to also handle `(x) = x;` and `x = ((x));`. `expression::without_paren` might be handy.

test/test-parse-warning.cpp
```diff
@@ -737,8 +737,36 @@ TEST_F(test_parse_warning, warn_on_pointless_nullish_coalescing_operator) {
     p.parse_and_visit_expression();
   }
 }
-}
-}
+
+TEST_F(test_parse_warning, warn_on_variable_assigned_to_self_is_noop) {
```
strager, line 741: :+1:

strager (changes requested):

docs/errors/E0383.md
holychowders: Would a better description be "useless variable assignment" instead of referring to it as a "no-op"?

test/test-parse-warning.cpp
holychowders: Added checks to the test:

- Assert `x = \u{78};` warns
- Assert `(x) = x;` warns
- Assert `x = ((x));` warns

docs/errors/E0383.md
```diff
@@ -0,0 +1,21 @@
...
+
+```javascript
+let next = 0;
+if (previous !== null) {
+  next = next;
+}
+switchToPage(next);
+```
+
+To fix this, change the assignment so it's not adding a variable to itself:
```
strager, line 13: Nit: I think you meant "assigning", not "adding":
```suggestion
To fix this, change the assignment so it's not assigning a variable to itself:
```

strager (approved):
Looks good to me. Let me know when I should format then merge. (This PR is still marked as a draft.)

holychowders:
Ready for you to format the code if you're fine with the current language of the diagnostic (https://github.com/quick-lint/quick-lint-js/pull/1031#discussion_r1240914268).

strager:
Landed in commit ea0d7c98dc73a60a7f7f1a97340a6e49fd2dfe36. Thanks!

Issue #1024, opened 2023-06-05
Website: use aria-current to indicate current nav


strager:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current

devinhadley:
I would like to give this a shot if possible.

devinhadley:
My manual testing has indicated that the aria-current="page" is being correctly defined for the current page. However, im struggling to find a way to actually test the implementation using accessibility software. Windows narrator (and chrome Screen Reader) don't seem to treat aria-current="page" specially. However, if I use aria-label it does correctly read whatever the label is for just the current page when it is hovered over.

The fact label is working, and aria-current="page" is valid according to the [docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) i'm not really sure how to proceed with testing. I also tested changing the spans to <a> with an href and still no dice. 

How should I proceed with testing?

devinhadley:
Update aria-current="page" works correctly when testing using the following accesability software: [NVDS](https://www.nvaccess.org/download/)

It appears microsoft narrator does not support aria-current yet.

PR #1038, opened 2023-06-28
use aria-current to indicate current nav

devinhadley:
#1024 
Added aria-current="page" to the correct main nav element. This was done by taking advantage of the pageLinkHTML function which already includes a check for if the current nav item is the current page.


devinhadley:

strager:
This looks good, but I'd like to know how to test it.

devinhadley:
https://github.com/quick-lint/quick-lint-js/issues/1024#issuecomment-1610839669

Issue #1042, opened 2023-07-20
FreeBSD build Broken


AidenThing:
The FreeBSD build is broken, a few things just seemed to be miss named in "file-handle-posix.cpp" in relation to pipes

/home/aiden/tools/quick-lint-js/src/quick-lint-js/io/file-handle-posix.cpp:148:3: error: unknown type name 'pipe_fds'
  pipe_fds pipe = make_pipe();
  ^
/home/aiden/tools/quick-lint-js/src/quick-lint-js/io/file-handle-posix.cpp:153:25: error: use of undeclared identifier 'posix_file_io_error'
    Result<std::size_t, posix_file_io_error> written =
                        ^
1 warning and 2 errors generated.
--- src/CMakeFiles/quick-lint-js-lib.dir/quick-lint-js/io/file-handle-posix.cpp.o ---

I have the patch ready.

strager:
Fixed by @AidenThing in commit 647cd8a01e8565a3a8bb6db31196a3978d2f3749.

PR #1043, opened 2023-07-20
Fix FreeBSD Build

AidenThing:
Fix for the FreeBSD build.


AidenThing:
#1042 is the issue created for this

strager:
Thanks!

PR #1044, opened 2023-07-20
feat: warning on xor operator as exponentiation

### LINKED ISSUE #997: 8$: warn on 2^8, 10^2, etc.
Inspiration: GCC's -Wxor-used-as-pow: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90885

Instead of `2^8`, the user meant `2**8` instead.

We should be cautious about what we diagnose. Perhaps for now only diagnose with a base of `2` or `10` and a literal exponent?

pedrobl85:
Closes #997 

As proposed in the original issue, we only diagnose for cases with a base of 2 or 10 and a literal exponent.

(Had to change a test not created by me because it involved the operation `2 ^ 2` and expected no diagnostics. Changed the operation to  `3 ^ 2`, if another solution is preferred for this just let me know)

### DIFF
--- a/docs/errors/E0710.md
+++ b/docs/errors/E0710.md
@@ -1,8 +1,7 @@
 # E0710: '^' is the XOR operator; to exponentiate, use '**' instead
 
 The Exclusive OR operator ^ can sometimes be mistaken for the
-exponentiation operator **, especially when applied to literal bases 2 and 10
-and literal exponent operations (which could be written as literal itself)
+exponentiation operator **.
 
 ```javascript
 let x = 2 ^ 8
@@ -16,7 +15,7 @@ let x = 2 ** 8
 ```
 
 Alternatively, if the intention was to use the ^ operator for XOR operation,
-The resulting literal could just be used directly (2 ^ 8 = 10)
+The resulting literal could be used directly (2 ^ 8 = 10)
 
 ```javascript
 let x = 10
--- a/docs/errors/E0710.md
+++ b/docs/errors/E0710.md
@@ -1,0 +1,23 @@
+# E0710: '^' is the XOR operator; to exponentiate, use '**' instead
+
+The Exclusive OR operator ^ can sometimes be mistaken for the
+exponentiation operator **, especially when applied to literal bases 2 and 10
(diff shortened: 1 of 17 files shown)


docs/errors/E0710.md
```diff
@@ -0,0 +1,23 @@
+# E0710: '^' is the XOR operator; to exponentiate, use '**' instead
+
+The Exclusive OR operator ^ can sometimes be mistaken for the
+exponentiation operator **, especially when applied to literal bases 2 and 10
+and literal exponent operations (which could be written as literal itself)
```
strager, line 5: Nit: I think the second clause of this sentence makes the whole statement more confusing than it needs to be. I'd remove it.
```suggestion
exponentiation operator **.
```

docs/errors/E0710.md
```diff
@@ -0,0 +1,23 @@
...
+
+If the intention was to exponentiate, the operator ^ should be replaced
+with **
+
+```javascript
+let x = 2 ** 8
+```
+
+Alternatively, if the intention was to use the ^ operator for XOR operation,
+The resulting literal could just be used directly (2 ^ 8 = 10)
```
strager, line 19: Nit: Avoid using "just" in documentation.
```suggestion
The resulting literal could be used directly (2 ^ 8 = 10)
```

strager (approved):
If you rebase, you will need to rename some identifiers:

`test_parser` -> `Test_Parser`
`test_parse_warning` -> `Test_Parse_Warning`
`diag_xor_used_as_exponentiation` -> `Diag_Xor_Used_As_Exponentiation`
`expression_kind` -> `Expression_Kind`
`parser` -> `Parser`
`expression::binary_operator` -> `Expression::Binary_Operator`
`string8_view` -> `String8_View`
`source_code_span` -> `Source_Code_Span`

strager:
I can rebase for you if you want.

strager:
Landed as commit 439f9732acc660319f43053f8c3612df851d53d8.

(I accidentally stripped authorship information. Oops! Sorry.)

PR #1046, opened 2023-07-22
fix: Website: Make code blocks scrollable on mobile view

nikohoffren:
This PR addresses the issue of code blocks not properly wrapping on the website in docs/cli on mobile view, which caused a horizontal scroll bar to appear on the bottom of the screen and the actual code block content overflowing.

As a solution i updated the AsciiDoc to apply a custom CSS class to code blocks and added a CSS rule to make any element with the new class horizontally scrollable, while preserving spacing and line breaks.


nikohoffren:

strager:
It looks like your solution effectively adds a margin between the right side of the code block and the right edge of the viewport. This tricks the viewer into thinking that the content does not scroll. (There's plenty of space available, so why would it scroll? There's also no scroll bar.) This happens even on desktop:

image

I think it's better to have content which is obviously discoverable rather than pretty-looking content which is harder to discover.

I'm open to your solution if the margin is removed or if a scroll bar is visible.

docs/cli.adoc
```diff
@@ -306,6 +307,7 @@ ____
 To lint three files, writing machine-readable messages to _/tmp/vim-qflist.json_:
 ____
 [subs=+quotes]
+[source,bash,subs=+quotes,role="scrollable-container"]
```
strager, line 310: Instead of applying this fix to specific items, could we apply it site-wide? Other pages are affected, such as https://quick-lint-js.com/contribute/create-diagnostic/#report.

nikohoffren: I removed `[source,bash,subs=+quotes,role="scrollable-container"]` from the items and added to main.css:
```css
@media screen and (max-width: 670px) {
  pre, code {
    overflow-x: scroll;
    max-height: 400px;
  }
}
```

However i cannot figure out an solution for the scroll bar to always be visible, maybe that would also be browser dependent too.

I do agree on your comment that it's better to have content which is obviously discoverable rather than fancy content which is harder to discover, so given the minor nature of this issue, implementing such a solution may be a bit of an overkill.

website/public/main.css
```diff
@@ -784,6 +784,14 @@ kbd.keystroke {
   }
 }
 
+/* Scrollable container */
+@media screen and (max-width: 670px) {
+  pre, code {
```
strager, line 789: `overflow-x` and `max-height` are not appropriate for `code`. Only apply these rules to `pre`:
```suggestion
  pre {
```

website/public/main.css
```diff
@@ -784,6 +784,14 @@ kbd.keystroke {
...
+    overflow-x: scroll;
```
strager, line 790: Minor: Is there a reason you chose `scroll` instead of `auto`? Don't think we want code blocks to have a scroll bar if the code isn't that wide.

strager (approved):
(Sorry for the delayed reply.)

I added a screenshot testing tool to the repo. Check out `website/backstop/README.md`.

I tried your latest patch on a few browsers. Here's what I saw:
* Firefox Linux: scroll bars only visible when interacting
* Firefox macOS: scroll bars always visible
* Brave Linux: scroll bars always visible
* Brave macOS: scroll bars always visible
* Safari macOS: scroll bars always visible
* Screenshot test (Chromium-based): scroll bars are not visible
This behavior is acceptable. :+1:

I noticed a problem on a few browsers: <http://127.0.0.1:9001/blog/bug-journey/>
bug demonstration in Brave Linux
With your patch, there is both a horizontal scrollbar and a vertical scrollbar. The horizontal scrollbar is expected, but the vertical scrollbar is surprising. This bug happens with Brave (Linux) and Safari (macOS), but does not happen with Firefox (Linux) or Brave (macOS). This issue might be due to some slight differences in font metrics or something... If there's no quick fix, I'm fine with the behavior of your patch.

Note to self: Looking through a few pages, we really should make some of them more mobile-friendly. There is wasted space in some code blocks due to excessive padding. Your patch makes the problems more obvious (which is a good thing, I think).

website/public/main.css
nikohoffren: Good call, i change it to `auto`.

nikohoffren:
> (Sorry for the delayed reply.)
> 
> I added a screenshot testing tool to the repo. Check out `website/backstop/README.md`.
> 
> I tried your latest patch on a few browsers. Here's what I saw:
> 
>     * Firefox Linux: scroll bars only visible when interacting
> 
>     * Firefox macOS: scroll bars always visible
> 
>     * Brave Linux: scroll bars always visible
> 
>     * Brave macOS: scroll bars always visible
> 
>     * Safari macOS: scroll bars always visible
> 
>     * Screenshot test (Chromium-based): scroll bars are not visible
>       This behavior is acceptable. +1
> 
> 
> I noticed a problem on a few browsers: http://127.0.0.1:9001/blog/bug-journey/ bug demonstration in Brave Linux With your patch, there is both a horizontal scrollbar and a vertical scrollbar. The horizontal scrollbar is expected, but the vertical scrollbar is surprising. This bug happens with Brave (Linux) and Safari (macOS), but does not happen with Firefox (Linux) or Brave (macOS). This issue might be due to some slight differences in font metrics or something... If there's no quick fix, I'm fine with the behavior of your patch.
> 
> Note to self: Looking through a few pages, we really should make some of them more mobile-friendly. There is wasted space in some code blocks due to excessive padding. Your patch makes the problems more obvious (which is a good thing, I think).

I can't figure out any quick fix for the bug with Brave (Linux) and Safari (macOS), but maybe it is something we could tackle in the future.

strager (approved):
This PR looks good to me. :+1:

One issue: The PR summary has a video of an older version of this PR. Can you update the video, or alternatively remove the videos to prevent confusion?

nikohoffren:
> This PR looks good to me. +1
> 
> One issue: The PR summary has a video of an older version of this PR. Can you update the video, or alternatively remove the videos to prevent confusion?

Great +1
I removed the old videos from this PR.

PR #1072, opened 2023-09-08
fix(docs): Reduce scope of lspconfig example to quick-lint-js

### LINKED ISSUE #909: 105$: [nvim-lspconfig] Only lint on insert with quick-lint
The vim docs for quick-lint say that I can lint while in insert mode.  However, while quick-lint is quick, most of my other linters are not quick.  Is there a way to lint in insert mode for _only_ quick-lint?  The docs only describe how to do it for _all_ linters:

```lua
    vim.lsp.handlers['textDocument/publishDiagnostics'] = vim.lsp.with(
      vim.lsp.diagnostic.on_publish_diagnostics, {
        update_in_insert = true,
      }
    )
```

RossBarnie:
Fixes #909

### DIFF
--- a/plugin/vim/quick-lint-js.vim/doc/quick-lint-js.txt
+++ b/plugin/vim/quick-lint-js.vim/doc/quick-lint-js.txt
@@ -213,11 +213,10 @@ module directly: >
     nvim_lsp.quick_lint_js.setup {}
 
 quick-lint-js is fast. To feel the speed, you might want to lint while in
-insert mode. Add the following code to your |init.lua| to make JavaScript
-linting as fast as possible: >
+insert mode. Edit the "handlers" key in the setup function call to make
+JavaScript linting as fast as possible: >
 
-    local nvim_lsp = require('lspconfig')
-    nvm_lsp.quick_lint_js.setup {
+    require('lspconfig/quick_lint_js').setup {
       handlers = {
         ['textDocument/publishDiagnostics'] = vim.lsp.with(
           vim.lsp.diagnostic.on_publish_diagnostics, {
--- a/plugin/vim/quick-lint-js.vim/doc/quick-lint-js.txt
+++ b/plugin/vim/quick-lint-js.vim/doc/quick-lint-js.txt
@@ -216,11 +216,16 @@ quick-lint-js is fast. To feel the speed, you might want to lint while in
 insert mode. Add the following code to your |init.lua| to make JavaScript
 linting as fast as possible: >
 
-    vim.lsp.handlers['textDocument/publishDiagnostics'] = vim.lsp.with(
-      vim.lsp.diagnostic.on_publish_diagnostics, {
-        update_in_insert = true,
+    local nvim_lsp = require('lspconfig')
+    nvm_lsp.quick_lint_js.setup {
+      handlers = {
+        ['textDocument/publishDiagnostics'] = vim.lsp.with(
+          vim.lsp.diagnostic.on_publish_diagnostics, {
+            update_in_insert = true
+          }
+        )
       }
-    )
+    }
 
 The `setup` function can take the following configuration options:
 


RossBarnie:

plugin/vim/quick-lint-js.vim/doc/quick-lint-js.txt
```diff
@@ -216,11 +216,16 @@ quick-lint-js is fast. To feel the speed, you might want to lint while in
 insert mode. Add the following code to your |init.lua| to make JavaScript
 linting as fast as possible: >
 
-    vim.lsp.handlers['textDocument/publishDiagnostics'] = vim.lsp.with(
-      vim.lsp.diagnostic.on_publish_diagnostics, {
-        update_in_insert = true,
+    local nvim_lsp = require('lspconfig')
```
strager, line 219: Must fix: When I tried this, I got the following error:
```
Error detected while processing /nix/store/01pnpqhja7iinp0xl74sygvsiixcr2i6-init.vim:
line   16:
E5108: Error executing lua [string ":lua"]:6: attempt to index global 'nvm_lsp' (a nil value)
stack traceback:
        [string ":lua"]:6: in main chunk
```
Here's the vimrc (Nix template; `${packages.quick-lint-js}` is replaced with a full path):
```vimrc
              lua << EOF
              require('lspconfig/quick_lint_js').setup {
                cmd = {"${packages.quick-lint-js}/bin/quick-lint-js", "--lsp-server"},
              }

              local nvim_lsp = require('lspconfig')
              nvm_lsp.quick_lint_js.setup {
                handlers = {
                  ['textDocument/publishDiagnostics'] = vim.lsp.with(
                    vim.lsp.diagnostic.on_publish_diagnostics, {
                      update_in_insert = true
                    }
                  )
                }
              }
              EOF
```
Did I do something wrong?

---

This vimrc works without issues:
```vimrc
              lua << EOF
              require('lspconfig/quick_lint_js').setup {
                cmd = {"${packages.quick-lint-js}/bin/quick-lint-js", "--lsp-server"},
                handlers = {
                  ['textDocument/publishDiagnostics'] = vim.lsp.with(
                    vim.lsp.diagnostic.on_publish_diagnostics, {
                      update_in_insert = true
                    }
                  )
                },
              }
              EOF
```
I think we should put the setup code inside the existing setup block instead of making a separate setup block.

RossBarnie: Yes that was my intention and this is consistent with the documentation below this section which shows new options but with the call to the setup call. I've added a new commit which should be more consistent and edited the sentence before to further clarify, let me know what you think. I'd be happy to make more changes if needed. The section below this for example covers some specific examples of options available to all language servers via the lspconfig vim API, so maybe we could move this configuration to its own tagged section, and/or link to the lspconfig docs.

strager: > I've added a new commit which should be more consistent and edited the sentence before to further clarify, let me know what you think.

New version looks great!

> The section below this for example covers some specific examples of options available to all language servers via the lspconfig vim API, so maybe we could move this configuration to its own tagged section, and/or link to the lspconfig docs.

Some config options should be set by our plugin and shouldn't normally be customized by the user (for example name, root_dir, on_new_config, and single_file_support (from nvim-lspconfig), and capabilities (from Neovim itself)).

Keeping our current docs, but referencing [nvim-lspconfig's docs](https://github.com/neovim/nvim-lspconfig/blob/cc388d3f6b9c7c943ae2b2dcd46ad470fd257f91/doc/lspconfig.txt#L68), sounds like a good idea. It's not necessary for this PR though.

strager:
Is this ready to merge?

strager (approved):

RossBarnie:
Yes, unless you want the commits squashed.

strager:
Landed as commit b7aa1c87883965af8364246959e7bc8dd86e57a2.

Issue #1067, opened 2023-08-31
8$: error on 'if(c);' in TypeScript mode


strager:
TypeScript does not allow `;` as the body of an `if` statement (and possibly other statements like `else` and `while`.

samir-hamud:
hopelessdev from twitch here. I asked some questions about this issue on your stream yesterday.

I want to give this issue a try. Won't accept payment.

samir-hamud:
Checking on typescriptlang, ts just doesnt allow the if statement. Else and while are allowed. Tested on all ts versions.
 
![image](https://github.com/quick-lint/quick-lint-js/assets/91815230/879457bd-ee4c-4bb7-bc23-b9df346ff242)

strager:
> Else and while are allowed.

Oh, that's strange. Thanks for checking!

PR #1073, opened 2023-09-09
fix(typescript): not allow empty body 'if' with semicolon #1067

samir-hamud:
Fix Issue #1067

### DIFF
--- a/src/quick-lint-js/fe/parse-statement.cpp
+++ b/src/quick-lint-js/fe/parse-statement.cpp
@@ -3977,6 +3977,11 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
     parse_and_visit_body();
     break;
 
+  case Token_Type::semicolon:
+    if (!this->options_.typescript) {
+      parse_and_visit_body();
+      break;
+    }
   case Token_Type::end_of_file:
   case Token_Type::kw_else:
   case Token_Type::right_curly:
--- a/src/quick-lint-js/fe/parse-statement.cpp
+++ b/src/quick-lint-js/fe/parse-statement.cpp
@@ -3977,11 +3977,16 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
     parse_and_visit_body();
     break;
 
+  // if (cond);  // Invalid TypeScript.
   case Token_Type::semicolon:
-    if (!this->options_.typescript) {
+    if (this->options_.typescript) {
+      this->diag_reporter_->report(Diag_Missing_Body_For_If_Statement{
+          .expected_body = this->peek().span(),
+      });
+    } else {
       parse_and_visit_body();
-      break;
     }
+    break;
   case Token_Type::end_of_file:
   case Token_Type::kw_else:
   case Token_Type::right_curly:
--- a/test/test-parse-statement.cpp
+++ b/test/test-parse-statement.cpp
@@ -1249,54 +1249,28 @@ TEST_F(Test_Parse_Statement, if_body_with_semicolon_typescript) {
   {
     Spy_Visitor p = test_parse_and_visit_statement(
         u8"if (a)\n; else e;"_sv,  //
-        u8"      ` Diag_Missing_Body_For_If_Statement"_diag,
+        u8"        ^ Diag_Missing_Body_For_If_Statement"_diag,
         typescript_options);
     EXPECT_THAT(p.visits, ElementsAreArray({
-                              "visit_variable_use", //a
-                          }));
-  }
-
-  {
-    Spy_Visitor p = test_parse_and_visit_statement(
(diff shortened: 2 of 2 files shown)


samir-hamud:

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -3977,6 +3977,11 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
     parse_and_visit_body();
     break;
 
+  case Token_Type::semicolon:
+    if (!this->options_.typescript) {
+      parse_and_visit_body();
+      break;
+    }
```
strager, line 3984: Please explicitly fall through:
```suggestion
  case Token_Type::semicolon:
    if (!this->options_.typescript) {
      parse_and_visit_body();
      break;
    }
    [[fallthrough]];
```

But I think it's better copy-paste the code below and invert the `if`:
```suggestion
  case Token_Type::semicolon:
    if (this->options_.typescript) {
      this->diag_reporter_->report(Diag_Missing_Body_For_If_Statement{
          .expected_body =
              Source_Code_Span::unit(this->lexer_.end_of_previous_token()),
      });
    } else {
      parse_and_visit_body();
    }
    break;
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -3977,6 +3977,11 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
     parse_and_visit_body();
     break;
 
+  case Token_Type::semicolon:
```
strager, line 3980: Nitpick: It's helpful to include a comment explaining what code this code path covers. For example:
```suggestion
  // if (cond);  // Invalid TypeScript.
  case Token_Type::semicolon:
```

test/test-parse-statement.cpp
```diff
@@ -1244,6 +1244,61 @@ TEST_F(Test_Parse_Statement, disallow_label_named_yield_in_generator_function) {
                             "visit_exit_function_scope",
                         }));
 }
+
+TEST_F(Test_Parse_Statement, if_body_with_semicolon_typescript) {
+  {
+    Spy_Visitor p = test_parse_and_visit_statement(
+        u8"if (a)\n; else e;"_sv,  //
+        u8"      ` Diag_Missing_Body_For_If_Statement"_diag,
```
strager, line 1252: Minor: I think the span is a bit confusing. If the user writes a body at the suggested point, they will end up with code like this:
```javascript
if (a) {
}
; else e;
```

I think we should instead highlight the `;` and hint that the `;` needs to be replaced with a body:
```javascript
if (a) {
} else e;
```

```suggestion
        u8"if (a)\n; else e;"_sv,  //
        u8"        ^ Diag_Missing_Body_For_If_Statement"_diag,
```

test/test-parse-statement.cpp
```diff
@@ -1244,6 +1244,61 @@ TEST_F(Test_Parse_Statement, disallow_label_named_yield_in_generator_function) {
...
+        u8"if (a);\n else e;"_sv,  //
+        u8"      ` Diag_Missing_Body_For_If_Statement"_diag,
+        typescript_options);
+    EXPECT_THAT(p.visits, ElementsAreArray({
+                              "visit_variable_use", //a
+                          }));
+  }  
+
+  {
+    Test_Parser p(u8"{\nif (a);\n} b;"_sv, typescript_options, capture_diags);
```
strager, line 1270: What motivated this test case?

test/test-parse-statement.cpp
```diff
@@ -1244,6 +1244,61 @@ TEST_F(Test_Parse_Statement, disallow_label_named_yield_in_generator_function) {
                             "visit_exit_function_scope",
                         }));
 }
+
+TEST_F(Test_Parse_Statement, if_body_with_semicolon_typescript) {
```
strager, line 1248: We should have a similar test but for `else`'s body. E.g. `if (a) {} else;`

strager (approved):

strager (commented):
Patch looks good to me overall. Let me know when I should merge.

test/test-parse-statement.cpp
strager: (According to [your comment](https://github.com/quick-lint/quick-lint-js/issues/1067#issuecomment-1712256942), such code should produce no diagnostic.)

test/test-parse-statement.cpp
samir-hamud: No relevant motivations.
Just thought the more test cases the better. Inexperience issues.

Changed for simplier and direct tests

strager (approved):

strager:
Landed as commit aa4c375452ef6c1184e89d3cb5c1d7af945c0f8a.

PR #1074, opened 2023-09-09
VSCode configuration for QLJS developers

HicaroD:
This file is only matters for developers who uses Visual Studio Code editor for development in quick-lint-js.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,15 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/clang",
+            "cppStandard": "c++20",
+            "intelliSenseMode": "linux-clang-x64"
+        }
+    ],
+    "version": 4
+}
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -6,15 +6,6 @@
             "compilerPath": "/usr/bin/clang",
             "cppStandard": "c++20",
             "intelliSenseMode": "linux-clang-x64",
-            "includePath": [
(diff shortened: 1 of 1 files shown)


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,15 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
```
strager, line 7: Hmm, I don't think this is a good includePath.

Perhaps we should use compileCommands instead?
```json
            "compileCommands": "${workspaceFolder}/build/compile_commands.json"
```

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,15 @@
+{
+    "configurations": [
```
strager, line 2: Do we need macOS and Windows configs too? Or will this config work fine on any OS?

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,15 @@
...
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/clang",
```
strager, line 9: Will VS Code figure out the clang path if it's not specified?

HicaroD: I removed that line `"compilerPath": "/usr/bin/clang"` and VSCode start using `/usr/bin/c++`.

.vscode/c_cpp_properties.json
HicaroD: It will only work on Linux. I unfortunately can't test on Windows or macOS.

strager (approved):

strager:
Landed as commit be4ad05e9d26044f8173a2ebd0a1bdeea195f859. Thanks!

PR #1079, opened 2023-09-13
fix(logging): fix reading UTF-16 trace files on big-endian architectures

strager:


strager:
@chereskata Can you try this commit on your s390x box and let me know if it fixes the test failures? Here's a tarball if you don't want to deal with Git: https://github.com/quick-lint/quick-lint-js/archive/refs/heads/fix-1078.tar.gz

strager:
I tested this myself on a S390x qemu VM. It seemed to work.

PR #1087, opened 2023-09-23
fix: new diag for async/generator on getter/setter

### LINKED ISSUE #1066: 10$: error on async/generator getter/setter
The following code is invalid, but quick-lint-js does not complain:
```js
class C {
  async get foo() { return 42; }
}
```

koopiehoop:
closes #1066 

Can you check new tests? Did I place them right and do they need ```Spy_Visitor``` with ```EXPECT_THAT(p.visits, ...)```?
And what do you think about code repetition (```error_if_generator_method```, ```error_if_async_method```)? Looks like this is the only clean solution without compromising readability

### DIFF
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -3261,6 +3261,26 @@ struct Diag_Missing_Comma_Between_Array_Elements {
                   ARG(expected_comma))]]  //
   Source_Code_Span expected_comma;
 };
+
+struct Diag_Class_Generator_On_Getter_Or_Setter {
+  [[qljs::diag("E0713", Diagnostic_Severity::error)]]  //
+  [[qljs::message("'*' keyword is not allowed on getters or setters",
+                  ARG(generator_keyword))]]                    //
+  [[qljs::message("'{0}' here", ARG(getter_setter_keyword))]]  //
+  Source_Code_Span method_start;
+  Source_Code_Span generator_keyword;
+  Source_Code_Span getter_setter_keyword;
+};
+
+struct Diag_Class_Async_On_Getter_Or_Setter {
+  [[qljs::diag("E0714", Diagnostic_Severity::error)]]  //
+  [[qljs::message("'async' keyword is not allowed on getters or setters",
+                  ARG(async_keyword))]]                        //
+  [[qljs::message("'{0}' here", ARG(getter_setter_keyword))]]  //
+  Source_Code_Span method_start;
+  Source_Code_Span async_keyword;
+  Source_Code_Span getter_setter_keyword;
+};
 }
 QLJS_WARNING_POP
 
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -3264,11 +3264,10 @@ struct Diag_Missing_Comma_Between_Array_Elements {
 
 struct Diag_Class_Generator_On_Getter_Or_Setter {
   [[qljs::diag("E0713", Diagnostic_Severity::error)]]  //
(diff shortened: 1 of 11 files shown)


src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3261,6 +3261,26 @@ struct Diag_Missing_Comma_Between_Array_Elements {
                   ARG(expected_comma))]]  //
   Source_Code_Span expected_comma;
 };
+
+struct Diag_Class_Generator_On_Getter_Or_Setter {
+  [[qljs::diag("E0713", Diagnostic_Severity::error)]]  //
+  [[qljs::message("'*' keyword is not allowed on getters or setters",
+                  ARG(generator_keyword))]]                    //
```
strager, line 3268: I think this is a confusing message. "'*' keyword" doesn't make sense because `*` is not a keyword.

I think a better phrasing is: "getters and setters cannot be generators"

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3261,6 +3261,26 @@ struct Diag_Missing_Comma_Between_Array_Elements {
...
+  [[qljs::message("'{0}' here", ARG(getter_setter_keyword))]]  //
+  Source_Code_Span method_start;
```
strager, line 3270: Nit: I think method_start is unnecessary, so I'd remove it.

src/quick-lint-js/fe/parse-class.cpp
```diff
@@ -1155,6 +1155,8 @@ void Parser::parse_and_visit_class_or_interface_member(
     void check_modifiers_for_method() {
       error_if_accessor_method();
       error_if_declare_method();
+      error_if_generator_method();
+      error_if_async_method();
```
strager, line 1159: Nit: I think "error if async method" is a misleading name. "error if async getter or setter" is more appropriate.
```suggestion
      error_if_generator_getter_or_setter();
      error_if_async_getter_or_setter();
```

test/test-parse-class.cpp
```diff
@@ -836,6 +836,48 @@ TEST_F(Test_Parse_Class, accessor_cannot_be_method) {
...
+      u8"          ^^^ .getter_setter_keyword"_diag,
+      javascript_options);
+  test_parse_and_visit_statement(
+      u8"class C { set async myMethod() { } }"_sv,  //
+      u8"                            ^ Diag_Class_Async_On_Getter_Or_Setter.method_start\n"_diag
+      u8"              ^^^^^ .async_keyword\n"_diag
+      u8"          ^^^ .getter_setter_keyword"_diag,
+      javascript_options);
+  test_parse_and_visit_statement(
+      u8"class C { async get myMethod() { } }"_sv,  //
```
strager, line 853: :+1:

strager (approved):
> Did I place [the tests] right

Yes!

> do [the tests] need `Spy_Visitor` with `EXPECT_THAT(p.visits, ...)`?

No. What you did is great.

> And what do you think about code repetition

We could parameterize the function (parameter for Token_Type::star/Token_Type::kw_async + template parameter for Diag_Class_Generator_On_Getter_Or_Setter/Diag_Class_Async_On_Getter_Or_Setter), but it might not be worth the effort.

strager:
This patch looks good to me. Let me know when I should merge it.

You can ignore the CI failures. I disconnected my ARM build bot (in my home lab) because I'm moving. It'll get hooked up in a few weeks.

koopiehoop:
> Let me know when I should merge it.

All suggested fixes have been applied, if the docs for the site looks good, then you can merge

strager:
Landed as commit be755cbf6b4c95e600a17f9e4dc7075e63d5009f. Thanks!

PR #1088, opened 2023-10-03
MIT=good, GPLv3=meh, not the other way around

mmkal:
Hi - I came across this project from the oxc readme and it looks interesting. However I noticed that in the feature comparison table, it lists the GPLv3 license as "good" where eslint's and Rome's MIT licenses as "meh". While GPLv3 might protect _this_ project better, there's no way GPLv3 is better from the point of view of someone who is considering using it or modifying it. GPLv3 is much more restrictive and many companies don't even allow usage of GPLv3 licensed software without explicit approval. So, I'm opening this PR to make this less misleading for anyone not familiar with those licenses.

### DIFF
--- a/website/public/index.ejs.html
+++ b/website/public/index.ejs.html
@@ -605,11 +605,11 @@ const language = "JavaScript";
             </tr>
             <tr>
               <th class="hide-in-mobile">license</th>
-              <td class="good quick-lint-js">free<br />GPLv3</td>
-              <td class="meh">open source<br />MIT</td>
-              <td class="meh hide-in-mobile">open source<br />MIT</td>
-              <td class="meh">open source<br />MIT</td>
-              <td class="meh hide-in-mobile">open source<br />MIT</td>
+              <td class="meh quick-lint-js">free<br />GPLv3</td>
+              <td class="good">open source<br />MIT</td>
+              <td class="good hide-in-mobile">open source<br />MIT</td>
+              <td class="good">open source<br />MIT</td>
+              <td class="good hide-in-mobile">open source<br />MIT</td>
             </tr>
             <tr>
               <th class="hide-in-mobile">JSX</th>


mmkal:

mmkal:
This is a good example of what I mean!

strager:
> there's no way GPLv3 is better from the point of view of someone who is considering using it or modifying it.

I think you are mistaken about what GPLv3 and MIT do for users and for contributors.

You are correct that MIT is better than GPLv3 for someone modifying the project **if they have access to the source code**. However, if you do not have access to the source code, then MIT is worse than GPLv3.

For someone using the project, GPLv3 or MIT does not matter short term. However, with GPLv3, you are (basically) guaranteed access to the source code, thus you can fix bugs or add features yourself (or commission someone else to fix bugs or add features). This guarantee does not exist for MIT-licensed software.

> GPLv3 is much more restrictive

Agreed. GPLv3 requires distribution of source code, unlike MIT. This "restriction" makes it easier for users to make modifications in the future.

> many companies don't even allow usage of GPLv3 licensed software without explicit approval.

Can you give an example of a GPLv3 dev tool which required explicit approval from an organization due to the GPLv3 license? I am familiar with these kinds of corporate restrictions for libraries, but not for dev tools.

---

> This is a good example of what I mean!

Are you referring to the CLA? The CLA is unrelated to the choice of GPLv3 or MIT.

The Oxc project requires signing a CLA. [Oxc's CLA](https://cla-assistant.io/web-infra-dev/oxc?pullRequest=628) allows ByteDance to close source your Oxc patches.

quick-lint-js's CLA does not allow me to close source your quick-lint-js patches. See the ["License obligations by Us" section](https://github.com/quick-lint/quick-lint-js/blob/4e36b74137ade4dea4480662e30b45349c4daa73/docs/CLA-v1.md#4-license-obligations-by-us) of the CLA for details.

mmkal:
I'll close this since I made the change because I thought this was an oversight. I still disagree - given the source code is available either way, a more permissive license IMO should be considered better for anyone considering using it.

PR #1094, opened 2023-10-21
Feature/warn missing fallthrough

### LINKED ISSUE #1081: Warn on missing fallthrough comment in switch
```javascript
switch (c) {
case a:
  foo();
case b:
  bar();
}
```
This code should warn after `foo();` saying that a fallthrough comment is missing.

yashmasani:
Add and report a diagnostic for missing fallthrough within switch statements.
Fixes #1081


src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -2766,15 +2766,51 @@ void Parser::parse_and_visit_switch(Parse_Visitor_Base &v) {
 
   bool keep_going = true;
   bool is_before_first_switch_case = true;
+  Token prev_token;
   Hash_Set<String8_View> cases;
   while (keep_going) {
+    auto is_comment_line = [&]() -> bool {
```
yashmasani, line 2772: @strager the function to find the comment, `is_comment_line` skips all whitespace, which is similar to the lexer function. However, the lexer handles many more cases of characters. Do you think all those cases should be included within `is_comment_line` as well ?

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -2766,15 +2766,51 @@ void Parser::parse_and_visit_switch(Parse_Visitor_Base &v) {
 
   bool keep_going = true;
   bool is_before_first_switch_case = true;
+  Token prev_token;
```
strager, line 2769: Nit: I think this variable's name is bad. `prev_token` does not refer to the token just before the current token; it refers to a prior `case` or `default` token or the first token of a statement.

Maybe `previous_statement_first_token` is a better name?

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -759,6 +759,13 @@ struct Diag_Duplicated_Cases_In_Switch_Statement {
   Source_Code_Span duplicated_switch_case;
 };
 
+struct Diag_Explicit_Fallthrough_Comment_In_Switch {
```
strager, line 762: Nit: We normally name the diagnostic struct after the *problem*, not the solution. Therefore this should be named something like `Diag_Fallthrough_Without_Comment_In_Switch`.

test/test-parse-warning.cpp
```diff
@@ -436,6 +436,33 @@ TEST_F(Test_Parse_Warning, warn_on_xor_operation_used_as_exponentiation) {
...
 }
+TEST_F(Test_Parse_Warning, Diag_Explicit_Fallthrough_Comment_In_Switch) {
+  test_parse_and_visit_statement(
+      u8"switch(cond1){case 1:\nfoo()\ncase 2:\nbar() //fallthrough\ndefault:}"_sv,  //
+      u8"                       ^^^ Diag_Explicit_Fallthrough_Comment_In_Switch"_diag);
+  test_parse_and_visit_statement(
+      u8"switch(cond1){case 1:\nfoo()\ncase 2:\nlongBarFn()\ndefault:}"_sv,  //
+      u8"                                       ^^^^^^^^^ Diag_Explicit_Fallthrough_Comment_In_Switch"_diag,  //
+      u8"                       ^^^ Diag_Explicit_Fallthrough_Comment_In_Switch"_diag);
+  // check for false positive
```
strager, line 447: :+1:

src/quick-lint-js/fe/parse-statement.cpp
strager: Must fix: We should not repeat comment-parsing code like you have done here.

I see two options:
* Call `Lexer`'s code directly instead of reimplementing or copy-pasting it here; or
* Teach the lexer to communicate when a comment was seen before a token.
We should try the latter solution as I think it will be more reliable and useful. I think adding a `bool has_leading_comment;` into the `Token` class should be simple enough for our needs. `Token::has_leading_newline` already exists, and the code for `has_leading_comment` might be very similar.

test/test-parse-warning.cpp
```diff
@@ -436,6 +436,33 @@ TEST_F(Test_Parse_Warning, warn_on_xor_operation_used_as_exponentiation) {
   test_parse_and_visit_expression(u8"4 ^ 3"_sv, no_diags);
   test_parse_and_visit_expression(u8"(x+2)^a"_sv, no_diags);
 }
+TEST_F(Test_Parse_Warning, Diag_Explicit_Fallthrough_Comment_In_Switch) {
+  test_parse_and_visit_statement(
+      u8"switch(cond1){case 1:\nfoo()\ncase 2:\nbar() //fallthrough\ndefault:}"_sv,  //
+      u8"                       ^^^ Diag_Explicit_Fallthrough_Comment_In_Switch"_diag);
```
strager, line 442: Must fix: I think the diagnostic is in the wrong place. As a user, I would expect the diagnostic to be between the statement and the following `case`, not at the beginning of the statement. In other words, the diagnostic should point me to where I should fix the problem. For example, the diagnostic could be just before `case`:
```c++
      u8"switch(cond1){case 1:\nfoo()\ncase 2:\nbar() //fallthrough\ndefault:}"_sv,  //
      u8"                              ` Diag_Explicit_Fallthrough_Comment_In_Switch"_diag);
```

docs/errors/E0423.md
```diff
@@ -0,0 +1,29 @@
+# E0423: missing fallthrough comment in switch case
```
strager, line 1: "missing fallthrough comment" is good, but if the programmer intended to write a 'break', this diagnostic message is misleading.

Maybe this is a better message:
> missing 'break;' or '// fallthrough' comment between statement and 'case'

For reference, here is the message from [ESLint's no-fallthrough rule](https://eslint.org/docs/latest/rules/no-fallthrough):
> Expected a 'break' statement before 'case'.

strager (changes requested):
Must fix: The following code should **not** report any diagnostic, but your code reports several diagnostics for it:
```javascript
function f(a) {
  for (let c in a) {
    switch (c) {
    case 1:
      return 42;
    case 2:
      break;
    case 3:
      continue;
    case 4:
      throw new Error();
    case 5:
    case 6:
      if (c > 5) {
        return 1;
      } else {
        return 2;
      }
    case 7:
    }
  }
}
```
Fixing most of these cases should be easy, but the `if` looks challenging. For now, I think it's fine if we allow some false negatives by turning off the diagnostic if we see an `if`, `for`, `while`, `try`, etc.

strager:
Thanks for the pull request. I think you're on the right track. There are a lot of things to consider, though, which is one reason I didn't mark the task as for-hire or good-first-issue.

docs/errors/E0423.md
yashmasani: Sounds good. 
Switched to `missing 'break;' or '// fallthrough' comment between statement and 'case'`
FYI: Also had to switch diag code to E0427 since E0423 was used recently.

test/test-parse-warning.cpp
yashmasani: You are right. I went with the suggestion of adding `has_leading_comment` 👍 
Also switched the diag to be just before `case` or `default`.

yashmasani:
> Thanks for the pull request. I think you're on the right track. There are a lot of things to consider, though, which is one reason I didn't mark the task as for-hire or good-first-issue.

Thank you for the review. All comments were very clear and concise so I could make the changes easily.

src/quick-lint-js/fe/parse-statement.cpp
yashmasani: You are right. I went with the suggestion of adding `has_leading_comment` and it worked very well 👍

docs/errors/E0427.md
```diff
@@ -0,0 +1,29 @@
...
+Switch Cases in javascript fallthrough to the next case if the `break` statement is not added at the end of the case.
+Since there is no explicit way of communication whether the fallthrough is intentional or not, it is recommended to use a comment indicating fallthrough.
+
+```javascript
+function test (c) {
+  switch (c) {
+    case 1:
+      foo();
+    default:
+      bar();
```
strager, line 12: These examples assume that globals `foo` and `bar` exist, but they don't. This causes warning E0057 ('use of undeclared variable: foo').

To fix this, I suggest writing `console.log` calls instead. (The `console` global should exist in these examples.)

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -2815,7 +2815,26 @@ void Parser::parse_and_visit_switch(Parse_Visitor_Base &v) {
 
   bool keep_going = true;
   bool is_before_first_switch_case = true;
+  Token previous_statement_first_token;
   Hash_Set<String8_View> cases;
+  auto is_valid_end_of_case = [](Token_Type tk) {
```
strager, line 2820: :+1: Nice helper function.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -2824,6 +2843,13 @@ void Parser::parse_and_visit_switch(Parse_Visitor_Base &v) {
       break;
 
     case Token_Type::kw_case: {
+      if (!is_before_first_switch_case &&
+          !is_valid_end_of_case(previous_statement_first_token.type) &&
```
strager, line 2847: [GCC reports a warning](https://github.com/quick-lint/quick-lint-js/actions/runs/6820808348/job/18550287972?pr=1094) because it thinks `previous_statement_first_token` might be uninitialized here.

To fix the warning, I think making `previous_statement_first_token` a `std::optional` is a decent solution.

test/test-parse.cpp
```diff
@@ -263,6 +263,7 @@ TEST_F(Test_Parse, asi_between_expression_statement_and_switch_label) {
       switch (x) {
         case a:
           f()
+          //fallthrough
```
strager, line 266: :+1:

strager (changes requested):
Must fix: Your patch introduces a false positive diagnostic in the [example for E0140](https://quick-lint-js.com/errors/E0140/):
```javascript
function getText(node) {
  switch (node.nodeType) {
    case document.TEXT_NODE:
      return node.nodeValue;
    case: {
      let result = "";
      for (let child of node.childNodes) {
        result += getText(child, document);
      }
      return result;
    }
    default:
      throw new Error("Unsupported DOM node type");
  }
}
```

strager:
If you want, I can fix the remaining issues. Let me know.

yashmasani:
> If you want, I can fix the remaining issues. Let me know.

I addressed the remaining issues. Thanks!

strager:
Landed as commit 422f3b46349c3de3e5fb737fd48556c352498c3c. :+1:

Issue #1039, opened 2023-06-29
8$: error on multiple 'export default'-s


strager:
```javascript
export default class A {}
export default class B {}
```
The above code is invalid. quick-lint-js should report something like the following:
> line 2: error: cannot export default more than once
> line 1: note: export default previously appeared here

k1ng440:
I want to give this a try. I can't promise a good clean code though and I will not accept payment.

PR #1096, opened 2023-10-24
feat: error on multiple export defaults in module

arieldon:
This closes issue #1039.

### DIFF
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -3307,6 +3307,17 @@ struct Diag_Class_Async_On_Getter_Or_Setter {
   Source_Code_Span async_keyword;
   Source_Code_Span getter_setter_keyword;
 };
+
+struct Diag_Multiple_Export_Defaults {
+  [[qljs::diag("E0715", Diagnostic_Severity::error)]]  //
+  [[qljs::message(
+      "cannot use multiple `export default` statements in one module",
+      ARG(second_export_default))]]  //
+  [[qljs::message("export default previously appeared here",
+                  ARG(first_export_default))]]  //
+  Source_Code_Span second_export_default;
+  Source_Code_Span first_export_default;
+};
 }
 QLJS_WARNING_POP
 
--- a/src/quick-lint-js/fe/parse-statement.cpp
+++ b/src/quick-lint-js/fe/parse-statement.cpp
@@ -35,6 +35,8 @@ bool Parser::parse_and_visit_module_catching_fatal_parse_errors(
 }
 
 void Parser::parse_and_visit_module(Parse_Visitor_Base &v) {
+  QLJS_ASSERT(
+      !this->first_export_default_statement_default_keyword_.has_value());
   bool done = false;
   Parse_Statement_Options statement_options = {
       .possibly_followed_by_another_statement = true,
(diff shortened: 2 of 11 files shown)


arieldon:

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3307,6 +3307,14 @@ struct Diag_Class_Async_On_Getter_Or_Setter {
   Source_Code_Span async_keyword;
   Source_Code_Span getter_setter_keyword;
 };
+
+struct Diag_Multiple_Export_Defaults {
+  [[qljs::diag("E0715", Diagnostic_Severity::error)]] //
+  [[qljs::message("cannot export default more than once", ARG(second_export_default))]]   //
+  [[qljs::message("export default previously appeared here", ARG(first_export_default))]] //
```
strager, line 3314: Nit: Message here should match the title in docs/errors/E0715.md. (It doesn't have to be an exact match.)

For example, you could title the docs "cannot export default more than once" or you could change the first message to "cannot use multiple `export default` statements in one module".

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1011,6 +1012,16 @@ void Parser::parse_and_visit_export(Parse_Visitor_Base &v,
...
   case Token_Type::kw_default:
+    if (this->export_default_statement_span_.has_value()) {
+      this->diag_reporter_->report(
+          Diag_Multiple_Export_Defaults{
+              .second_export_default = this->peek().span(),
+              .first_export_default = *this->export_default_statement_span_,
+          });
+    } else {
+      this->export_default_statement_span_ = this->peek().span();
+    }
```
strager, line 1023: :+1:

src/quick-lint-js/fe/parse.h
```diff
@@ -545,6 +545,10 @@ class Parser {
   void parse_and_visit_named_exports_for_typescript_type_only_import(
       Parse_Visitor_Base &v, Source_Code_Span type_keyword);
 
+  // If set, refers to the first `export default` statement in this module. A
+  // module cannot contain more than one `export default`.
+  std::optional<Source_Code_Span> export_default_statement_span_ = std::nullopt;
```
strager, line 550: :+1:

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1011,6 +1012,16 @@ void Parser::parse_and_visit_export(Parse_Visitor_Base &v,
...
     // export default class C {}
   case Token_Type::kw_default:
+    if (this->export_default_statement_span_.has_value()) {
+      this->diag_reporter_->report(
+          Diag_Multiple_Export_Defaults{
+              .second_export_default = this->peek().span(),
+              .first_export_default = *this->export_default_statement_span_,
+          });
+    } else {
+      this->export_default_statement_span_ = this->peek().span();
```
strager, line 1022: Nit: I think the name `export_default_statement_span_` is wrong. The span is not for the whole statement, but for the 'default' keyword. `first_export_default_statement_default_keyword_` is a good (but long) name. (I think length isn't a problem in this case; clarity is more important.)

strager (approved):
This looks great! Let me know if you want to make changes or if I should merge it now.

arieldon:
I agree with your nitpicks, so I made the changes. I also fixed the error code in the header of the documentation file. It was `E0714`. Now it's `E0715` as it should be. I ran `tools/format` too.

strager:
Landed as commit 4bf1896ce5c973805f3a1723a0ee2444cb6a5a9d. Thanks for the improvement!

Issue #88, opened 2020-11-24
kw_if isssue


JanPoonthong:
This is my code 
```
document.addEventListener('DOMContentLoaded', () => {
    document.querySelector('select').onchange = function() {
        document.querySelector('#hello').style.color = this.value;
    }
});
asjdfhjksdahfj
if :
kjaslkjd;kjaslkjd
```

when i run quick-lint-js, i get 
```
/home/jan/tmp/quick-lint-js/src/parse.cpp:602: fatal: token not implemented in parse_expression_remainder: kw_if on line 8 column 1
zsh: illegal hardware instruction  /home/jan/tmp/quick-lint-js/build/quick-lint-js
```

strager:
Crashes fixed by commits c49dfdc180252896d503a702de3940cad5783da5, e0525e46a1ed52842d4381f785e7f13d3771a6f3, and 25169123a35c3e49c198c8ca195f9c09d6de5f1f.

JanPoonthong:
Perfect!

PR #1101, opened 2023-10-28
fix(fe): don't report missing `else` for bad `if` expression

### LINKED ISSUE #1008: 10$: bad expression in 'if' condition causes E0065: 'else' has no corresponding 'if'
```javascript
    if (e isntanceof DataNotLoadedError) {
    } else {  // E0065: 'else' has no corresponding 'if'
      throw e;
    }
```
quick-lint-js reports E0065 but it should not.

arieldon:
This one line change (potentially) closes #1008. However, test `Test_Parse.utter_garbage` fails -- despite my edits. I don't understand the reason it fails either. To me, it appears like the conditions the test asserts hold true.

The issue provides this snippet of code.
```javascript
function DataNotLoadedError() {}
const e = new DataNotLoadedError();

if (e isntanceof DataNotLoadedError) {
} else {
  throw e;
}
```

For convenience, the CLI version of quick-lint-js behaves like so _before_ this change for the snippet of code provided in the issue.
```sh
$ ~/code/quick-lint-js/build/quick-lint-js x.js 
x.js:4:6: error: if statement is missing ')' around condition [E0018]
x.js:4:17: error: missing semicolon after statement [E0027]
x.js:4:36: error: unmatched parenthesis [E0056]
x.js:7:1: error: 'else' has no corresponding 'if' [E0065]
x.js:4:7: warning: use of undeclared variable: isntanceof [E0057]
```

With the change, it behaves like this.
```sh
$ ~/code/quick-lint-js-fork/build/quick-lint-js x.js 
x.js:4:7: error: unexpected identifier in expression; missing operator before [E0147]
x.js:4:18: error: unexpected identifier in expression; missing operator before [E0147]
x.js:4:7: warning: use of undeclared variable: isntanceof [E0057]
```

For test `Test_Parse.utter_garbage` that it currently fails, the program produces this output.
```
Value of: diagnostics
Expected: has 3 elements and there exists some permutation of elements such that:
 - element #0 has type Diag_Unexpected_Token, and
 - element #1 has type Diag_Unexpected_Identifier_In_Expression, and
 - element #2 has type Diag_Expected_Parentheses_Around_If_Condition
  Actual: { Diag_Unexpected_Token, Diag_Unexpected_Identifier_In_Expression, Diag_Expected_Parentheses_Around_If_Condition }, where the following matchers don't match any elements:
matcher #1: has type Diag_Unexpected_Identifier_In_Expression,
matcher #2: has type Diag_Expected_Parentheses_Around_If_Condition
and where the following elements don't match any matchers:
element #1: Diag_Unexpected_Identifier_In_Expression,
element #2: Diag_Expected_Parentheses_Around_If_Condition
```
But it appears like all the expected diagnoses appear in the actual results, no? A element corresponds to each matcher.

### DIFF
--- a/test/test-parse-statement.cpp
+++ b/test/test-parse-statement.cpp
@@ -595,6 +595,27 @@ TEST_F(Test_Parse_Statement, if_with_else) {
   }
 }
 
+TEST_F(Test_Parse_Statement, if_else_with_malformed_condition) {
+  {
+    Test_Parser p(
+        u8"if (e isntanceof DataNotLoadedError) { } else { throw e; }"_sv,
+        capture_diags);
+    p.parse_and_visit_statement();
+    EXPECT_THAT(p.visits,
+                ElementsAreArray({"visit_variable_use",  // e
+                                  "visit_variable_use",  // typoed instanceof
+                                  "visit_variable_use",  // DataNotLoadedError
+                                  "visit_enter_block_scope",  //
+                                  "visit_exit_block_scope",   //
+                                  "visit_enter_block_scope",  //
+                                  "visit_variable_use",       // e
+                                  "visit_exit_block_scope"}));
+    EXPECT_THAT(
+        p.errors,
+        ::testing::Not(::testing::Contains(DIAG_TYPE(Diag_Else_Has_No_If))));
+  }
+}
+
 TEST_F(Test_Parse_Statement, if_without_body) {
   {
     Spy_Visitor p = test_parse_and_visit_statement(
--- a/test/test-parse.cpp
+++ b/test/test-parse.cpp
@@ -320,8 +320,9 @@ TEST_F(Test_Parse, utter_garbage) {
     assert_diagnostics(
         p.code, p.errors,
         {
-            u8"   ^ Diag_Expected_Parentheses_Around_If_Condition"_diag,  //
-            u8"   ^ Diag_Unexpected_Token"_diag,
+            u8"   ^ Diag_Unexpected_Token"_diag,  // :
+            u8"      ^^^^^^^^ Diag_Unexpected_Identifier_In_Expression"_diag,  // kjaslkjd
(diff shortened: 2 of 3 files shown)


arieldon:
I fixed my original mistake in test `utter_garbage` -- I didn't account for the position and length of spans at first. I also added a test that explicitly confirms E0065 is not included in the set of diagnoses given a bad `if` condition.

test/test-parse-statement.cpp
```diff
@@ -717,6 +717,25 @@ TEST_F(Test_Parse_Statement, else_without_if) {
                               "visit_exit_block_scope",
                           }));
   }
+
+  {
```
strager, line 721: This test case is inside Test_Parse_Statement.else_without_if, but this test case is an `else` *with* and `if`. I think you should make a separate `TEST_F`. Perhaps `Test_Parse_Statement.if_else_with_malformed_condition`? (Naming is hard.)

test/test-parse.cpp
```diff
@@ -320,8 +320,9 @@ TEST_F(Test_Parse, utter_garbage) {
     assert_diagnostics(
```
strager, line 320: Not your patch: I probably should have documented why I introduced this test in the first place. 😅 It's probably just testing a crash, so changing the diagnostics sounds fine to me.

strager (approved):
Looks great! Let me know if you want to make changes or if I should merge it now.

test/test-parse-statement.cpp
arieldon: Thanks for catching that. I moved the new test into a separate `TEST_F` with the name you suggest.

test/test-parse.cpp
arieldon: It's related to issue #88 based on the Git log, and it is testing a crash.

arieldon:
> Looks great! Let me know if you want to make changes or if I should merge it now.

If you're pleased with the most recent changes then you can merge.

strager:
Landed as Git commit 9565209bd978f0e35642a9133ffe77cef3d071dd.

PR #1116, opened 2023-12-14
fix(typescript): report E0429 for class 'extends' clause

### LINKED ISSUE #1114: 8$: report E0429 for class 'extends' clause
The following TypeScript code is invalid, but quick-lint-js accepts it without any diagnostic:
```typescript
export class D<T> {}
export class D2 extends D<<U>() => number> {}
```
quick-lint-js should report E0429 like it does for `implements` and `interface ... extends`.

arieldon:
This closes #1114.

It matches functionality of `implements` and `interface ... extends` by reporting an error when the user defines a class that uses a generic within an `extends` clause without a space between `<` like so `<<`.

For clarity, quick-lint-js reports an error in the following snippet that uses `<<`.
```
export class D<T> {}
export class D2 extends D<<U>() => number> {}
```

quick-lint-js does _not_ report an error in this snippet since it uses `< <`.
```
export class D<T> {}
export class D2 extends D< <U>() => number> {}
```

### DIFF
--- a/src/quick-lint-js/fe/parse-expression.cpp
+++ b/src/quick-lint-js/fe/parse-expression.cpp
@@ -1511,9 +1511,22 @@ next:
   case Token_Type::less:
   case Token_Type::less_less:
     if (this->options_.typescript) {
+      // class C< <T>() => RetType> // Valid.
+      // class C<<T>() => RetType>  // Invalid.
+      if (prec.in_class_extends_clause &&
+          this->peek().type == Token_Type::less_less) {
+        this->diag_reporter_->report(
+            Diag_TypeScript_Generic_Less_Less_Not_Split{
+                .expected_space =
+                    Source_Code_Span::unit(this->peek().begin + 1),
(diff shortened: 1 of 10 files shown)


src/quick-lint-js/fe/parse-expression.cpp
```diff
@@ -1511,9 +1511,22 @@ Expression* Parser::parse_expression_remainder(Parse_Visitor_Base& v,
   case Token_Type::less:
   case Token_Type::less_less:
     if (this->options_.typescript) {
+      // class C< <T>() => RetType> // Valid.
+      // class C<<T>() => RetType>  // Invalid.
+      if (prec.in_class_extends_clause &&
```
strager, line 1516: Oh cool, I didn't know we had `in_class_extends_clause`. That makes this so easy!

strager (approved):
Awesome!

strager:
Landed as commit 58d0516e2d1bca1d50489dc6bfbd2fbbee384658.

PR #1117, opened 2023-12-16
fix(typescript): fix cyclic type false positive

ruipserra:
This fixes a false positive that was reported when linting recursive type definitions such as these:

```ts
type Last<T extends Array<any>> =
  T extends [infer Head] ?
    Head :
    T extends [infer _, ...infer Tail] ?
      Last<Tail> :
      unknown;
```


ruipserra:
Oh, I think this was just fixed in e25d0307f0281f1b3b29b3c2cafebc0f855234b6 🙌

strager:
Yes, it was fixed. (Apparently I squashed two commits together by mistake. Oops!)

I initially had your idea of not reporting if it was generic. However, that still reports false positives such as in `type T = number extends string ? T : false;` (because the false branch is taken).

PR #1120, opened 2023-12-17
fix(typescript): interface properties can be keywords followed by a type

ruipserra:
This fixes an `E0054` (unexpected token) error that was reported whenever an interface property matched a keyword followed by a type.

For example, 
```ts
interface I { get: any }
```
is valid TypeScript, but quick-lint-js was reporting: 

```sh
/test.ts:1:18: error: unexpected token [E0054]
```


strager (approved):
Sweet, thanks!

PR #1133, opened 2023-12-23
feat: new diag for using ‘.’ after ‘?.’

### LINKED ISSUE #1128: 10$: warn on using . after ?.
```javascript
a?.b.c
```
Accessing `.c` might fail because `a?.b` might return `undefined`. (That's the point of `?.`, after all.)

pepplejoshua:
This closes #1128 -- it warns the user about using a dot After an optional chain. This cover optional chain with function call as well as optional chain with index expressions

### DIFF
--- a/docs/errors/E0717.md
+++ b/docs/errors/E0717.md
@@ -7,40 +7,18 @@
 }
 ```
 
-In Javascript, the `?.` operator performs a conditional access only if the left-hand side is not `null` or `undefined`. If the left-hand side is `null` or `undefined`, the expression returns `undefined` and the right-hand side is not evaluated. Using it suppresses the error that would otherwise be thrown if the left-hand side were `null` or `undefined` and the operator `.` were used instead. However, if the left-hand side is not `null` or `undefined`, the expression returns the value of the right-hand side, which may be `null` or `undefined`. If the right-hand side is `null` or `undefined`, using the `.` operator will throw an error:
+In JavaScript, `x?.y` ignores `y` and returns `undefined` if `x` is `null` or `undefined`. If `?.` returns `undefined`, then using `.` after will throw an error:
 
 ```js
-let a = { b: 1 };
-let b = a?.b; // ok
-let c = a?.b?.a; // ok
-let d = a?.b?.a.c; // throws an error
-```
-
-In the example above, the last statement will throw an error, because `a.b` is not `null` or `undefined`, but `a.b.a` is `undefined`, and `undefined.c` throws an error.
-
-This also affects optional chaining with function calls (`?.()`):
-
-```js
-let _a = (_arg) => {
-  return; // so returns undefined
-};
-
-let e = _a?.(a)?.(_a)?.(a); // ok
-let f = _a?.(a).b; // throws an error since _a(a) is undefined
-```
-
-and optional chaining with index expressions (`?.[]`) for the same reason:
-
-```js
-let g = [1, [2, { a: 3 }]];
-let h = g?.[1]; // ok
-let j = g?.[1]?.[2].b; // throws an error since g[1][2] is undefined
+let bug = { milestone: null  };
+console.log(bug.milestone);               // null
+console.log(bug.milestone?.name);         // undefined
+console.log(bug.milestone?.name.trim());  // throws an error
 ```
 
 Replacing the `.` with `?.` will prevent the errors from being thrown at runtime:
 
 ```js
-let d = a?.b?.a?.c; // ok
-let f = _a?.(a)?.b; // ok
-let j = g?.[1]?.[2]?.b; // ok
+let bug = { milestone: null  };
+console.log(bug.milestone?.name?.trim());  // undefined
 ```
--- a/docs/errors/E0717.md
+++ b/docs/errors/E0717.md
@@ -1,0 +1,46 @@
+# E0717: using a '.' after a '?.' might fail, since '?.' might return 'undefined'
+
+```config-for-examples
+{
+  "globals": {
+  }
+}
+```
+
+In Javascript, the `?.` operator performs a conditional access only if the left-hand side is not `null` or `undefined`. If the left-hand side is `null` or `undefined`, the expression returns `undefined` and the right-hand side is not evaluated. Using it suppresses the error that would otherwise be thrown if the left-hand side were `null` or `undefined` and the operator `.` were used instead. However, if the left-hand side is not `null` or `undefined`, the expression returns the value of the right-hand side, which may be `null` or `undefined`. If the right-hand side is `null` or `undefined`, using the `.` operator will throw an error:
+
+```js
+let a = { b: 1 };
+let b = a?.b; // ok
+let c = a?.b?.a; // ok
+let d = a?.b?.a.c; // throws an error
+```
+
+In the example above, the last statement will throw an error, because `a.b` is not `null` or `undefined`, but `a.b.a` is `undefined`, and `undefined.c` throws an error.
+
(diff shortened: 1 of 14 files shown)


pepplejoshua:

docs/errors/E0717.md
```diff
@@ -0,0 +1,46 @@
...
+let b = a?.b; // ok
+let c = a?.b?.a; // ok
+let d = a?.b?.a.c; // throws an error
+```
+
+In the example above, the last statement will throw an error, because `a.b` is not `null` or `undefined`, but `a.b.a` is `undefined`, and `undefined.c` throws an error.
+
+This also affects optional chaining with function calls (`?.()`):
+
+```js
```
strager, line 23: Must fix: Our docs require one wrong/invalid example followed by any number of correct examples. Here you have more than one wrong/invalid example. (This is machine-checked by the [failing "check error docs" CI job](https://github.com/quick-lint/quick-lint-js/actions/runs/7307489068/job/19971420944?pr=1133).)

I think the second example (line 23) and third example (line 34) don't add much value. The point of the error docs (in my mind) isn't to cover every case, but instead to help the reader confirm the problem then figure out a solution.

I think the fix here is to remove the extra examples.

docs/errors/E0717.md
```diff
@@ -0,0 +1,46 @@
+# E0717: using a '.' after a '?.' might fail, since '?.' might return 'undefined'
+
+```config-for-examples
+{
+  "globals": {
+  }
+}
+```
+
+In Javascript, the `?.` operator performs a conditional access only if the left-hand side is not `null` or `undefined`. If the left-hand side is `null` or `undefined`, the expression returns `undefined` and the right-hand side is not evaluated. Using it suppresses the error that would otherwise be thrown if the left-hand side were `null` or `undefined` and the operator `.` were used instead. However, if the left-hand side is not `null` or `undefined`, the expression returns the value of the right-hand side, which may be `null` or `undefined`. If the right-hand side is `null` or `undefined`, using the `.` operator will throw an error:
```
strager, line 10: Nit: I think this explanation is wordy. A beginner is probably going to be confused by so many words.

I think this explanation focuses on what's important for the diagnostic:

> In JavaScript, `x?.y` ignores `y` and returns `undefined` if `x` is `null` or `undefined`. If `?.` returns `undefined`, then using `.` after will throw an error:

docs/errors/E0717.md
```diff
@@ -0,0 +1,46 @@
...
+
+```js
+let a = { b: 1 };
+let b = a?.b; // ok
+let c = a?.b?.a; // ok
+let d = a?.b?.a.c; // throws an error
```
strager, line 16: Nit: This example is confusing to me at a glance because there are multiple things named `a` and `b`. I think this example would be a bit clearer with `console.log` and different names:
```suggestion
let a = { b: 1 };
console.log(a?.b);      // 1
console.log(a?.b?.c);   // undefined
console.log(a?.b?.c.d); // throws an error
```
Even better would be to use more realistic names you might see in a program:
```suggestion
let bug = { milestone: null  };
console.log(bug.milestone);               // null
console.log(bug.milestone?.name);         // undefined
console.log(bug.milestone?.name.trim());  // throws an error
```

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3485,6 +3485,15 @@ struct Diag_Unintuitive_Bitshift_Precedence {
   Source_Code_Span bitshift_operator;
   Source_Code_Span and_operator;
 };
+
+struct Diag_Using_Dot_After_Optional_Chaining {
+  [[qljs::diag("E0717", Diagnostic_Severity::warning)]]
+  [[qljs::message("using a '{0}' after a '{1}' might fail, "
+                  "since '{1}' might return 'undefined'.",
```
strager, line 3492: Nit: It will help maintainers and translators if you use substitutions sparingly. (I don't always follow this rule myself...) `{0}` will always be `.`, and `{1}` will always be `?.`.
Nit: Remove the period at the end of the message.
```suggestion
  [[qljs::message("using a '.' after a '?.' might fail, "
                  "since '?.' might return 'undefined'",
```

src/quick-lint-js/fe/parse-expression.cpp
```diff
@@ -2610,7 +2615,8 @@ Expression* Parser::parse_arrow_function_expression_remainder(
 }
 
 Expression::Call* Parser::parse_call_expression_remainder(Parse_Visitor_Base& v,
-                                                          Expression* callee) {
+                                                          Expression* callee,
+                                                          std::optional<Source_Code_Span> optional_chaining_op_) {
```
strager, line 2619: Nit: Parameters and local variables should not have a trailing underscore:
```suggestion
                                                          std::optional<Source_Code_Span> optional_chaining_op) {
```

src/quick-lint-js/fe/parse-expression.cpp
```diff
@@ -2651,11 +2657,13 @@ Expression::Call* Parser::parse_call_expression_remainder(Parse_Visitor_Base& v,
...
           /*left_paren_span=*/left_paren_span,
-          /*span_end=*/call_span_end));
+          /*span_end=*/call_span_end,
+          /*optional_chaining_op=*/optional_chaining_op_));
 }
 
 Expression* Parser::parse_index_expression_remainder(Parse_Visitor_Base& v,
-                                                     Expression* lhs) {
+                                                     Expression* lhs,
+                                                     std::optional<Source_Code_Span> optional_chaining_op_) {
```
strager, line 2666: Nit: Parameters and local variables should not have a trailing underscore.

src/quick-lint-js/fe/parse.cpp
```diff
@@ -466,6 +466,60 @@ void Parser::error_on_invalid_as_const(Expression* ast,
   }
 }
 
+void Parser::warn_on_dot_operator_after_optional_chain(
+    Expression::Dot *ast) {
+  Expression *lhs = ast->child_;
+  Source_Code_Span operator_span_ = ast->operator_span_;
```
strager, line 472: Nit: Parameters and local variables should not have a trailing underscore:
```suggestion
  Source_Code_Span operator_span = ast->operator_span_;
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -466,6 +466,60 @@ void Parser::error_on_invalid_as_const(Expression* ast,
...
+      if (lhs_operator_span.has_value()) {
+        this->diag_reporter_->report(
+            Diag_Using_Dot_After_Optional_Chaining {
+                .dot_op = operator_span_,
+                .optional_chain_op = *lhs_operator_span,
+            });
+      }
+      break;
+    }
+    default: { /* do nothing */ }
```
strager, line 518: Nit:
```suggestion
    default:
      break;
```

src/quick-lint-js/fe/parse.cpp
```diff
@@ -466,6 +466,60 @@ void Parser::error_on_invalid_as_const(Expression* ast,
   }
 }
 
+void Parser::warn_on_dot_operator_after_optional_chain(
```
strager, line 469: Nice and simple. :+1:

strager (changes requested):
Your patch looks great except for the documentation (see 'must fix' comment).

strager (approved):

strager:
Landed as commit bdfccbdc8c405c5f0d9d40ce8c6736d9c31a6a4a. Thanks for the patch!

Issue #1113, opened 2023-12-11
30$: report misleading indentation for braceless if/etc.


strager:
```javascript
if (x)
  if (y)
    z();
else
  w();
```
quick-lint-js should warn that the `else` should be indented.

vegerot:
A special case can be

```js
if (x)
  y();
  z();
```

quick-lint should warn that `z()` is indented but is not part of the `if`

Issue #282, opened 2021-05-10
30$: report 'unclosed code block' based on indentation


strager:
In the following code, a human can tell that the `if` statement's `{` is unmatched:
```js
function f() {
  if (true) {
    body();
}
```
quick-lint-js is a stupid computer and says that `f`'s `{` is unmatched.

quick-lint-js should track indentation somehow (at least for the purpose of reporting this error) so it can point to the correct unmatched bracket.

PR #1135, opened 2023-12-26
feat: track whitespace

arieldon:
This set of commits addresses issues #282 and #1113 by tracking whitespace and adding a few new diagnostics. quick-lint will print warnings about mismatched curly brace indentation, `if/else` indentation, and misleading indentation after a braceless `if`.

Commit 2379dea modifies a couple existing tests by adding whitespace to them to avoid triggering `Diag_Mismatched_Curly_Indentation`.

### DIFF
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -2841,6 +2841,16 @@ struct Diag_Unclosed_Code_Block_V2 {
   Source_Code_Span expected_block_close;
 };
 
+struct Diag_Misleading_Braceless_If_Else_Indentation {
+  [[qljs::diag("E0442", Diagnostic_Severity::warning)]]  //
+  [[qljs::message("indentation of '{0}' does not match '{1}'", ARG(if_span),
+                  ARG(else_span))]]  //
+  [[qljs::message("indentation of '{0}' does not match '{1}'", ARG(else_span),
+                  ARG(if_span))]]  //
+  Source_Code_Span if_span;
+  Source_Code_Span else_span;
+};
+
 struct Diag_Unclosed_Interface_Block {
   [[qljs::diag("E0215", Diagnostic_Severity::error)]]  //
   [[qljs::message("unclosed interface; expected '}' by end of file",
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -2833,6 +2833,14 @@ struct Diag_Unclosed_Code_Block {
   Source_Code_Span block_open;
 };
 
+struct Diag_Unclosed_Code_Block_V2 {
+  [[qljs::diag("E0441", Diagnostic_Severity::error)]]          //
+  [[qljs::message("missing '}'", ARG(expected_block_close))]]  //
+  [[qljs::message("matching '{0}' here", ARG(block_open))]]    //
+  Source_Code_Span block_open;
+  Source_Code_Span expected_block_close;
+};
+
 struct Diag_Unclosed_Interface_Block {
   [[qljs::diag("E0215", Diagnostic_Severity::error)]]  //
   [[qljs::message("unclosed interface; expected '}' by end of file",
--- a/src/quick-lint-js/diag/diagnostic-types-2.h
+++ b/src/quick-lint-js/diag/diagnostic-types-2.h
@@ -2851,6 +2851,13 @@ struct Diag_Misleading_Braceless_If_Else_Indentation {
   Source_Code_Span else_span;
 };
 
+struct Diag_Misleading_If_Or_Else_Body_Indentation {
+  [[qljs::diag("E0443", Diagnostic_Severity::warning)]]  //
+  [[qljs::message("misleading indentation after '{0}' body",
+                  ARG(if_or_else_span))]]  //
+  Source_Code_Span if_or_else_span;
+};
+
 struct Diag_Unclosed_Interface_Block {
   [[qljs::diag("E0215", Diagnostic_Severity::error)]]  //
   [[qljs::message("unclosed interface; expected '}' by end of file",
--- a/src/quick-lint-js/fe/lex.cpp
+++ b/src/quick-lint-js/fe/lex.cpp
@@ -116,6 +116,7 @@ Lexer::Lexer(Padded_String_View input, Diag_Reporter* diag_reporter)
     : input_(input.data()),
       diag_reporter_(diag_reporter),
       original_input_(input) {
+  this->reset_indent_level();
   this->last_token_.end = this->input_;
   this->parse_bom_before_shebang();
   this->parse_current_token();
@@ -148,6 +149,9 @@ void Lexer::parse_bom_before_shebang() {
   while (!this->try_parse_current_token()) {
     // Loop.
   }
+
+  this->last_token_.indent_level =
+      this->indent_level_ * (this->last_token_.type != Token_Type::end_of_file);
 }
 
 bool Lexer::try_parse_current_token() {
@@ -1843,6 +1847,11 @@ void Lexer::parse_non_ascii() {
   }
 }
 
+inline void Lexer::reset_indent_level() {
+  this->indent_level_ = 0;
+  this->increasing_indent_ = true;
+}
+
 QLJS_WARNING_PUSH
 QLJS_WARNING_IGNORE_CLANG("-Wunknown-attributes")
 QLJS_WARNING_IGNORE_CLANG("-Wunreachable-code")
@@ -1851,24 +1860,30 @@ void Lexer::skip_whitespace() {
   const Char8* input = this->input_;
 
 next:
+  this->increasing_indent_ =
+      this->last_token_.has_leading_newline || input == this->original_input_;
+next_with_possible_indentation:
   Char8 c = input[0];
   unsigned char c0 = static_cast<unsigned char>(input[0]);
   unsigned char c1 = static_cast<unsigned char>(input[1]);
   unsigned char c2 = static_cast<unsigned char>(input[2]);
   if (c == ' ' || c == '\t' || c == '\f' || c == '\v') {
+    this->indent_level_ += this->increasing_indent_;
     input += 1;
-    goto next;
+    goto next_with_possible_indentation;
   } else if (c == '\n' || c == '\r') {
+    this->reset_indent_level();
     this->last_token_.has_leading_newline = true;
     input += 1;
-    goto next;
+    goto next_with_possible_indentation;
   } else if (c0 >= 0xc2) {
     [[unlikely]] switch (c0) {
     case 0xe1:
       if (c1 == 0x9a && c2 == 0x80) {
         // U+1680 Ogham Space Mark
+        this->indent_level_ += this->increasing_indent_;
         input += 3;
-        goto next;
+        goto next_with_possible_indentation;
       } else {
         goto done;
       }
@@ -1894,9 +1909,10 @@ next:
         case 0xa8:  // U+2028 Line Separator
         case 0xa9:  // U+2029 Paragraph Separator
           QLJS_ASSERT(this->newline_character_size(input) == 3);
+          this->reset_indent_level();
           this->last_token_.has_leading_newline = true;
           input += 3;
-          goto next;
+          goto next_with_possible_indentation;
 
         default:
           goto done;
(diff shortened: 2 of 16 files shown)


test/test-parse-function.cpp
```diff
@@ -1445,6 +1445,16 @@ TEST_F(Test_Parse_Function, incomplete_function_body) {
   }
 }
 
+TEST_F(Test_Parse_Function, report_unclosed_code_block_based_on_indentation) {
+  {
+    test_parse_and_visit_statement(
+        u8"function f() {\n\tif (true) {\n\t\tbody();\n}"_sv,  //
+        u8"             ^ Diag_Unclosed_Code_Block"_diag,
+        u8"                            ^ Diag_Mismatched_Curly_Indentation.left_curly\n"_diag  //
+        u8"                                            ^ .right_curly"_diag);
```
strager, line 1454: For this example, I think the best experience would be to report something like this instead:
```suggestion
    test_parse_and_visit_statement(
        u8"function f() {\n\tif (true) {\n\t\tbody();\n}"_sv,  //
        u8"                                            ` Diag_Unclosed_Code_Block_V2.expected_right_curly\n"_diag,
        u8"                            ^ Diag_Unclosed_Code_Block_V2.left_curly\n"_diag);
```
We would report only one diagnostic and it would tell the user how to fix the problem ("missing '}'", "matching '{' here"). This was my intent with #282.

However, if we cannot get such a good diagnostic, then what you have is fine.

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -2833,6 +2833,33 @@ struct Diag_Unclosed_Code_Block {
...
+                  ARG(if_span))]]  //
+  Source_Code_Span if_span;
+  Source_Code_Span else_span;
+};
+
+struct Diag_Misleading_If_Or_Else_Body_Indentation {
+  [[qljs::diag("E0443", Diagnostic_Severity::warning)]]  //
+  [[qljs::message("misleading indentation after 'if' body",
+                  ARG(if_or_else_span))]]  //
+  Source_Code_Span if_or_else_span;
```
strager, line 2860: I think this is the wrong spot. We should point at the following statement or at the indentation before the following statement. The location of the `if` or `else` is supplementary information.
```javascript
  if (cond)
      f();
      g();
//^^^^ warning: misleading indentation after 'if' body
```

src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -2833,6 +2833,33 @@ struct Diag_Unclosed_Code_Block {
...
+                  ARG(else_span))]]  //
+  [[qljs::message("indentation of '{0}' does not match '{1}'", ARG(else_span),
+                  ARG(if_span))]]  //
+  Source_Code_Span if_span;
+  Source_Code_Span else_span;
+};
+
+struct Diag_Misleading_If_Or_Else_Body_Indentation {
+  [[qljs::diag("E0443", Diagnostic_Severity::warning)]]  //
+  [[qljs::message("misleading indentation after 'if' body",
```
strager, line 2858: Nit: I think the diagnostic should say "'else' body" if appropriate:
```suggestion
  [[qljs::message("misleading indentation after '{0}' body",
```

test/test-parse-statement.cpp
```diff
@@ -593,6 +606,19 @@ TEST_F(Test_Parse_Statement, if_with_else) {
...
+    Spy_Visitor p = test_parse_and_visit_statement(
+        u8"if(x)\n\tif (y)\n\t\tz();\nelse\n\tw();"_sv,  //
+        u8"         ^^ Diag_Misleading_Braceless_If_Else_Indentation.if_span\n"_diag  //
+        u8"                           ^^^^ .else_span"_diag);
+    EXPECT_THAT(p.visits, ElementsAreArray({
+                              "visit_variable_use",  // x
+                              "visit_variable_use",  // y
+                              "visit_variable_use",  // z
+                              "visit_variable_use",  // w
+                          }));
```
strager, line 620: Nit: This assertion is unnecessary.
```suggestion
```

test/test-parse-statement.cpp
```diff
@@ -593,6 +606,19 @@ TEST_F(Test_Parse_Statement, if_with_else) {
                               "visit_variable_use",               // b
                           }));
   }
+
+  {
+    Spy_Visitor p = test_parse_and_visit_statement(
+        u8"if(x)\n\tif (y)\n\t\tz();\nelse\n\tw();"_sv,  //
+        u8"         ^^ Diag_Misleading_Braceless_If_Else_Indentation.if_span\n"_diag  //
+        u8"                           ^^^^ .else_span"_diag);
```
strager, line 614: This looks good, though I find `\t` a bit hard to read. Perhaps two spaces is a good substitute:
```suggestion
        u8"if(x)\n  if (y)\n    z();\nelse\n  w();"_sv,  //
        u8"         ^^ Diag_Misleading_Braceless_If_Else_Indentation.if_span\n"_diag  //
        u8"                           ^^^^ .else_span"_diag);
```
Eh, maybe not. We don't have great testing for diagnostics with multi-line code samples...

strager (changes requested):
Thanks for tackling this issue. However, I am concerned about false positives (i.e. diagnostics on correct code). See my comments marked 'must fix'.

src/quick-lint-js/fe/lex.cpp
```diff
@@ -1851,17 +1860,21 @@ void Lexer::skip_whitespace() {
...
 
 next:
+  this->increasing_indent_ = this->last_token_.has_leading_newline || input == this->original_input_;
+next_with_possible_indentation:
   Char8 c = input[0];
   unsigned char c0 = static_cast<unsigned char>(input[0]);
   unsigned char c1 = static_cast<unsigned char>(input[1]);
   unsigned char c2 = static_cast<unsigned char>(input[2]);
   if (c == ' ' || c == '\t' || c == '\f' || c == '\v') {
+    this->indent_level_ += this->increasing_indent_;
```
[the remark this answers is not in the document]
arieldon: I have not tried to implement different ways to track whitespace. I figured the addition would be fairly cheap even in a hot loop, but I didn't measure. I also increment `this->indent_level` in the Ogham Space Mark branch of `skip_whitespace` now -- thanks for catching that.

As for the difference between tabs and spaces, I assume we should let the programmer configure this? I guess we can try to autodetect this too. I don't mix tabs and spaces myself though, so I don't understand the intent of doing so.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1880,36 +1880,43 @@ void Parser::parse_and_visit_statement_block_no_scope(Parse_Visitor_Base &v) {
...
     bool parsed_statement =
         this->parse_and_visit_statement(v, statement_options);
     if (!parsed_statement) {
       switch (this->peek().type) {
       case Token_Type::right_curly:
+        if (this->peek().indent_level != left_curly.indent_level) {
+          this->diag_reporter_->report(Diag_Mismatched_Curly_Indentation{
+              .left_curly = left_curly.span(),
+              .right_curly = this->peek().span(),
+          });
```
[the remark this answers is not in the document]
arieldon: I removed this diagnostic completely in this new set of patches. I replaced it with `Diag_Unclosed_Code_Block_V2`, which I think matches your intention in #282 much more closely.

It will not report any warnings for the snippet of code above. But it will report an error for the snippet of code in #282 as well as others obviously. See the tests for more examples.

Note, this new diagnostic does not catch something like this, where more code trails an expected but missing closing curly:
```javascript
function f() {
  z;
  if (false) {
    z;
    if (1) {
      z;
    }

  if (true) {
  }
}
```
It will report that the curly brace after `if (false)` is missing its matching curly. However, it will suggest the matching rightly curly go at the end of the file instead of before `if (true)`.

arieldon:
I addressed your "must fix" comments. The diagnostics are more flexible now. They do not necessarily report wonky indentation, but they still use indentation to try to track where to place a closing right curly brace. They also still report misleading indentation for `if` and `else` statements.

I have not tried your suggestion about changing the way we track indentation in `Lexer` yet. `Diag_Misleading_If_Or_Else_Body_Indentation` also still reports the `if` and not the actual misleading indentation.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -4402,23 +4420,26 @@ void Parser::parse_and_visit_with(Parse_Visitor_Base &v) {
 
 void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
   QLJS_ASSERT(this->peek().type == Token_Type::kw_if);
-  Source_Code_Span if_token_span = this->peek().span();
+  Token if_token = this->peek();
   this->skip();
 
+  bool previous_in_if_statement = this->in_if_statement_;
+  this->in_if_statement_ = true;
```
strager, line 4427: Must fix: Use `Bool_Guard`. (It's a bit of boilerplate, but the safety is worth it.) See `Switch_Guard` for example usage.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -4436,14 +4458,20 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
       QLJS_PARSER_UNIMPLEMENTED();
     }
 
+    if (!allow_several_statements && this->peek().type != Token_Type::kw_else &&
+        this->peek().indent_level > keyword_token.indent_level) {
+      this->diag_reporter_->report(Diag_Misleading_If_Or_Else_Body_Indentation{
+          .if_or_else_span = keyword_token.span()});
+    }
```
strager, line 4465: This code reports diagnostics for the following code:
```javascript
// misleading indentation after 'if' body [E0443]
if (true) console.log();
 console.log();

function f() {
 if (true) console.log();
 // misleading indentation after 'if' body [E0443]
 else console.log()
  console.log();
}

```
Was this intentional? If so, I think it deserves a test to codify the behavior.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -4468,6 +4496,15 @@ void Parser::parse_and_visit_if(Parse_Visitor_Base &v) {
 
 parse_maybe_else:
   if (this->peek().type == Token_Type::kw_else) {
+    Token else_token = this->peek();
+    if (previous_in_if_statement &&
```
strager, line 4500: Must fix: This detection of whether we're in a braceless `if` or not is faulty. For example, the following code triggers the diagnostic, but if you remove the outer `if` (first and last lines) there is no diagnostic:
```javascript
if (true) {
   // indentation of 'if' does not match 'else' [E0442]
   if (true) {}
  else console.log()

}
```
Another example of faulty `if` detection:
```javascript
if (true) {
  a.map((o) => {
         if (true) return 1;
    else if (true) return 2;
    else return 3;
  });
}
```

test/test-parse-statement.cpp
```diff
@@ -593,6 +606,44 @@ TEST_F(Test_Parse_Statement, if_with_else) {
...
+
+    // if (x)
+    //   if (y)
+    //     z();
+    //   else
+    //     w();
+    test_parse_and_visit_statement(
+        u8"if(x)\n\tif (y)\n\t\tz();\n\telse\n\t\tw();"_sv, no_diags);
```
strager, line 627: For the no_diag tests, we can split the string onto multiple source lines:
```suggestion
    test_parse_and_visit_statement(
        u8"if(x)\n"_sv
        u8"\tif (y)\n"_sv
        u8"\t\tz();\n"_sv
        u8"\telse\n"_sv
        u8"\t\tw();"_sv,
        no_diags);
```

strager (changes requested):
I think you should focus on only one diagnostic for now (probably either `Diag_Unclosed_Code_Block` or `Diag_Misleading_If_Or_Else_Body_Indentation`). You can add the other diagnostics later in a separate pull request.

Must fix: The following code reports a false positive ("indentation of 'if' does not match 'else' [E0442]"):
```javascript
if (true) {
  if (true)
    console.log();
//      console.log();
  else
    console.log();
}
```
I think the comment is significant. Deleting the comment fixes the false positive. (Indentation level of the comment doesn't matter.)

Similarly, there's a false negative in the following code (expected "misleading indentation after 'if' body [E0443]", but got nothing):
```javascript
if (true) {
  if (true)
    console.log();
  //      console.log();
  //else
    console.log();
}
```

arieldon:
I'm not able to figure out a heuristic and an implementation for it that doesn't feel complex and hacky. I'm going to close this PR for now. Thank you for your suggestions and help nonetheless.

Issue #1138, opened 2023-12-30
refactor: rename import_alias -> namespace_alias


strager:

SamInverted:
Does this mean anything named with import_alias should be renamed to namespace_alias?

strager:
Anything related to the following syntax is a namespace alias (despite the fact that the keyword `import` is used):
```typescript
import ns = otherns;
```
Anything related to the following syntax is an import alias (thus shouldn't be renamed):
```typescript
import m = require('mod');
```
(I forgot about the second case when I filed this issue.)

This task would involve going through each instance of `import_alias` or `import alias` in the code, figuring out whether it's an import alias or a namespace alias, and renaming when it's a namespace aliases.

cmstoddard:
@strager close this?

strager:
@cmstoddard No, this issue is still present.

UnfairBots:
Hello, I have a few questions:
- I found some files in test folder, which has some function name like ...import_alias... but verifies namespace alias, does it mean to change function name to ...namespace_alias...?
- In both cases import alias and namespace alias you mentioned, the same error message display (TypeScript import aliases are not allowed in JavaScript- [E0274](https://quick-lint-js.com/errors/E0274/)), do we also need to have a new message to match with namespace alias case? If yes that would be a new implementation, not refactor.

strager:
> I found some files in test folder, which has some function name like ...import_alias... but verifies namespace alias, does it mean to change function name to ...namespace_alias...?

Yes.

> In both cases import alias and namespace alias you mentioned, the same error message display (TypeScript import aliases are not allowed in JavaScript- [E0274](https://quick-lint-js.com/errors/E0274/)), do we also need to have a new message to match with namespace alias case?

I wasn't thinking about this, but it'd be nice to have different messages for the different cases, yes.

PR #1140, opened 2023-12-30
feat(typescript): add new diag for "import type" with import alias

### LINKED ISSUE #1139: 8$: disallow 'import type a = mynamespace;'
The following TypeScript code is invalid:
```typescript
namespace mynamespace {}
import type a = mynamespace;
```
The `type` keyword should be removed from the namespace alias.

koopiehoop:
This closes #1139. I've moved `type_span` before switch statement so we can have better diagnostic instead of pointing on `import` keyword.

### DIFF
--- a/docs/errors/E0717.md
+++ b/docs/errors/E0717.md
@@ -1,0 +1,19 @@
+# E0717: namespace alias cannot use 'import type'
+
+
+The error message "namespace alias cannot use 'import type'" occurs when
+trying to use the 'import type' syntax with an alias.
+
+
+```typescript
+import type A = ns;
+```
+
+
+To fix this error, you need to use the regular 'import' syntax instead
+of 'import type' when using an alias.
+
+
+```typescript
+import A = ns;
+```
--- a/src/quick-lint-js/fe/parse-statement.cpp
+++ b/src/quick-lint-js/fe/parse-statement.cpp
@@ -4607,6 +4607,7 @@ void Parser::parse_and_visit_import(
     }
   };
 
+  std::optional<Source_Code_Span> type_span = std::nullopt;
   switch (this->peek().type) {
   // import var from "module";  // Invalid.
   QLJS_CASE_RESERVED_KEYWORD_EXCEPT_AWAIT_AND_YIELD:
@@ -4703,12 +4704,12 @@ void Parser::parse_and_visit_import(
   // import type from "module";
   case Token_Type::kw_type: {
     // Do not set is_current_typescript_namespace_non_empty_.
-    Source_Code_Span type_span = this->peek().span();
+    type_span = this->peek().span();
     auto report_type_only_import_in_javascript_if_needed = [&] {
       if (!this->options_.typescript) {
         this->diag_reporter_->report(
             Diag_TypeScript_Type_Import_Not_Allowed_In_JavaScript{
-                .type_keyword = type_span,
+                .type_keyword = *type_span,
             });
       }
     };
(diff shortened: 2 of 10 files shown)


src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -4607,6 +4607,7 @@ void Parser::parse_and_visit_import(
     }
   };
 
+  Source_Code_Span type_span = this->peek().span();
```
strager, line 4610: I'd be happier if this was optional. As written, it looks like `type_span` will always be meaningful, but it's only *sometimes* meaningful (but usually it's garbage).
```suggestion
  std::optional<Source_Code_Span> type_span = std::nullopt;
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -4855,6 +4856,12 @@ void Parser::parse_and_visit_import(
           QLJS_PARSER_UNIMPLEMENTED_IF_NOT_TOKEN(Token_Type::right_paren);
           this->skip();
         } else {
+          // import type a = b; // Invalid.
+          if (declared_variable_kind == Variable_Kind::_import_type) {
```
strager, line 4860: Nit: This comment belongs inside the `if`.
```suggestion
          if (declared_variable_kind == Variable_Kind::_import_type) {
            // import type a = b; // Invalid.
```

test/test-parse-typescript-namespace.cpp
```diff
@@ -668,6 +668,19 @@ TEST_F(Test_Parse_TypeScript_Namespace,
...
 
+TEST_F(Test_Parse_TypeScript_Namespace, import_alias_cannot_use_import_type) {
+  {
+    Spy_Visitor p = test_parse_and_visit_statement(
+        u8"import type A = ns;"_sv,  //
+        u8"       ^^^^ Diag_TypeScript_Import_Alias_Cannot_Use_Import_Type.type_keyword"_diag,
+        typescript_options);
+    EXPECT_THAT(p.visits, ElementsAreArray({
+                              "visit_variable_declaration",    // A
+                              "visit_variable_namespace_use",  // ns
```
strager, line 679: :+1:

docs/errors/E0717.md
```diff
@@ -0,0 +1,19 @@
...
+import type A = ns;
+```
+
+
+To fix this error, you need to use the regular 'import' syntax instead
+of 'import type' when using an alias.
+
+
+```typescript
+import { ns as A } from './my-module';
```
strager, line 18: Must fix: This is not the correct fix for the code. The correct fix would be to remove `type`:
```suggestion
import A = ns;
```

docs/errors/E0717.md
```diff
@@ -0,0 +1,19 @@
+# E0717: import alias cannot use 'import type'
```
strager, line 1: Minor: This syntax should be called a **namespace alias**, not an **import alias**. quick-lint-js currently confuses the two terms, so I don't blame you for writing "import alias" here. There's a separate task to fix the wording throughout quick-lint-js (#1138).

strager (changes requested):
Your patch looks good. Thanks! However, there is an issue with the documentation (see the 'must fix' comment).

src/quick-lint-js/fe/parse-statement.cpp
koopiehoop: made it optional

src/quick-lint-js/fe/parse-statement.cpp
koopiehoop: done

docs/errors/E0717.md
koopiehoop: changed "import" to "namespace" and corrected the diagnostic name

strager (approved):

PR #1160, opened 2024-01-05
perf(fe): reduce size of Call and Index expressions

strager:
perf(fe): reduce size of Call and Index expressions

Replace three words (std::optional<Source_Code_Span>) with one word
(Char8*).


strager:
(I was testing something.)

PR #1161, opened 2024-01-05
refactor(test): move React-specific JSX tests into their own file

strager:
refactor(test): move React-specific JSX tests into their own file

I want to eventually introduce Preact support which will interfere
with React support. Move React-specific test cases into
test-parse-jsx-react.cpp.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/quick-lint/quick-lint-js/pull/1161).
* #1162
* __->__ #1161
* #1160


strager:
(I was testing something.)

PR #1162, opened 2024-01-05
fix(fe): don't report React-specific JSX errors in non-React code

strager:
fix(fe): don't report React-specific JSX errors in non-React code

Preact uses JSX but has different rules for attributes. Disable our
React-specific rules if a React import is not detected.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/quick-lint/quick-lint-js/pull/1162).
* __->__ #1162
* #1161
* #1160


strager:
(I was testing something.)

PR #1163, opened 2024-01-05
perf(fe): reduce size of Call and Index expressions

strager:
perf(fe): reduce size of Call and Index expressions

Replace three words (std::optional<Source_Code_Span>) with one word
(Char8*).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/quick-lint/quick-lint-js/pull/1163).
* #1165
* #1164
* __->__ #1163


strager:
(I was testing something.)

PR #1168, opened 2024-01-05
feat: error on 'for await' in non-async function

### LINKED ISSUE #1137: 8$: error on 'for await' in non-async function

dlarocque:
Added diagnostic when a 'for await' statement is in a non-async function, since this statement can only be used in contexts where 'await' can be used.

closes #1137

### DIFF
--- a/src/quick-lint-js/fe/parse-statement.cpp
+++ b/src/quick-lint-js/fe/parse-statement.cpp
@@ -3998,6 +3998,15 @@ void Parser::parse_and_visit_for(Parse_Visitor_Base &v) {
 
   bool is_for_await = this->peek().type == Token_Type::kw_await;
   if (is_for_await) {
+    if (this->in_top_level_) {
+      // TODO: Emit a diagnostic if the top level await mode does not allow
+      // await operators (Parser_Top_Level_Await_Mode::no_await_operator).
+    } else {
+      if (!this->in_async_function_) {
+        this->diag_reporter_->report(
+            Diag_Await_Operator_Outside_Async{this->peek().span()});
+      }
+    }
     this->skip();
   }
 
--- a/test/test-parse-statement.cpp
+++ b/test/test-parse-statement.cpp
@@ -1312,6 +1312,23 @@ TEST_F(Test_Parse_Statement, if_body_with_semicolon_typescript) {
     EXPECT_THAT(p.errors, IsEmpty());
   }
 }
+
+TEST_F(Test_Parse_Statement, disallow_for_await_in_non_async_function) {
+  Spy_Visitor p = test_parse_and_visit_statement(
+      u8"function f() { for await (let x of []) {} }"_sv,  //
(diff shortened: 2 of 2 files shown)


dlarocque:
@strager Do you know why the macOS AArch64 build is failing, or how I might be able to fix it? I don't see how my changes would cause the builds to fail, and I can't find any output from the failing checks.

strager:
> Do you know why the macOS AArch64 build is failing

Those jobs run on my machine in my home lab, and the machine was turned off. :upside_down_face: It should be fixed now.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -3998,6 +3998,11 @@ void Parser::parse_and_visit_for(Parse_Visitor_Base &v) {
 
   bool is_for_await = this->peek().type == Token_Type::kw_await;
   if (is_for_await) {
+    Source_Code_Span await_token_span = this->peek().span();
+    if (!this->in_async_function_) {
+      this->diag_reporter_->report(
+          Diag_Await_Operator_Outside_Async{await_token_span});
+    }
```
strager, line 4005: Clean :+1:

test/test-parse-loop.cpp
```diff
@@ -893,7 +893,8 @@ TEST_F(Test_Parse_Loop, for_of_loop) {
 
   {
     Spy_Visitor p = test_parse_and_visit_statement(
-        u8"for await (let x of xs) { body; }"_sv, no_diags, javascript_options);
+        u8"for await (let x of xs) { body; }"_sv,  //
+        u8"    ^^^^^ Diag_Await_Operator_Outside_Async"_diag);
```
strager, line 897: Must fix: This is wrong, actually. `await` is allowed at the top level.

There are five cases:
1. top level with Parser_Top_Level_Await_Mode::auto_detect
2. top level with Parser_Top_Level_Await_Mode::await_operator
3. top level with Parser_Top_Level_Await_Mode::no_await_operator (not implemented in quick-lint-js yet)
4. in non-async function
5. in async function

Cases 1, 2, and 5 should emit no diagnostic. Cases 3 and 4 should emit a diagnostic. This line is case 1, so there should be no diagnostic.

strager (changes requested):

test/test-parse-loop.cpp
dlarocque: That makes sense- thanks for outlining all the cases! I've updated my commit to revert the change I made to this test, and added a check to make sure that we're not at the top level before reporting the diagnostic.

Since #1137 is regarding an error on `for await...of` in a non-async function, rather than at the top-level, should we leave the third case as a TODO?

strager (approved):

test/test-parse-loop.cpp
strager: > should we leave the third case as a TODO?

Yeah, your TODO comment is fine.

PR #1175, opened 2024-01-08
fix(lex): parse `let f = (): RT<T>=>null;`

### LINKED ISSUE #1169: incorrect parse of '(): RT<T>=> { throw ''; }'
The following TypeScript code is parsed incorrectly by quick-lint-js:
```typescript
class RT<X> {}
class T {}

let f = (): RT<T>=> { throw ''; };
```

vegerot:
fix(lex): parse `let f = (): RT<T>=>null;`

Test Plan:
`let f = (): RT<T>=>null;`

Fixes #1169

### DIFF
--- a/src/quick-lint-js/fe/lex.cpp
+++ b/src/quick-lint-js/fe/lex.cpp
@@ -998,7 +998,13 @@ void Lexer::skip_less_less_as_less() {
 void Lexer::skip_as_greater() {
   switch (this->last_token_.type) {
   case Token_Type::greater_equal:
-    this->last_token_.type = Token_Type::equal;
+    if (this->input_[0] == '>') {
+      // >=>  ➤  > =>
+      this->last_token_.type = Token_Type::equal_greater;
+      this->input_ += 1;
+    } else {
+      this->last_token_.type = Token_Type::equal;
+    }
     break;
   case Token_Type::greater_greater_equal:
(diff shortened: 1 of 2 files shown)


src/quick-lint-js/fe/lex.cpp
```diff
@@ -996,7 +996,12 @@ void Lexer::skip_less_less_as_less() {
 void Lexer::skip_as_greater() {
   switch (this->last_token_.type) {
   case Token_Type::greater_equal:
-    this->last_token_.type = Token_Type::equal;
+    if (this->input_[0] == '>') {
+      this->last_token_.type = Token_Type::equal_greater;
```
strager, line 1000: Nit: I think a comment would be helpful:
```suggestion
      // >=>  ➤  > =>
      this->last_token_.type = Token_Type::equal_greater;
```

strager, line 1000: Must fix: The `=>` token's .end is incorrect after this change. The following example demonstrates the bug:
```typescript
class C<T> {
  f(): C<T>=> {
  }
}
```

strager (changes requested):

src/quick-lint-js/fe/lex.cpp
vegerot: Fixed.  What do you think?

PR #1187, opened 2024-01-15
unexpected 'await' keyword on function declaration

HicaroD:


src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3599,6 +3599,15 @@ struct Diag_Using_Dot_After_Optional_Chaining {
   Source_Code_Span dot_op;
   Source_Code_Span optional_chain_op;
 };
+
+struct Diag_Unexpected_Await_On_Function_Declaration {
+  [[qljs::diag("E0719", Diagnostic_Severity::warning)]]  //
+  // clang-format off
+  [[qljs::message("Unexpected 'await' keyword on function declaration. Maybe you meant 'async'?",
+                  ARG(await_keyword))]]  //
```
strager, line 3607: You should follow the syntactic style of existing diagnostics (like E0178) instead:
```suggestion
  [[qljs::message("unexpected 'await' keyword on function declaration; maybe you meant 'async'?",
                  ARG(await_keyword))]]  //
```

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (C) 2020  Matthew "strager" Glazar
 // See end of file for extended copyright information.
 
+#include <cstdio>
```
strager, line 4: Nit: Undo.

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -417,6 +418,20 @@ bool Parser::parse_and_visit_statement(Parse_Visitor_Base &v,
...
+                 .begin = this->peek().begin,
+                 .require_name = Name_Requirement::required_for_statement,
+                 .async_keyword = std::nullopt,
+                 .declare_keyword = std::nullopt,
+             });
+      if (this->peek().type != Token_Type::left_paren) {
+        this->diag_reporter_->report(
+            Diag_Unexpected_Await_On_Function_Declaration{
+                .await_keyword = await_token.span()});
+      }
```
strager, line 434: Must fix: Your special handling of `function` is wrong in a few ways:

* `await function foo() {}();` is parsed incorrectly. You have a test for it, but `test_parse_and_visit_statement` doesn't assert that the whole input is parsed. (It probably should... I should fix that.) This code is parsing `await function foo() {}` as a statement, leaving `();` as a second statement (which is wrong!).
* `await function () {}` falsely reports E0061 ("missing name in function statement").

Also, adding code here in parse_and_visit_expression doesn't catch mistakes such as the following:
```javascript
foo(await function() {
});
```

`parse_await_expression` is probably a better place to put your logic.

strager (changes requested):

PR #1202, opened 2024-02-20
refactor(util): use parse_integer from simdjson

### LINKED ISSUE #915: 5$: refactor: Reuse simdjson's number parsing in quick-lint-js
quick-lint-js/port/integer.cpp contains implementations for `from_chars` functions. These functions either use `std::from_chars` or `std::strtol` or a manual (slow) implementation.

We already use simdjson. We should try using simdjson's `parse_unsigned` and `parse_integer` functions instead. This might improve performance, but more importantly, it might reduce binary size.

msharipov:
simdjson provides functions for parsing decimal integers that should run faster than manually coded equivalent. Changed parse_integer_exact to use the parse_integer function from simdjson. Due to inconsistencies between what simdjson and the original implementation consider to be an error, some pre-emptive checks are run on the input string to avoid false positives/negatives. closes #915

### DIFF
--- a/src/quick-lint-js/util/integer.cpp
+++ b/src/quick-lint-js/util/integer.cpp
@@ -21,6 +21,15 @@
 #include <string>
 #include <type_traits>
 
+#ifdef SIMDJSON_SWAR_NUMBER_PARSING
+#if SIMDJSON_SWAR_NUMBER_PARSING
+#undef SIMDJSON_SWAR_NUMBER_PARSING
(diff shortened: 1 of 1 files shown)


msharipov:

msharipov:
recheck

strager:
As I wrote in the GitHub issue (#915), I don't think reusing simdjson makes sense for us. I am therefore won't merge your changes and am closing this pull request.

PR #1204, opened 2024-02-24
Configure git to handle line endings automatically

jaitjacob:
Configure Git to handle line endings automatically with people who use different operating systems. The advantage of this is that your end of line configuration now travels with your repository and you don't need to worry about whether or not collaborators have the proper global settings.

Reference: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,3 +9,6 @@
 /docs/** linguist-documentation
 /docs/analytics/** -linguist-documentation
 /website/** linguist-documentation
+
+# Auto detect text files and perform LF normalization
+* text=auto


.gitattributes
```diff
@@ -9,3 +9,6 @@
 /docs/** linguist-documentation
 /docs/analytics/** -linguist-documentation
 /website/** linguist-documentation
+
+# Auto detect text files and perform LF normalization
+* text=auto
```
strager, line 14: How does this fix the issue you described in your email? From my understanding of the documentation, this gitattributes entry is equivalent to doing nothing. It's a no-op.

jaitjacob: Sorry,  this is incomplete and doesn't do what the title says. There are 3 more lines + comments that needs to go with this. I will add them now.

strager:
I don't think this patch is working.

I disabled my Git config to restore Git's defaults. I then cloned a fresh repo on my Windows machine with this branch. Then I followed the instructions to [build quick-lint-js](https://quick-lint-js.com/contribute/build-from-source/windows/). `git status` shows that there were changes:
```
> git status
On branch pr1204
Your branch is up to date with 'origin/pr1204'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   po/messages.pot
        modified:   src/quick-lint-js/debug/public/trace-generated.mjs
        modified:   src/quick-lint-js/diag/diagnostic-metadata-generated.cpp
        modified:   src/quick-lint-js/diag/diagnostic-metadata-generated.h
        modified:   src/quick-lint-js/fe/lex-keyword-generated.cpp
        modified:   src/quick-lint-js/i18n/translation-table-generated.cpp
        modified:   src/quick-lint-js/i18n/translation-table-generated.h
        modified:   src/quick-lint-js/i18n/translation-table-test-generated.h
        modified:   src/quick-lint-js/logging/trace-metadata-generated.cpp
        modified:   src/quick-lint-js/logging/trace-reader-generated.cpp
        modified:   src/quick-lint-js/logging/trace-reader-generated.h
        modified:   src/quick-lint-js/logging/trace-writer-generated.h
        modified:   test/test-configuration-loader.cpp
        modified:   tools/trace.bt

no changes added to commit (use "git add" and/or "git commit -a")
```
(The change to `test/test-configuration-loader.cpp` is my fault. I had to make a change manually in order to successfully build. Seems like a bug in master unrelated to your patch...)

strager (changes requested):
How did you test this patch? Maybe I'm confused about what it's supposed to be doing exactly.

PR #1211, opened 2024-03-11
feat(fe): add warning for 'let({x} = y);'

### LINKED ISSUE #1203: 10$: Warn on let ({x} = y);
Before refactor:
```javascript
let x;
({x} = y);
```
After refactor:
```javascript
let ({x} = y);  // Bug! This uses an undeclared variable 'x' and calls a function named 'let'.
```

Conditions for this diagnostic:
* Function call with a variable `let` as the callee.
* Exactly one function argument.
* Function argument is an assignment expression with `=`.
* Function argument's left-hand side is an object literal.

msharipov:
Added a new diagnostic E0720 that closes #1203. Issues a warning when a function let() is called on an object assignment expression.

https://github.com/quick-lint/quick-lint-js/issues/1203

### DIFF
--- a/docs/errors/E0720.md
+++ b/docs/errors/E0720.md
@@ -1,39 +1,18 @@
-# E0720: calling `let` on an assignment is ambiguous
+# E0720: function 'let' call may be confused for destructuring; remove parentheses to declare a variable
 
-If `let` is declared as a function, then this code:
+In JavaScript, variables can be named `let` and interpreted as a function
+call if it is followed by parentheses. This code calls function `let`
+instead of destructuring an object:
 
 ```javascript
-function let(arg) {
-  console.log(arg);
-}
-
-const y = {x: "hello", z: "world"};
-let ({x} = y);
+const words = {first: "hello", second: "world"};
+let ({first} = words);
 ```
 
-is actually equivalent to:
+If you want to declare a variable, remove the parentheses:
 
 ```javascript
-function let(arg) {
-  console.log(arg);
-}
-
-const y = {x: "hello", z: "world"};
-let(y);
-```
-
-instead of destructuring `y`.
-
-To remove the ambiguity, separate the declaration and the assignment into two
-statements:
-
-```javascript
-function let(arg) {
-  console.log(arg);
-}
-
-const y = {x: "hello", z: "world"};
-let x;
-({x} = y);
+const words = {first: "hello", second: "world"};
+let {first} = words;
 
 ```
--- a/docs/errors/E0720.md
+++ b/docs/errors/E0720.md
@@ -1,0 +1,39 @@
+# E0720: calling `let` on an assignment is ambiguous
+
+If `let` is declared as a function, then this code:
+
+```javascript
+function let(arg) {
+  console.log(arg);
+}
+
+const y = {x: "hello", z: "world"};
+let ({x} = y);
+```
+
+is actually equivalent to:
+
+```javascript
+function let(arg) {
+  console.log(arg);
+}
+
+const y = {x: "hello", z: "world"};
+let(y);
+```
+
+instead of destructuring `y`.
+
+To remove the ambiguity, separate the declaration and the assignment into two
+statements:
+
+```javascript
+function let(arg) {
+  console.log(arg);
+}
+
+const y = {x: "hello", z: "world"};
+let x;
+({x} = y);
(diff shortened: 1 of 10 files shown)


src/quick-lint-js/diag/diagnostic-types-2.h
```diff
@@ -3589,6 +3589,16 @@ struct Diag_TypeScript_Namespace_Alias_Cannot_Use_Import_Type {
                   ARG(type_keyword))]]  //
   Source_Code_Span type_keyword;
 };
+
+struct Diag_Ambiguous_Let_Call {
+  [[qljs::diag("E0720", Diagnostic_Severity::warning)]]  //
+  [
+      [qljs::message("ambiguous use of the keyword 'let'; "
+                     "let is a function in this scope",
```
strager, line 3597: "ambiguous" is incorrect. All JS engines will consider `let` to be a function reference in this case; there is no ambiguity. Perhaps a better word is "confusing": "confusing use of the keyword 'let'; let is a function in this scope"

There's another problem with this message: It doesn't help the user solve the issue. So I propose wording such as the following:
> 'let' is treated as a function; remove parentheses to instead declare a variable

src/quick-lint-js/fe/parse-statement.cpp
```diff
@@ -162,6 +162,18 @@ bool Parser::parse_and_visit_statement(Parse_Visitor_Base &v,
       Expression *ast =
           this->parse_expression(v, Precedence{.in_operator = true});
       this->visit_expression(ast, v, Variable_Context::rhs);
+
+      // let ({x} = y);  // Ambiguous if 'let' is a function
+      if (ast->kind() == Expression_Kind::Call) {
+        Expression::Call *call = expression_cast<Expression::Call *>(ast);
+        if (call->child_count() == 2 &&
```
strager, line 169: Must fix: You are missing a check for `call->child_0()`. This code reports a diagnostic for `let.prop({x}=y);` but that code should not have a diagnostic.

docs/errors/E0720.md
```diff
@@ -0,0 +1,39 @@
+# E0720: calling `let` on an assignment is ambiguous
```
strager, line 1: The title here should match the message in the code:
```suggestion
# E0720: ambiguous use of the keyword 'let'; let is a function in this scope
```

docs/errors/E0720.md
```diff
@@ -0,0 +1,39 @@
...
+```javascript
+function let(arg) {
+  console.log(arg);
+}
+
+const y = {x: "hello", z: "world"};
+let(y);
+```
+
+instead of destructuring `y`.
```
strager, line 25: Must fix: This explanation is wrong. You removed the assignment to `x`. The second snippet should look more like this:
```javascript
function let(arg) {
  console.log(arg);
}

const y = {x: "hello", z: "world"};
({x} = y);      // <-- your explanation is missing this line.
let(y);
```
The explanation also says "If `let` is declared as a function", but it doesn't matter what `let` is declared as (or if it's even declared at all).

Also, the style of the explanation doesn't match our typical style. We shouldn't split the sentence into pieces with code blocks in the middle, and we shouldn't use two code blocks to explain the buggy code.

I suggest a rewrite of this section to something like the following:
```suggestion
In JavaScript, variables can be named `let`. If a `let` statement has parentheses in the wrong spot, `let` itself is treated as a variable name. The statement is thus a function call, not a normal `let` statement:

```javascript
const person = {name: "Alice", age: 9001};
// The following line does
// *not* declare any variables.
let ({name, age} = person);
```-
```

docs/errors/E0720.md
```diff
@@ -0,0 +1,39 @@
...
+
+To remove the ambiguity, separate the declaration and the assignment into two
+statements:
```
strager, line 28: The preferred solution is actually to remove parentheses: `let {x} = y;`

strager (changes requested):
Looks good overall, but there's a bug which must be fixed ('must fix' comment), and the docs need some improvement ('must fix' comment).

msharipov:
@strager Thank you for the feedback! I have made the necessary corrections to the pull request.

strager:
Landed your patch as Git commit c8b32584f1ad8d1bc88a293e451886880c6d0bb9.

PR #1219, opened 2024-04-28
feat(fe): add warning for 'x==y;' statement

### LINKED ISSUE #1132: 12$: warn on 'x == y;' statement
The following C++ code has a bug:
```c++
  for (Used_Variable &used_var : current_scope.variables_used) {
    if (used_var.kind == Used_Variable_Kind::use) {
      used_var.kind == Used_Variable_Kind::use_in_type;
    }
  }
```
JavaScript code might have a similar bug. For example:
```javascript
let x;
x == 42;
```
quick-lint-js should warn about such redundant `==` (or `===`) and suggest `=` instead.

CoderMuffin:
fixes #1132
If you are satisfied I am happy for you to merge immediately :)
While making this I was thinking that maybe this could be widened (or a second diagnostic could be introduced) for statements that are completely redundant (e.g. `x+y;`)? If this is implemented expressions that can have side effects like `x()`, `x.a()` or `x() ? y : z` will have to be checked for, as warning on these could advise removal that would incorrectly affect program flow.

### DIFF
--- a/docs/errors/E0459.md
+++ b/docs/errors/E0459.md
@@ -1,0 +1,23 @@
+# E0459: equality check result is unused; did you mean to use assignment (=) instead?
+
+Using `==` as a statement rather than a comparison is almost never intended.
+
+```javascript
+let x = 4;
+let y = 5;
+
+// set x to y - oops, one = instead of two so no assignment is made!
+x == y;
+```
+
+To resolve this, either correct to `=` instead:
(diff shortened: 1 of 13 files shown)


CoderMuffin:
Not sure what to do regarding the failing docs check on E0343 and E0342, is there any way to declare that the new E0459 should also be triggered here?

PR #1228, opened 2024-08-30
feat(fe): add a warning when comparing typeof to invalid string literals

### LINKED ISSUE #1213: [diagnostic] typeof operator should warn when comparing to invalid string literals
When performing checks like `typeof x == "string"`, it is perfectly possible to misspell it as something like `typeof x == "strng"` (missing `i`). Maybe there should be a warning for this? I would be happy to implement this if approved :)

dubloom:
fixes #1213 

Add a new diagnostic E0721 that warns on comparisons between typeof operator and a string literal which is not one of the existing possibilites listed here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

Only one test case, which throw a warning, has been added.


dubloom:

dubloom:
recheck

dubloom:
recheck

PR #1229, opened 2024-09-23
Improve installation section in README

andjimenezuf:


strager:
Thanks for the improvements. I definitely think that your README is more helpful than the existing one.

Issue #1233, opened 2025-01-07
warn on mixing ?? with > or ==


strager:
```javascript
config?.items?.length ?? 0 > 0
// intended: (config?.items?.length ?? 0) > 0
// actual: config?.items?.length ?? (0 > 0)
```

https://x.com/strager/status/1876508636236964158

VIJESHG:
Hello @strager I have created PR for this issue: https://github.com/quick-lint/quick-lint-js/pull/1234 Please review and add

PR #1234, opened 2025-02-23
add E0722 to warn on mixing ?? with > or == #1233

VIJESHG:
# E0722: mixing `??` with `>` or `==` may lead to unexpected behavior; use parentheses to clarify the intended precedence


VIJESHG:

